PR #2, opened 2016-01-04
[WIP] Adding support for java projects.

liamjbennett:
- [x] `sous detect` working with pom.xml
- [x] `sous dockerfile` generating something sensible.
- [ ] use the java-specific base image / or install maven
- [ ] support tags with the format `<name>-<version>` (currently warning with 0.0.0)
- [ ] support multi-module projects

### DIFF
--- a/core/detect.go
+++ b/core/detect.go
@@ -12,6 +12,7 @@ func DetectProjectType(packs []Pack) Pack {
 	var pack Pack
 	for _, p := range packs {
 		if err = p.Detect(); err != nil {
+			cli.Logf(err.Error())
 			continue
 		}
 		if pack != nil {
--- a/packs/java/app_target.go
+++ b/packs/java/app_target.go
@@ -1,0 +1,104 @@
+package java
+
+import (
+	"fmt"
+	"path"
+	"strconv"
+
+	"github.com/opentable/sous/core"
+	"github.com/opentable/sous/tools/cli"
+	"github.com/opentable/sous/tools/docker"
+	"github.com/opentable/sous/tools/file"
+	"github.com/opentable/sous/tools/ports"
+)
+
+type AppTarget struct {
+	*JavaTarget
+	artifactPath string
+}
+
+func NewAppTarget(pack *Pack) *AppTarget {
+	return &AppTarget{NewJavaTarget("app", pack), ""}
+}
+
+func (t *AppTarget) DependsOn() []core.Target {
+	return []core.Target{
+		NewCompileTarget(t.pack),
+	}
+}
+
+func (t *AppTarget) RunAfter() []string { return []string{"compile"} }
+
+func (t *AppTarget) Desc() string {
+	return "The Java app target simply places a compiled jar file inside a plain ubuntu base image"
+}
+
+// Checking a Java project always passes.
+func (t *AppTarget) Check() error {
+	return nil
+}
+
+func (t *AppTarget) PreDockerBuild(c *core.Context) {
+	if t.artifactPath == "" {
+		cli.Fatalf("Artifact path not set by compile target.")
+	}
+	if !file.Exists(t.artifactPath) {
+		cli.Fatalf("Artifact not at %s", t.artifactPath)
+	}
+	filename := path.Base(t.artifactPath)
+	localArtifact := filename
+	file.TemporaryLink(t.artifactPath, "./"+localArtifact)
+	t.artifactPath = localArtifact
+}
+
+func (t *AppTarget) Dockerfile(c *core.Context) *docker.Dockerfile {
+	if t.artifactPath == "" {
+		// Actually, it is first set by compile target, then the PreDockerBuild
+		// step links it into the WD and resets artifactPath to a local, relative
+		// path.
+		t.artifactPath = "<¡ artifact path set by compile target !>"
+	}
+	df := &docker.Dockerfile{}
+	df.From = t.pack.baseImageTag("app")
+
+	df.AddAdd(t.artifactPath, "/srv/app/")
+
+	df.Entrypoint = []string{
+		fmt.Sprintf("/srv/app/%s-%s", c.CanonicalPackageName(), c.BuildVersion),
+	}
+	return df
+}
+
+func (t *AppTarget) SetState(fromTarget string, state interface{}) {
+	if fromTarget != "compile" {
+		return
+	}
+	m, ok := state.(map[string]string)
+	if !ok {
+		cli.Fatalf("app target got a %T from compile target, expected map[string]string", state)
+	}
+	artifactPath, ok := m["artifactPath"]
+	if !ok {
+		cli.Fatalf("app target got %+v from compile target; expected key 'artifactPath'", m)
+	}
+	t.artifactPath = artifactPath
+}
+
+func (t *AppTarget) DockerRun(c *core.Context) *docker.Run {
+	dr := docker.NewRun(c.DockerTag())
+	port0, err := ports.GetFreePort()
+	if err != nil {
+		cli.Fatalf("Unable to get free port: %s", err)
+	}
+	dr.AddEnv("PORT0", strconv.Itoa(port0))
+	dr.AddEnv("TASK_HOST", core.DivineTaskHost())
+	return dr
+}
+
+func (t *AppTarget) ContainerName(c *core.Context) string {
+	return c.CanonicalPackageName()
+}
+
+func (t *AppTarget) ContainerIsStale(c *core.Context) (bool, string) {
+	return true, "it is not reusable"
+}
--- a/packs/java/compile_target.go
+++ b/packs/java/compile_target.go
@@ -1,0 +1,105 @@
+package java
+
+import (
+	"fmt"
+
+	"github.com/opentable/sous/core"
+	"github.com/opentable/sous/tools/cmd"
+	"github.com/opentable/sous/tools/dir"
+	"github.com/opentable/sous/tools/docker"
+)
+
+type CompileTarget struct {
+	*JavaTarget
+}
+
+func NewCompileTarget(pack *Pack) *CompileTarget {
+	return &CompileTarget{NewJavaTarget("compile", pack)}
+}
+
+func (t *CompileTarget) DependsOn() []core.Target {
+	return nil
+}
+
+func (t *CompileTarget) RunAfter() []string { return nil }
+
+func (t *CompileTarget) Desc() string {
+	return "The Java compile target generates a single jar file"
+}
+
+// Checking a Java project always passes.
+func (t *CompileTarget) Check() error {
+	return nil
+}
+
+func (t *CompileTarget) Dockerfile(c *core.Context) *docker.Dockerfile {
+	df := &docker.Dockerfile{}
+	df.From = t.pack.baseImageTag(t.Name())
+	c.TemporaryLinkResource("build-prep.bash")
+	buildPrepContainerPath := "/build-prep.bash"
+	df.AddAdd("build-prep.bash", buildPrepContainerPath)
+	df.AddRun(fmt.Sprintf("chmod +x %s", buildPrepContainerPath))
+	// This is a non-portable image, since it includes the UID of the
+	// logged-in user.
+	uid := cmd.Stdout("id", "-u")
+	gid := cmd.Stdout("id", "-g")
+	username := cmd.Stdout("whoami")
+	// Just use the username for group name, it doesn't matter as long as
+	// the IDs are right.
+	groupAdd := fmt.Sprintf("groupadd -g %s %s", gid, username)
+	// Explanation of some of the below useradd flags:
+	//   -M means do not create home directory, which we do not need
+	//   --no-log-init means do not create a 32G sparse file (which Docker commit
+	//       cannot handle properly, and tries to create a non-sparse 32G file.)
+	userAdd := fmt.Sprintf("useradd --no-log-init -M --uid %s --gid %s %s", uid, gid, username)
+	df.AddRun(fmt.Sprintf("%s && %s", groupAdd, userAdd))
+	df.Entrypoint = []string{"/build-prep.bash"}
+	return df
+}
+
+func (t *CompileTarget) ContainerName(c *core.Context) string {
+	return fmt.Sprintf("%s_reusable-builder", c.CanonicalPackageName())
+}
+
+func (t *CompileTarget) ContainerIsStale(c *core.Context) (bool, string) {
+	return true, "it is not reusable"
+}
+
+func (t *CompileTarget) DockerRun(c *core.Context) *docker.Run {
+	containerName := t.ContainerName(c)
+	run := docker.NewRun(c.DockerTag())
+	run.Name = containerName
+	run.AddEnv("ARTIFACT_NAME", t.artifactName(c))
+	uid := cmd.Stdout("id", "-u")
+	gid := cmd.Stdout("id", "-g")
+	artifactOwner := fmt.Sprintf("%s:%s", uid, gid)
+	run.AddEnv("ARTIFACT_OWNER", artifactOwner)
+	artDir := t.artifactDir(c)
+	dir.EnsureExists(artDir)
+	run.AddVolume(artDir, "/artifacts")
+	run.AddVolume(c.WorkDir, "/wd")
+
+	// TODO: if project is a multi-module build then cd to the root to build
+	run.Command = "mvn package"
+	return run
+}
+
+// State returns any interesting state from this target to and dependent targets
+// in the build chain.
+func (t *CompileTarget) State(c *core.Context) interface{} {
+	return map[string]string{
+		"artifactPath": t.artifactPath(c),
+	}
+}
+
+func (t *CompileTarget) artifactPath(c *core.Context) string {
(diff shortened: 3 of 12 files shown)


liamjbennett:
//cc @samsalisbury @stevenschlansker

core/detect.go
```diff
@@ -12,6 +12,7 @@ func DetectProjectType(packs []Pack) Pack {
 	var pack Pack
 	for _, p := range packs {
 		if err = p.Detect(); err != nil {
+			cli.Outf(err.Error())
```
samsalisbury: This was not printed to stdout before since all the detectors (for each pack) get run often, and this would end up with a lot of noisy output. However we do need a way to print these errors when needed.

First, I would use `cli.Logf` as opposed to `cli.Outf`, so we print the error message on stderr which is more usual for free-form messages (I tend to keep only data useful for creating pipelines on stdout).

Second, I would like to add a `pack` parameter, e.g. `sous detect [pack]` which if provided tries to detect only a single pack, and outputs all error messages for that pack.

What do you think of those suggestions @liamjbennett ?

tools/version/version.go
```diff
@@ -30,6 +29,8 @@ func NewVersion(s string) (*V, error) {
 	s = strings.TrimPrefix(s, "v")
 	s = strings.Replace(s, "x", "0", -1)
 	s = strings.Replace(s, "X", "0", -1)
+	s = strings.Replace(s, "SNAPSHOT", "rc", -1)
```
samsalisbury: Note that the `x` and `X` string replacements above are used to convert major/minor/patch to digits to conform with semver, however I think `SNAPSHOT` whilst not standard, is still valid semver when it appears in the correct place, so it may be better to keep that to follow the principle of least surprise. Also, I've just noticed that those other string replacements are really stupid and would convert "fox" to "fo0" :/ I'll fix those ones in master now with some regex.. cc @liamjbennett

core/detect.go
liamjbennett: 1) That makes sense. I will update accordingly.
2) Also a good idea - we should raise a separate Issue/PR for that.

samsalisbury: Added #3 to track this.

tools/version/version.go
samsalisbury: Added #4 to track this.

liamjbennett: Officially SNAPSHOT should be supported but it is not supported by the wmark/semver library we are using.

packs/java/app_target.go
```diff
@@ -0,0 +1,107 @@
...
+	file.TemporaryLink(t.artifactPath, "./"+localArtifact)
+	t.artifactPath = localArtifact
+}
+
+func (t *AppTarget) Dockerfile(c *core.Context) *docker.Dockerfile {
+	if t.artifactPath == "" {
+		// Actually, it is first set by compile target, then the PreDockerBuild
+		// step links it into the WD and resets artifactPath to a local, relative
+		// path.
+		t.artifactPath = "<¡ artifact path set by compile target !>"
```
stevenschlansker: I don't understand this magic string?  This looks pretty dodgy.

packs/java/app_target.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	df := &docker.Dockerfile{}
+	df.From = t.pack.baseImageTag("app")
+
+	// Since the artifact is tar.gz, and the dest is a directory, docker automatically unpacks it.
+	df.AddAdd(t.artifactPath, "/srv/app/")
+	// Pick out the contents of NPM start to invoke directly (using npm start in
```
stevenschlansker: No NPM here.

packs/java/app_target.go
```diff
@@ -0,0 +1,107 @@
...
+func (t *AppTarget) PreDockerBuild(c *core.Context) {
+	if t.artifactPath == "" {
+		cli.Fatalf("Artifact path not set by compile target.")
+	}
+	if !file.Exists(t.artifactPath) {
+		cli.Fatalf("Artifact not at %s", t.artifactPath)
+	}
+	filename := path.Base(t.artifactPath)
+	localArtifact := filename
+	file.TemporaryLink(t.artifactPath, "./"+localArtifact)
```
stevenschlansker: Does this rely on the sous working directory to be writeable?  Would be nice to not use the working directory as a scratch space, as that tends to leave files lying around if there are any problems / crashes / someone hits Ctrl-C

packs/java/compile_target.go
```diff
@@ -0,0 +1,103 @@
...
+// Checking a Java project always passes.
+func (t *CompileTarget) Check() error {
+	return nil
+}
+
+func (t *CompileTarget) Dockerfile(c *core.Context) *docker.Dockerfile {
+	df := &docker.Dockerfile{}
+	df.From = t.pack.baseImageTag(t.Name())
+	c.TemporaryLinkResource("build-prep.bash")
+	buildPrepContainerPath := "/build-prep.bash"
```
stevenschlansker: Might consider sticking to Posix `sh` rather than `bash` -- while `bash` is common it does hurt portability e.g. Debian now does not install it by default at all and uses `dash` instead.

packs/java/compile_target.go
```diff
@@ -0,0 +1,103 @@
...
+	df.AddAdd("build-prep.bash", buildPrepContainerPath)
+	df.AddRun(fmt.Sprintf("chmod +x %s", buildPrepContainerPath))
+	// This is a non-portable image, since it includes the UID of the
+	// logged-in user.
+	uid := cmd.Stdout("id", "-u")
+	gid := cmd.Stdout("id", "-g")
+	username := cmd.Stdout("whoami")
```
stevenschlansker: Can't you just call `getuid(2)` `getgid(2)` and the like, rather than shelling out?

packs/java/compile_target.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (t *CompileTarget) Dockerfile(c *core.Context) *docker.Dockerfile {
+	df := &docker.Dockerfile{}
+	df.From = t.pack.baseImageTag(t.Name())
+	c.TemporaryLinkResource("build-prep.bash")
+	buildPrepContainerPath := "/build-prep.bash"
+	df.AddAdd("build-prep.bash", buildPrepContainerPath)
+	df.AddRun(fmt.Sprintf("chmod +x %s", buildPrepContainerPath))
+	// This is a non-portable image, since it includes the UID of the
+	// logged-in user.
```
stevenschlansker: This is a serious problem, it is extremely unlikely that the users on the build machine match those available in the Docker image.

packs/java/compile_target.go
```diff
@@ -0,0 +1,103 @@
...
+	gid := cmd.Stdout("id", "-g")
+	username := cmd.Stdout("whoami")
+	// Just use the username for group name, it doesn't matter as long as
+	// the IDs are right.
+	groupAdd := fmt.Sprintf("groupadd -g %s %s", gid, username)
+	// Explanation of some of the below useradd flags:
+	//   -M means do not create home directory, which we do not need
+	//   --no-log-init means do not create a 32G sparse file (which Docker commit
+	//       cannot handle properly, and tries to create a non-sparse 32G file.)
+	userAdd := fmt.Sprintf("useradd --no-log-init -M --uid %s --gid %s %s", uid, gid, username)
```
stevenschlansker: Why not just have the launch script `su` to the `nobody` user, rather than doing this whole dance?  It seems unlikely that we want our containers to have the users of the build environment leak into them.

packs/java/compile_target.go
```diff
@@ -0,0 +1,103 @@
...
+	run := docker.NewRun(c.DockerTag())
+	run.Name = containerName
+	run.AddEnv("ARTIFACT_NAME", t.artifactName(c))
+	uid := cmd.Stdout("id", "-u")
+	gid := cmd.Stdout("id", "-g")
+	artifactOwner := fmt.Sprintf("%s:%s", uid, gid)
+	run.AddEnv("ARTIFACT_OWNER", artifactOwner)
+	artDir := t.artifactDir(c)
+	dir.EnsureExists(artDir)
+	run.AddVolume(artDir, "/artifacts")
```
stevenschlansker: Relying on volumes like this may break badly for people who use a Mac or Windows and have to use a `docker-machine` VM.

packs/java/compile_target.go
```diff
@@ -0,0 +1,103 @@
...
+// State returns any interesting state from this target to and dependent targets
+// in the build chain.
+func (t *CompileTarget) State(c *core.Context) interface{} {
+	return map[string]string{
+		"artifactPath": t.artifactPath(c),
+	}
+}
+
+func (t *CompileTarget) artifactPath(c *core.Context) string {
+	return fmt.Sprintf("%s/%s.tar.gz", t.artifactDir(c), t.artifactName(c)) //TODO: need to create this tar.gz from the jar
```
stevenschlansker: Does Go have any sort of portable path handling?  Doing path handling like this is expedient but a bit of a portability concern / antipattern.

packs/java/java.go
```diff
@@ -0,0 +1,14 @@
+package java
+
+import (
+  "encoding/xml"
+)
+
+type JavaProject struct {
```
stevenschlansker: Nit: This is really a `MavenProject` not a `JavaProject` as there are multiple viable build systems for Java.  Maybe we don't care (I personally think we should standardize on Maven)

packs/java/java.go
```diff
@@ -0,0 +1,14 @@
...
+  XMLName xml.Name `xml:"project"`
+  ModelVersion string `xml:"modelVersion"`
+  GroupId string `xml:"groupId"`
+  ArtifactId string `xml:"artifactId"`
+  Packaging string `xml:"packaging"`
```
stevenschlansker: Technically there is also a `Classifier` here which is empty in 99.9% of cases

packs/java/test_target.go
```diff
@@ -0,0 +1,90 @@
...
+func NewTestTarget(pack *Pack) *TestTarget {
+	return &TestTarget{NewJavaTarget("test", pack)}
+}
+
+func (t *TestTarget) DependsOn() []core.Target { return nil }
+
+func (t *TestTarget) RunAfter() []string { return nil }
+
+func (t *TestTarget) Desc() string {
+	return "The Java test target executes `mvn test`"
```
stevenschlansker: You might want to consider running `mvn verify` instead, which runs both tests as well as various static analysis.

packs/java/test_target.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (t *TestTarget) Dockerfile(c *core.Context) *docker.Dockerfile {
+	df := &docker.Dockerfile{}
+	df.From = t.pack.baseImageTag("test")
+	c.TemporaryLinkResource("build-prep.bash")
+	buildPrepContainerPath := "/build-prep.bash"
+	df.AddAdd("build-prep.bash", buildPrepContainerPath)
+	df.AddRun(fmt.Sprintf("chmod +x %s", buildPrepContainerPath))
+	uid := cmd.Stdout("id", "-u")
+	gid := cmd.Stdout("id", "-g")
```
stevenschlansker: This is copypasta?

tools/version/version.go
```diff
@@ -30,6 +29,8 @@ func NewVersion(s string) (*V, error) {
 	s = strings.TrimPrefix(s, "v")
 	s = strings.Replace(s, "x", "0", -1)
 	s = strings.Replace(s, "X", "0", -1)
```
stevenschlansker: A `v` prefix, and `1.0.x` placeholders are not respected in Maven AFAIK.  If you have them, the version will be treated as a literal string, so this munging may be more harmful than helpful.

packs/java/pack.go
```diff
@@ -0,0 +1,72 @@
...
+func (p *Pack) Desc() string {
+	return "Java Build Pack"
+}
+
+func (p *Pack) Detect() error {
+	if len(file.Find("pom.xml")) == 0 {
+		return fmt.Errorf("No file matching pom.xml file found")
+	}
+
+	if !file.ReadXML(&p.Project, "pom.xml") {
```
stevenschlansker: One caution here is that a `pom.xml` file is actually not an ironclad sign of having a buildable Docker application -- a Maven project could have zero or multiple built applications.  For example David's Tag service build process produces *two* entirely separate deployable applications, and our libraries have `pom.xml` files but produce none at all.

stevenschlansker:
Thanks for looking at this!  Hopefully my comments are helpful.

packs/java/app_target.go
samsalisbury: This is for human users typing "sous dockerfile". I want to ensure that the dockerfile fails builds, and at the same time provide exposition as to why it cannot be built (it relies on an artifact from another build target). Please explain "dodgy" in this context.

Please note, @stevenschlansker, that this is the third build pack written for sous, and we are still working out the details. A lot of what happens in each build pack is repeated at present. The majority of the code here will eventually be moved to a single place, and pieces like this will quite like be refactored out. We are not ready to do that refactoring right now, as not all of the requirements are known yet, so are focussing on working, rather than pretty code.

stevenschlansker: Okay, what confused me was the `<i !>` syntax, which looks like some sort of templating thing.  My misunderstanding :smile:

packs/java/app_target.go
samsalisbury: Yes, we add some symlinks to the working directory. There is of course a possibility that they don't get cleaned up in the case of a SIGKILL or other untrappable signal. However in the case of SIGINT (Ctrl-C) and SIGTERM, we have a reliable cleanup mechanism in place. Let me know if you would like me to show you how it works.

packs/java/app_target.go
```diff
@@ -0,0 +1,107 @@
...
+	t.artifactPath = artifactPath
+}
+
+func (t *AppTarget) DockerRun(c *core.Context) *docker.Run {
+	dr := docker.NewRun(c.DockerTag())
+	port0, err := ports.GetFreePort()
+	if err != nil {
+		cli.Fatalf("Unable to get free port: %s", err)
+	}
+	dr.AddEnv("PORT0", strconv.Itoa(port0))
```
stevenschlansker: Many of our Java applications require a specific number of ports set, e.g. neither Frontdoor nor Discovery can actually launch with just `PORT0` they also need a `PORT1`.

packs/java/compile_target.go
samsalisbury: I concur. However I remember using some feature that wasn't in POSIX sh when I first wrote that script... It may not be an issue any more though. I'll check at some point.

packs/java/compile_target.go
samsalisbury: Certainly could do, and probably should. I'll look into this when we eventually refactor these build packs to share common code like this.

stevenschlansker: When you revisit this, maybe this will end up being helpful: https://golang.org/src/os/proc.go

packs/java/compile_target.go
samsalisbury: Note that this is the "compile" target, it is an intermediate container used for building artifacts later injected into application containers, which are then run on Mesos. The users on the build machine absolutely do not match the users in the compile Docker container, however that container must write compiled artifacts to the build machine's filesystem, and leave them readable by the user who invoked `docker run`. Therefore it needs to write the artifacts using that user's ID.

It is not a problem that the _compile_ container is non-portable. It is just a transient container whose job is only to create artifacts. The _application_ containers which are built using these artifacts are, of course, portable.

packs/java/compile_target.go
samsalisbury: See my note above as to why this is necessary: https://github.com/opentable/sous/pull/2#discussion_r49174924 

Would the nobody user be able to write files accessible to the user who invoked `docker run` and no other user? That is the requirement, and I don't think it's achievable by any other method. If it is though, let me know, I'm not a fan of this dance myself either.

packs/java/compile_target.go
samsalisbury: This is all tested and working with `docker-machine` :) fortunately, the Docker guys realised what a massive pain it would be using volumes on a remote host, and by default they always map the client-user's `$HOME` dir into the docker VM, in such a way that volumes mounted from the home dir _just work_. I wish all their design decisions had been so thoughtful...

packs/java/compile_target.go
samsalisbury: Yes it does, that's a good point in general, I am often lazy and just use strings, assuming that we don't have any Windows developers. In fact, though, there are many aspects of sous that only work on POSIXy systems anyway, for example all the shelling out, so I think this may be better addressed along with a major update to add Windows compatibility. TBH I doubt it will ever be worth the hassle, unless Windows makes a huge resurgence. (I really don't think any OS outside of Linux/Darwin/Windows is worth considering for support right now.)

packs/java/java.go
samsalisbury: I think I prefer "MavenProject" as well, it's more specific and would leave the door open for other Java build systems in future.

tools/version/version.go
samsalisbury: Note that this is used to construct a semver, by explicitly removing the "v" prefix, and changing "x" and "X" to "0".

packs/java/compile_target.go
stevenschlansker: Oh, thanks for the explanation!  I'll think more about this and let you know if I come up with anything.

liamjbennett:
Closing in favor of https://github.com/opentable/sous-buildpacks/pull/1

PR #8, opened 2016-03-30
Added MIT license

matteofigus:
I think we generally converged to MIT for other OSS projects. Any objections?


samsalisbury:
Thanks @matteofigus !

PR #15, opened 2016-04-04
Add arch doc

nyarly:
Issuing a PR because would like review here. If there's clear answers I'm missing, let's fix those before incorporating.

PR #16, opened 2016-04-04
Fix no-upstream PR build (travis secret not available)

matteofigus:
As current seems [not to work](https://github.com/opentable/sous/pull/15)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
 script:
   - go test ./... | grep -v 'no test files'
   - ./bin/check-gofmt
-  - '[ "$TRAVIS_SECURE_ENV_VARS" = "false" ] || ./bin/publish "$TRAVIS_TAG"'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,7 +8,7 @@ install:
 script:
   - go test ./... | grep -v 'no test files'
   - ./bin/check-gofmt
-  - '[ "$TRAVIS_SECURE_ENV_VARS" = "false" ] || ./bin/publish "$TRAVIS_TAG"'
+  - '[ $TRAVIS_SECURE_ENV_VARS == false ] || ./bin/publish "$TRAVIS_TAG"'
```
samsalisbury: Thanks for the fix, the `==` is definitely right. I usually quote variables inside conditionals, as otherwise when they're empty, the script thinks there's nothing there at all, which is bad syntax. However, in this particular case I think I'd prefer to blow up hard if travis stops populating that variable! So, merging, cheers.

PR #21, opened 2016-04-08
Updating data model with deployment concept

nyarly:


sous/manifest.go
```diff
@@ -7,35 +7,27 @@ type (
 	Manifest struct {
```
nyarly: As mentioned earlier, I'm not sure about calling this thing a Manifest, but I wanted to be sure we're on the same page generally before I made that change.

nyarly:
Retracting: tests are failing

PR #22, opened 2016-04-08
Deployments restructure

samsalisbury:
This PR is the same as #21 with missing dependency added, and rebased onto master.


sous/named_version.go
```diff
@@ -0,0 +1,190 @@
...
+	} else {
+		chunk = append(chunk, first...)
+	}
+
+	for char := iter.Next(); !iter.Done() && !bytes.Equal(char, delim); char = iter.Next() {
+		chunk = append(chunk, char...)
+	}
+	chunks[0] = chunk
+
+	chunk = make([]byte, 50)
```
samsalisbury: This should probably be `chunk = []byte{}` as otherwise the `append` calls append after the first 50 bytes, leaving 50 nul bytes at the start of each chunk.

If you preallocate the slice like this, you'd need to write to specific indexes in the chunk, which is prob more trouble than it's worth in this case, and you'd prob need to trim it at the end to remove trailing nuls.

sous/named_version.go
```diff
@@ -0,0 +1,190 @@
...
+	for char := iter.Next(); !iter.Done() && !bytes.Equal(char, delim); char = iter.Next() {
+		chunk = append(chunk, char...)
+	}
+	chunks[2] = chunk
+
+	return chunks
+}
+
+func namedVersionFromChunks(source string, chunks [3][]byte) (nv NamedVersion, err error) {
+	if !(len(chunks[0]) > 0) {
```
samsalisbury: Can be simplified to `if len(chunks[0]) != 0` since by definition, len cannot return `< 0`

sous/named_version.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (err *MissingVersion) Error() string {
+	return fmt.Sprintf("No version found in %q (did find repo: %q)", err.parsing, err.repo)
+}
+
+func (err *MissingPath) Error() string {
+	return fmt.Sprintf("No path found in %q (did find repo: %q)", err.parsing, err.repo)
+}
+
+func parseChunks(source string) [3][]byte {
```
samsalisbury: I wonder if this func can be simplified by checking the input is valid UTF-8 as a guard (i.e. `utf8.ValidString(source)`), then just iterating through the runes using `for index, runeValue := range source`, and finally normalising at the point we need to convert to bytes... See http://blog.golang.org/strings for in-depth discussion of runes etc.

Just an idea, I had a quick go at fixing this in its current form, but couldn't get the delimiter detection conditional working...

sous/named_version.go
```diff
@@ -0,0 +1,190 @@
...
+	var chunks [3][]byte
+
+	var iter norm.Iter
+	chunk := make([]byte, 50)
```
samsalisbury: Should be `chunk = []byte{}` see https://github.com/opentable/sous/pull/22/commits/5711d60b65752214b114114d1aea7cad42e0696f#r59008059

sous/named_version.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	chunks[0] = chunk
+
+	chunk = make([]byte, 50)
+	for char := iter.Next(); !iter.Done() && !bytes.Equal(char, delim); char = iter.Next() {
+		chunk = append(chunk, char...)
+	}
+	chunks[1] = chunk
+
+	chunk = make([]byte, 50)
```
samsalisbury: Should be `chunk := []byte{}` see https://github.com/opentable/sous/pull/22/commits/5711d60b65752214b114114d1aea7cad42e0696f#r59008059

sous/named_version.go
```diff
@@ -0,0 +1,190 @@
...
+		err = &MissingPath{string(nv.RepositoryName), source}
+		return
+	}
+	nv.Path = Path(chunks[2])
+
+	return
+}
+
+func canonicalNameFromChunks(source string, chunks [3][]byte) (cn CanonicalName, err error) {
+	if !(len(chunks[2]) > 0) {
```
samsalisbury: Can be simplified to `if len(chunks[2]) != 0 {`

samsalisbury:
^ now correctly failing :)

PR #25, opened 2016-04-11
minor style tidy-ups

samsalisbury:
These just follow the commentary on a couple of recent commits...

### DIFF
--- a/sous/named_version_test.go
+++ b/sous/named_version_test.go
@@ -16,8 +16,7 @@ func TestParseName_NamedVersion(t *testing.T) {
 
 	nv, ok := name.(NamedVersion)
 	if !ok {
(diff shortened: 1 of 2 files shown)


sous/named_version_test.go
```diff
@@ -16,8 +16,7 @@ func TestParseName_NamedVersion(t *testing.T) {
 
 	nv, ok := name.(NamedVersion)
 	if !ok {
-		t.Errorf("Didn't parse a NamedVersion, instead: %T", nv)
-		return
+		t.Fatalf("Parsed a %T; want a  NamedVersion", nv)
```
samsalisbury: Fatalf is just shorthand for t.Errorf/t.FailNow, and FailNow just ends execution, much like return...

PR #35, opened 2016-04-28
Query ADS

nyarly:
This is the engine code to retrieve deployment data from singularity and docker registry. It's a fairly large commit, and probably warrants offline review.


cli/sous_query.go
```diff
@@ -0,0 +1,38 @@
+package cli
```
nyarly: This is basically a stub - I started writing it as a way to structure my thinking, but it doesn't actually work yet. Leaving it just to retain the work done.

cli/sous_query_adc.go
```diff
@@ -0,0 +1,36 @@
+package cli
```
nyarly: Likewise a stub

exploratory/cert_san/main.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"bytes"
+	"crypto/x509"
+	"encoding/pem"
+	"log"
+	"os"
+
+	"github.com/docopt/docopt-go"
+)
+
```
nyarly: In order to use a cert with a docker registry via an IP address, the cert needs to be signed with "SANs" (Subject Alternate Name) for the IP. This tool just inspects a certificate to verify what IP SANs are present.

exploratory/docker_labels/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/docopt/docopt-go"
+	"github.com/opentable/sous/util/docker_registry"
+)
+
+func main() {
```
nyarly: Proof of concept for the label retrieval.

sous/actual_deployment_set.go
```diff
@@ -0,0 +1,156 @@
...
+	errCh := make(chan error)
+	deps = make(Deployments, 0)
+	sings := make(map[string]*singularity.Client)
+
+	reqCh := make(chan singReq, len(singUrls)*ReqsPerServer)
+	depCh := make(chan Deployment, ReqsPerServer)
+	defer close(depCh)
+
+	regClient := docker_registry.NewClient()
+	regClient.BecomeFoolishlyTrusting()
```
nyarly: This method puts the client into a mode where it ceases to validate the certificate of the registry. Time being this was faster than repeating the song and dance to add our self-signed cert to it's root chain; in production this should be removed and we'll need to clean up some details. (there's a ticket for this change already.)

sous/deployment_builder.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+type deploymentBuilder struct {
+	target    Deployment
+	depMarker sDepMarker
+	deploy    sDeploy
+	request   sRequest
+}
+
+func (uc *deploymentBuilder) retrieveDeploy(req singReq) error {
```
nyarly: This method and the next both make HTTP requests - I considered adding goroutines for this but the data for each is required to begin the next phase of construction, and it's happening in a goroutine anyway...

sous/deployment_builder.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/opentable/sous/util/docker_registry"
+	"github.com/samsalisbury/semv"
+)
+
+type deploymentBuilder struct {
+	target    Deployment
+	depMarker sDepMarker
+	deploy    sDeploy
+	request   sRequest
+}
```
nyarly: This is an intermediary data structure used while the query is constructed - the process was complicated enough that the builder made sense to define and then use to hang the collection methods off of.

nyarly:
This is a big enough PR, I've put line comments on the interesting parts of the code. There's also a bunch of support files for the integration test environment for the query part - Dockerfiles and docker-composes and their associated configs. Then there's the godeps which are huge because we're now bringing in a docker package.

nyarly:
On the other hand, I have no idea why Travis is failing. Something about godeps, but I haven't been able to fix it.

samsalisbury:
OK, this is def looking better, thanks... I set `TEST_DOCKER_MACHINE` to my machine. The first time I ran `go test -v ./...` after this I got a panic:

    2016/05/06 13:26:59 actual_deploy_test.go:80: Panic: building registry certs failed: exit status 1

The second and subsequent times I ran that, I'm getting prompted for a password, but I'm not 100% sure which password is required, so, digging into this now.

samsalisbury:
This succeeds locally, so it doesn't seem to be the sudo that's asking for password...

    $ docker-machine ssh ot-docker sudo uptime                        
    13:55:29 up 3 days,  9:59,  1 users,  load average: 0.00, 0.03, 0.04

samsalisbury:
I finally figured out why it's asking for a password... It's trying to use `LocalDaemon` on my Mac with Docker Machine installed. Wow, that was a rabbit hole. I assume the detection is not working as expected? Will pick this up again on Monday, for now I've made a few small style tweaks, coming in a PR to your PR now

samsalisbury:
So, I've got it working now, without requiring a docker-machine called "default" (a lot of people who've been using docker machine for some time will have machines with other names). The fix ups are all over in https://github.com/nyarly/sous/pull/1

PR #37, opened 2016-05-03
Adding differ

nyarly:
The diff itself runs on a collection of channels, since the eventual
desing goal is to resolve those diffs as quickly as possible.

### DIFF
--- a/sous/deployment.go
+++ b/sous/deployment.go
@@ -19,7 +19,7 @@ type (
 		SourceVersion
 		// Owners is a map of named owners of this repository. The type of this
 		// field is subject to change.
-		Owners map[string]bool
+		Owners OwnerSet
 		// Kind is the kind of software that SourceRepo represents.
 		Kind ManifestKind
 	}
@@ -44,6 +44,8 @@ type (
 		cluster string
 		source  SourceLocation
 	}
+
+	OwnerSet map[string]struct{}
 )
 
 const (
@@ -53,10 +55,22 @@ const (
 	PassedOver                 = iota
 )
 
+func (os OwnerSet) Add(owner string) {
+	os[owner] = struct{}{}
+}
+
+func (os OwnerSet) Remove(owner string) {
+	delete(os, owner)
+}
+
+func (ds *Deployments) Add(d Deployment) {
+	*ds = append(*ds, d)
+}
+
 func BuildDeployment(m *Manifest, spec PartialDeploySpec, inherit DeploymentSpecs) (*Deployment, error) {
-	ownMap := make(map[string]bool)
+	ownMap := OwnerSet{}
 	for i := range m.Owners {
-		ownMap[m.Owners[i]] = true
+		ownMap.Add(m.Owners[i])
 	}
 	return &Deployment{
 		Cluster: spec.clusterName,
@@ -78,7 +92,7 @@ func (d *Deployment) Name() DepName {
 	}
 }
 
-func (d *Deployment) Equal(o *Deployment) bool {
+func (d *Deployment) Equal(o Deployment) bool {
 	if !(d.Cluster == o.Cluster && d.SourceVersion == o.SourceVersion && d.Kind == o.Kind && len(d.Owners) == len(o.Owners)) {
 		return false
 	}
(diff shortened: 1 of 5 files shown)


sous/deployment_diff_test.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+		Owners: map[string]bool{"judson": true, "sam": true},
+	}
+}
+
+func TestRealDiff(t *testing.T) {
+	log.SetFlags(log.Flags() | log.Lshortfile)
+	assert := assert.New(t)
+
+	intended := make(Deployments, 0)
```
samsalisbury: Can be `intended := Deployments{}` for brevity

sous/deployment_diff_test.go
```diff
@@ -0,0 +1,92 @@
...
+	existing := make(Deployments, 0)
+
+	repoOne := "https://github.com/opentable/one"
+	repoTwo := "https://github.com/opentable/two"
+	repoThree := "https://github.com/opentable/three"
+	repoFour := "https://github.com/opentable/four"
+
+	intended = append(intended, makeDepl(repoOne, 1)) //remove
```
samsalisbury: May be worth adding an `Add` method to `Deployments` (this will prob be useful outside of tests as well).

E.g. `intended.Add(makeDepl(repoOne, 1))`

Methods that mutate slice receivers should use pointer to slice so they can mutate the slice itself (slices are just values that point to a section of some underlying array).

sous/manifest.go
```diff
@@ -100,3 +100,18 @@ const (
 	// exits when it completes its task.
 	ScheduledJob = "scheduled-job"
 )
+
+func (dc *DeployConfig) Equal(o *DeployConfig) bool {
+	if dc.NumInstances != o.NumInstances || len(dc.Env) != len(o.Env) {
+		return false
+	}
+	if dc.Resources[""] != o.Resources[""] || dc.Resources[""] != o.Resources[""] || dc.Resources[""] != o.Resources[""] {
```
samsalisbury: I don't understand this line.. It's the same condition ORed three times? Also, a resource with empty name should fail validation... I think the intention should be to compare each resource (similar to how the env is compared below)...

I think it may be beneficial to add Equals methods to the `Resources` and the env collection types, and call them here.

sous/deployment.go
```diff
@@ -17,9 +17,9 @@ type (
 		Cluster string
 		// SourceID is the precise version of the software to be deployed.
 		SourceVersion
-		// Owners is a list of named owners of this repository. The type of this
+		// Owners is a map of named owners of this repository. The type of this
 		// field is subject to change.
-		Owners []string
+		Owners map[string]bool
```
samsalisbury: Bool isn't bad, but the canonical "set" representation for type T is usually `map[T]struct{}` as the empty struct is the same as saying "no information exists here" ... unless the bool is used somewhere, in which case it's fine.

sous/deployment.go
```diff
@@ -78,7 +92,7 @@ func (d *Deployment) Name() DepName {
 	}
 }
 
-func (d *Deployment) Equal(o *Deployment) bool {
+func (d *Deployment) Equal(o Deployment) bool {
 	if !(d.Cluster == o.Cluster && d.SourceVersion == o.SourceVersion && d.Kind == o.Kind && len(d.Owners) == len(o.Owners)) {
```
samsalisbury: I think this may be slightly more readable using `!=` and `||` rather than negating the whole block... Not blocking the merge though, just an observation.

sous/manifest.go
```diff
@@ -101,15 +105,30 @@ const (
 	ScheduledJob = "scheduled-job"
 )
 
-func (dc *DeployConfig) Equal(o *DeployConfig) bool {
-	if dc.NumInstances != o.NumInstances || len(dc.Env) != len(o.Env) {
+func (dc *DeployConfig) Equal(o DeployConfig) bool {
+	return (dc.NumInstances == o.NumInstances && dc.Env.Equal(o.Env) && dc.Resources.Equal(o.Resources))
```
samsalisbury: The outer parentheses are not necessary, nor harmful.. I usually leave them out, but happy if you prefer to leave them in.

PR #38, opened 2016-05-10
Rectifier

nyarly:


sous/rectifier.go
```diff
@@ -0,0 +1,564 @@
...
+*/
+
+type (
+	Rectifier struct {
+		sing RectificationClient
+	}
+
+	RectificationClient interface {
+		// Deploy(cluster, dep id, req id, docker image name, resources)
+		Deploy(string, string, string, string, Resources) error
```
samsalisbury: Henceforth, I shall prefix with "Note" any comment I make on a diff which is not blocking a merge.

Note that you can add parameter names to an interface method signature if it aids readability, as it would here. The other alternative is to use named types to document the meaning of parameters, however for interfaces intended to be called or implemented outside of the package, built-in types like you've used here are probably preferable.

You could write this method signature as:

    Deploy(cluster, depID, reqID, dockerImage string, r Resources)

Or similar. Once that's done, it's important to document the expected behaviour of the interface in the comment. This is even more important than documenting actual functions' behaviours, since at least functions have an implementation you can read to figure out what's going on.

sous/rectifier.go
```diff
@@ -0,0 +1,564 @@
...
+intendedSet := getFromManifests()
+existingSet := getFromSingularity()
+
+dChans := intendedSet.Diff(existingSet)
+
+Rectify(dChans)
+*/
+
+type (
+	Rectifier struct {
```
samsalisbury: Note that every upper-cased, exported `type`, `func`, `var` and `const` should be documented, even if the documentation says "similar to some other thing I documented earlier, except for ..."

sous/rectifier.go
```diff
@@ -0,0 +1,564 @@
...
+		sing RectificationClient
+	}
+
+	RectificationClient interface {
+		// Deploy(cluster, dep id, req id, docker image name, resources)
+		Deploy(string, string, string, string, Resources) error
+
+		// PostRequest(cluster, request id, instance count)
+		PostRequest(string, string, int) error
```
samsalisbury: I'm not sure about naming the interface methods after Singularity things (I assume that's what "request" is?). What Singularity calls a `Request` is something we need a proper name for. Looking at the [ontology] maybe we should call `Request` `Service` or `Software` or `Bucket`, or something instead? I've always disliked Singularity's overloading of "Request" sounds like an ephemeral message rather than a persistent store.

Re `Post`, this sounds very HTTP related, when I don't think that makes sense for the usage of the method. When I first saw `PostRequest` I assumed this was a HTTP client ;)

Perhaps `UpdateBucket` or something would work better?

The more I think about it, the more I wonder if we're better off not surfacing that abstraction (the difference between a `Deploy` and the thing that contains a group of related deploys) in this interface. Since a full `Deploy` object contains everything anyway, so changing the version deployed or the number of instances should both feel like just updating the deployment, not being separate steps at this level. Of course, the Singularity implementation would need to be smart and figure out how to enact different kinds of updates, but I think that's where the logic should live rather than leaking it here... What do you think?

[ontology]: https://github.com/opentable/sous/blob/master/doc/ontology.md

sous/rectifier.go
```diff
@@ -0,0 +1,564 @@
...
+	wg := &sync.WaitGroup{}
+	wg.Add(3)
+	go rect.rectifyCreates(dcs.Created, wg)
+	go rect.rectifyDeletes(dcs.Deleted, wg)
+	go rect.rectifyModifys(dcs.Modified, wg)
+
+	go func(r *sync.WaitGroup) {
+		r.Wait()
+		close(done)
+	}(wg)
```
samsalisbury: Note that it's totally fine to just rely on the closure here, since it's not inside a loop. You could write this goroutine as:

```go
go func() {
    wg.Wait()
    close(done)
}()
```
If the closure was in a loop and referred to variables changed by the loop, however, your method would be a safe way to copy a snapshot of the value into the function. (Otherwise the closed-upon variable would continue to be modified by the loop.)

sous/rectifier.go
```diff
@@ -0,0 +1,564 @@
...
+
+	return done
+}
+
+func (r *Rectifier) rectifyCreates(cc chan Deployment, done *sync.WaitGroup) {
+	defer func(c *sync.WaitGroup) { c.Done() }(done)
```
samsalisbury: Is this any different than

    defer done.Done()

? :)

sous/rectifier.go
```diff
@@ -0,0 +1,564 @@
...
+			r.sing.Deploy(pair.post.Cluster, newDepID(), computeRequestId(pair.prior), r.computeImageName(pair.post), pair.post.Resources)
+		}
+	}
+}
+
+func (r *Rectifier) computeImageName(d *Deployment) string {
+	return r.sing.ImageName(d)
+}
+
+func (r Rectifier) ChangesReq(pair DeploymentPair) bool {
```
samsalisbury: Since the receiver isn't used, you can leave out a variable name for it.

Even better, you can make this a standalone func. The only real reason to add methods that don't use the receiver is when you want to implement an interface, but for whatever reason don't need the receiver.

samsalisbury: Note that (although it may well be optimised out) adding the receiver name here means that a _copy_ of Rectifier will be placed on the stack. Rectifier is a tiny type though so it won't make any noticeable difference here.

sous/rectifier.go
```diff
@@ -0,0 +1,564 @@
...
+	return pair.prior.NumInstances != pair.prior.NumInstances
+}
+
+func (r Rectifier) ChangesDep(pair DeploymentPair) bool {
```
samsalisbury: See https://github.com/opentable/sous/pull/38/commits/9a3f462c86c3f379983cdfbcca858ce41aca4a64#r63089456

sous/rectifier.go
nyarly: In the large, I completely agree with you. The singularity terminology is awful and confusing. The worst part though is that it's not ours to change, so we're stuck with it when dealing with Singularity. It's only more confusing for us to say "a Blarg is exactly a Singularity Request, just that it's Sous's name for it."

I am 1000% on board with disregarding Singularity's terms where we need a name for something similar, and where we can close the abstraction though. I like that (iirc) Sous says "instances" and "services" as opposed to "requests" and "deploys." I wish there was a better name than "deployment," but I think it captures the idea well. 

Note for the future: I'm happy for Sous to call them "tasks" or whatever and avoid admitting that tasks and services are at all similar. What do you think there?

In the meantime, the tight coupling to the actions on Singularity was considered and intentional: this isn't really a general purpose interface; it's there to surface the Singularity web requests to the rectifier and provide a reasonable place to mock for testing. If you're working in that domain (i.e. how to talk to Singularity), not having to translate from e.g. UpdateBucket into a call to the (swagger defined) PostRequest action is nice.

sous/rectifier.go
nyarly: No, it's not :) Definitely an artifact of some concurrency refactoring - there was something more complicated in there until I trimmed it down, and then missed the opportunity to simplify still further.

sous/rectifier.go
nyarly: Mostly the idea was to introduce a further namespace on the method. I've since de-exported the rectifier type, so it's methods (really: this function) could similarly be de-exported.

sous/rectifier.go
samsalisbury: For some reason, none of your replies sent me notifications on GitHub, and
clicking the "view it on GitHub" link in this email told me the commits
this refers to has gone missing... Let me know if you see this message in
cast GH decides it can't use it...

Ah, that's fine, if this is an interface only for Singularity, then simply
naming the interface "SingularityRectificationClient" or similar removes
any need for generic terminology. The name, and the fact it was an
interface rather than a concrete type, made me think this was something
more generic.

PR #41, opened 2016-05-17
Storage

samsalisbury:
This adds the "hy" (hierarchical yaml) util, and uses it to read and write Sous state (i.e. definitions and manifests).

Sorry it's a big diff, happy to walk you through it or answer questions ad hoc @nyarly


cli/graph.go
```diff
@@ -43,9 +43,6 @@ type (
 	LocalGitClient struct{ *git.Client }
 	// LocalGitRepo is the git repository containing WorkDir.
 	LocalGitRepo struct{ *git.Repo }
-	// LocalGitContext is the git context snapshot of the user when they invok
-	// Sous.
-	LocalGitContext struct{ *git.Context }
```
samsalisbury: This is incidental cleanup of unused type.

cli/sous_init.go
```diff
@@ -0,0 +1,52 @@
...
+					Resources:    sous.Resources{},
+					Env:          map[string]string{},
+					NumInstances: 3,
+				},
+				Version: v,
+			},
+		},
+	}
+	return SuccessYAML(m)
+}
```
samsalisbury: Hmm, this branch was open too long, and picked up some unrelated stuff. I'm happy to remove this stub init command if you prefer, or leave it in. Let me know.

All it does is print to stdout an auto-configured manifest for the app in the current dir.

ext/git/context.go
```diff
@@ -0,0 +1,82 @@
...
+		return ""
+	}
+	if primaryRemote.FetchURL == "" {
+		return ""
+	}
+	// We don't care about this error, empty string is
+	// an acceptable return value.
+	u, _ := CanonicalRepoURL(primaryRemote.FetchURL)
+	return u
+}
```
samsalisbury: New file because I split up the git stuff into more manageable chunks.

samsalisbury: Incidental work, again.

ext/git/remote.go
```diff
@@ -0,0 +1,55 @@
...
+		p := strings.SplitN(path, ":", 2)
+		host = p[0]
+		path = "/" + p[1]
+	}
+	if strings.ContainsRune(host, '@') {
+		p := strings.SplitN(host, "@", 2)
+		host = p[1]
+	}
+	return host + strings.TrimSuffix(path, ".git"), nil
+}
```
samsalisbury: Again, incidental rearrangement.

ext/git/repo.go
```diff
@@ -1,12 +1,5 @@
 package git
```
samsalisbury: Incidental rearrangement of the git external stuff...

sous/manifest.go
```diff
@@ -1,14 +1,16 @@
...
+
 	"fmt"
 
 	"github.com/samsalisbury/semv"
 )
 
 type (
 	// Manifests is a collection of Manifest.
-	Manifests []*Manifest
+	Manifests map[string]*Manifest
```
samsalisbury: Necessary for saving to multiple files via hy.

samsalisbury:
The test failure is genuine, fixing it now @nyarly

util/hy/context.go
```diff
@@ -0,0 +1,202 @@
...
+	"strings"
+)
+
+type (
+	ctx struct {
+		path      string
+		unmarshal func([]byte, interface{}) error
+		marshal   func(interface{}) ([]byte, error)
+	}
+	walkFunc func(name, tag string, val reflect.Value) (*target, error)
```
nyarly: Just a note: Looking at readTarget and writeTarget, I suspect that the more legible solution here would be to use a Visitor interface { fileTarget, dirTarget, treeTarget }, and do the string manipulation on the vehicle side of the visitor pattern.

nyarly:
First of all, wow. This is a job of work done, for sure.

Testing my understanding here: it looks at if sous/state.go needs to be a updated further - the Clusters type became a map, but the Defs>Clusters field (I think?) should be `hy:"clusters/**"` and possibly Cluster>Name should be dropped or else the `hy:",filename"` be removed.

My sense is that it there's some repetition there (i.e. the clusters map includes the cluster names and the cluster object does as well). My experience is that the latter will be a little clearer to read, but the "nickname" from the file name is a potential source of confusion.

It seems like we should have a round-trip test on the actual Sous state structures so that we've got that confidence.

I made a line comment in hy/context.go - it looks to me like this is a big chunk of code we'll be able to treat as a black box for a long time to come, so I'm not sure how much value there is to refactoring in here. That said, it seems to me like actually building visitor objects with tree methods instead of visitor functions with a 3 arm switch would be a little more readable.

This is also a good place for me to ask: the "all the files in the directory are the package" thing that Go does leads to some confusion on my part as to how to arrange code into files. I'd've been stricter keeping methods with their types (i.e. all the target methods in target.go), with loose functions near their callers. Likewise, a readerVisitor would become its own file (and by no means have I been good about this in my own code so far...) Conversely, I can see how all functions related to e.g. marshalling are sensibly adjacent, regardless of their receivers. Do you have opinions or rules of thumb on this?

The follow-on and meta-commentary to that is that I certainly feel like that kind of fussy code structuring can make intention clearer and that smaller files are easier to follow, all things considered. Some thing to fill the cavernous slack time in our sprints with... :)

All that said, this remains mere commentary. I'd definitely like to see the state marshalling tests come whenever we start to work with the actual state objects, but I'm happy with this code as it is and would rather save improvements for future iterations.

util/hy/context.go
samsalisbury: Yeah I think you're right.. This could definitely do with some refactoring, I knew I was duplicating work there as I did it, but then left it alone once it worked. I'll come back to this eventually; hy is a case of a lib that should be standalone eventually, so I'll leave this the way it is until I get round to breaking it out.

samsalisbury:
> Testing my understanding here: it looks at if sous/state.go needs to be a updated further - the Clusters type became a map, but the Defs>Clusters field (I think?) should be hy:"clusters/**" and possibly Cluster>Name should be dropped or else the hy:",filename" be removed.

Yeah, this needs cleaning up. I did not implement the `,filename` tag yet, but for sure we would only want one or the other. I'll work out how to sort out this today.

samsalisbury:
> It seems like we should have a round-trip test on the actual Sous state structures so that we've got that confidence.

This file does test round-tripping of sous state implicitly by testing both marshaling and unmarshaling against a known fs tree: https://github.com/opentable/sous/blob/master/ext/storage/storage_test.go

samsalisbury:
> I made a line comment in hy/context.go - it looks to me like this is a big chunk of code we'll be able to treat as a black box for a long time to come, so I'm not sure how much value there is to refactoring in here. That said, it seems to me like actually building visitor objects with tree methods instead of visitor functions with a 3 arm switch would be a little more readable.

I concur, see reply to line comment.

samsalisbury:
> This is also a good place for me to ask: the "all the files in the directory are the package" thing that Go does leads to some confusion on my part as to how to arrange code into files. I'd've been stricter keeping methods with their types (i.e. all the target methods in target.go), with loose functions near their callers. Likewise, a readerVisitor would become its own file (and by no means have I been good about this in my own code so far...) Conversely, I can see how all functions related to e.g. marshalling are sensibly adjacent, regardless of their receivers. Do you have opinions or rules of thumb on this?

I don't have any hard and fast rules about this. I tend to keep methods that interact with each other close by (same file), although there's definitely benefit to having all methods on a type in the same file too. I haven't had issues navigating code under either scheme which is probably why I'm not strict about it, but happy to try following a rule if it makes code easier to read for others. One thing I don't do is have a separate file for every type declaration, but rather group these according to some aggregate root. Otherwise we'd have an explosion of files which would be more difficult to use I think.

samsalisbury:
> All that said, this remains mere commentary. I'd definitely like to see the state marshalling tests come whenever we start to work with the actual state objects, but I'm happy with this code as it is and would rather save improvements for future iterations.

By all means expand the tests in https://github.com/opentable/sous/blob/master/ext/storage/storage_test.go they're very simple right now, but marshalling and unmarshalling is tested there.

PR #44, opened 2016-05-20
DCOPS-7200 - NameCache is a docker manifest search

nyarly:
This also includes big extensions to the the rectifier (including a working singularity client(!)),
in particular an errors channel to report issues while rectifying.

### DIFF
--- a/sous/image_mapping.go
+++ b/sous/image_mapping.go
@@ -1,0 +1,110 @@
+package sous
+
+import (
+	"fmt"
+
+	"github.com/opentable/sous/util/docker_registry"
+)
+
+type (
+	NameCache struct {
+		DockerNameLookup
+		SourceNameLookup
+	}
+
+	SourceRecord struct {
+		SourceLocation SourceLocation
+		etag           string
+	}
+
+	ImageName string
+
+	SourceNameLookup map[ImageName]SourceRecord
+	DockerNameLookup map[SourceLocation]ImageName
+
+	NoImageNameFound struct {
+		SourceLocation
+	}
+
+	NoSourceLocationFound struct {
+		ImageName
+	}
+)
+
+var theNameCache = NameCache{
+	make(DockerNameLookup),
+	make(SourceNameLookup),
+}
+
+func (e NoImageNameFound) Error() string {
+	return fmt.Sprintf("No image name for %v", e.SourceLocation)
+}
+
+func (e NoSourceLocationFound) Error() string {
+	return fmt.Sprintf("No image name for %v", e.ImageName)
+}
+
+func (dl DockerNameLookup) GetImageName(sl SourceLocation) (ImageName, error) {
+	if in, ok := dl[sl]; ok {
+		return in, nil
+	} else {
+		return "", NoImageNameFound{sl}
+	}
+}
+
+func (sn SourceNameLookup) getSourceRecord(in ImageName) (SourceRecord, error) {
+	if sr, ok := sn[in]; ok {
+		return sr, nil
+	} else {
+		return SourceRecord{}, NoSourceLocationFound{in}
+	}
+}
+
+func (sn SourceNameLookup) GetSourceLocation(in ImageName) (SourceLocation, error) {
+	if sr, ok := sn[in]; ok {
+		return sr.SourceLocation, nil
+	} else {
+		return SourceLocation{}, NoSourceLocationFound{in}
+	}
+}
+
+func (dl DockerNameLookup) InsertDockerName(sl SourceLocation, in ImageName) error {
+	dl[sl] = in
+	return nil
+}
+
+func (sn SourceNameLookup) InsertSourceLocation(in ImageName, sl SourceLocation, etag string) error {
+	sn[in] = SourceRecord{sl, etag}
+	return nil
+}
+
+func (nc *NameCache) Insert(sl SourceLocation, in ImageName, etag string) error {
+	err := nc.InsertSourceLocation(in, sl, etag)
+	if err != nil {
+		return err
+	}
+
+	err = nc.InsertDockerName(sl, in)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func GetSourceLocation(dr docker_registry.Client, in ImageName) (SourceLocation, error) {
+	sr, err := theNameCache.getSourceRecord(in)
+
+	newSL, etag, err := retreiveSourceLocation(in, sr.etag)
+	if exists, ok := err.(NotModifiedErr); ok {
+		return sr.SourceLocation, nil
+	}
+	if err != nil {
+		return SourceLocation{}, err
+	}
+
+	if sr.SourceLocation != newSL {
+		theNameCache.Insert(newSL, in, etag)
+	}
+	return newSL, nil
+}
--- a/sous/image_mapping.go
+++ b/sous/image_mapping.go
@@ -7,20 +7,17 @@ import (
 )
 
 type (
+	// NameCache is a primative database for looking up SourceVersions based on Docker image names and vice versa.
 	NameCache struct {
+		registryClient docker_registry.Client
 		DockerNameLookup
 		SourceNameLookup
 	}
 
-	SourceRecord struct {
-		SourceVersion SourceVersion
-		etag          string
-	}
-
 	ImageName string
 
-	SourceNameLookup map[ImageName]SourceRecord
-	DockerNameLookup map[SourceVersion]ImageName
+	SourceNameLookup map[ImageName]*SourceRecord
+	DockerNameLookup map[SourceVersion]*SourceRecord
 
 	NotModifiedErr struct{}
 
@@ -31,12 +28,11 @@ type (
 	NoSourceVersionFound struct {
 		ImageName
 	}
-)
 
-var theNameCache = NameCache{
-	make(DockerNameLookup),
-	make(SourceNameLookup),
-}
+	SourceRecord struct {
+		md docker_registry.Metadata
+	}
+)
 
 func (e NoImageNameFound) Error() string {
 	return fmt.Sprintf("No image name for %v", e.SourceVersion)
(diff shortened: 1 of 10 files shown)


nyarly:
Of particular note: I made a conscious decision not to build further singularity integration tests here - next steps are to build  the rectification driver (pull manifest data, build deployment sets, feed them to the differ, let rectification happen) which feels like the right place to build integration tests at.

sous/rectifier.go
```diff
@@ -43,66 +43,257 @@ type (
...
+}
+
+func (e *ChangeError) IntendedDeployment() *Deployment {
+	return e.Deployments.post
+}
+
+func Rectify(dcs DiffChans, s RectificationClient) (chan RectificationError, chan struct{}) {
 	rect := rectifier{s}
 	done := make(chan struct{})
+	errs := make(chan RectificationError)
```
nyarly: One especially important observation:

(Testing my understanding) Defined this way, the errs channel *must* be drained by the caller of Rectify, or the first error will block the appropriate goroutine.

Alternatives would be to define errs with a size (which would kick the can down the road - IMO, we'd just hit the blocking behavior later - and therefore further from this moment of understanding) or to spin off a goroutine for each rectification, aot to just each rectification channel - so that if the errs channel isn't drained, the errored rectifications would block instead of the whole channel. In the case that the errs channel is ignored, we'd leak goroutines, which probably isn't terrible... Still, none of these options thrill me.

util/docker_registry/raw_client.go
```diff
@@ -0,0 +1,713 @@
+package docker_registry
```
nyarly: This file is kind of the rotting carcass of a copy of a file from docker/distribution. It's here mostly for reference and to make clear what code from the existing registry client hasn't yet found a home in Sous - we may not ever need more than this, so maybe it's best simply to delete this.

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// GetSourceVersion retreives a source version for an image name, updating it from the server if necessary
+// Each call to GetSourceVersion implies an HTTP request, although it may be abbreviated by the use of an etag.
+func GetSourceVersion(in string) (SourceVersion, error) {
+	return theNameCache.GetSourceVersion(in)
+}
+
+// InsertContainerRecord stores a SourceVersion/image name pair into the global name cache
+func InsertContainerRecord(sv SourceVersion, in, etag string) error {
```
samsalisbury: "Container" should be "Image" I think.. Will send a follow-up PR to address this.

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+	return theNameCache.Insert(sv, in, etag)
+}
+
+func GetContainerName(sv SourceVersion) (string, error) {
```
samsalisbury: "Container" -> "Image" will add this to PR

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+		Labels:        sv.DockerLabels(),
+	}}
+
+	return nc.insertRecord(&record)
+}
+
+func (dl DockerNameLookup) GetImageName(sv SourceVersion) (string, error) {
+	if sr, ok := dl[sv]; ok {
+		return sr.md.CanonicalName, nil
+	} else {
```
samsalisbury: [Redundant else](https://golang.org/doc/effective_go.html#control-structures), will address in follow-up PR

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+		return "", NoImageNameFound{sv}
+	}
+}
+
+func (nc *NameCache) GetSourceVersion(in string) (SourceVersion, error) {
+	sr, err := nc.getSourceRecord(ImageName(in))
```
samsalisbury: unchecked error

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+
+	oldSV, err := sr.SourceVersion()
```
samsalisbury: unchecked error, use `_` if this is intended, assuming we should give up on error for this and the last error case, will add to PR

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+	nsv, nerr := newSR.SourceVersion()
+	osv, oerr := sr.SourceVersion()
+	if newSR.md.CanonicalName == sr.md.CanonicalName || (nerr == nil && oerr == nil && nsv == osv) {
+		newSR.md.AllNames = union(newSR.md.AllNames, sr.md.AllNames)
+	}
+	nc.insertRecord(&newSR)
+	return newSR.SourceVersion()
+}
+
+func (sn SourceNameLookup) GetCanonicalName(in string) (string, error) {
```
samsalisbury: Not specific to this changeset, but we need to drop "CanonicalName" for "SourceLocation" everywhere. I'll send a PR for this.

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+	if sr, ok := sn[ImageName(in)]; ok {
+		return sr.md.CanonicalName, nil
+	} else {
```
samsalisbury: redundant else

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func (nc *NameCache) insertRecord(sr *SourceRecord) error {
+	err := nc.insertSourceVersion(sr)
+	if err != nil {
+		return err
+	}
+
+	err = nc.insertDockerName(sr)
```
samsalisbury: Can be `return nc.insertDockerName(sr)` to shorten the method.

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	for _, s := range right {
+		set[s] = struct{}{}
+	}
+
+	res := make([]string, len(set))
+
+	for k, _ := range set {
+		res = append(res, k)
```
samsalisbury: Does this work as expected? Looks like you'll have `len(set)` empty strings, followed by `len(set)` expected strings...

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	return res
+}
+
+func (sn SourceNameLookup) getSourceVersion(in string) (SourceVersion, error) {
+	if sr, ok := sn[ImageName(in)]; ok {
+		return sr.SourceVersion()
+	} else {
```
samsalisbury: redundant else

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+		return SourceVersion{}, NoSourceVersionFound{ImageName(in)}
+	}
+}
+
+func (sn SourceNameLookup) getSourceRecord(in ImageName) (*SourceRecord, error) {
+	if sr, ok := sn[in]; ok {
+		return sr, nil
+	} else {
```
samsalisbury: redundant else

sous/image_mapping_test.go
```diff
@@ -0,0 +1,86 @@
...
+	assert := assert.New(t)
+
+	mds := make(mdChan, 10)
+	nc := NameCache{
+		&DummyRegistryClient{mds},
+		make(DockerNameLookup),
+		make(SourceNameLookup),
+	}
+
+	v, _ := semv.Parse("1.2.3")
```
samsalisbury: Can use `semv.MustParse` to panic instead of returning error.

sous/image_mapping_test.go
```diff
@@ -0,0 +1,86 @@
...
+	sv := SourceVersion{
+		Version:    v,
+		RepoURL:    RepoURL("https://github.com/opentable/wackadoo"),
+		RepoOffset: RepoOffset("nested/there"),
+	}
+	base := "docker.repo.io/wackadoo"
+	in := string(append([]byte(base), ":version-1.2.3"...))
```
samsalisbury: Can use `in := base + ":version-1.2.3"` rather than going to bytes first.

sous/image_mapping_test.go
```diff
@@ -0,0 +1,86 @@
...
+	digest := "sha256:deadbeef1234567890"
+	nc.Insert(sv, in, digest)
+
+	cn, err := nc.GetCanonicalName(in)
+	if assert.Nil(err) {
```
samsalisbury: Just wondering how this is different from `if err == nil`? Guessing it panics or immediately fails the test? If so, wrapping in `if` wouldn't do anything, if not, would prefer plain Go comparison.

I can see that testify can make some tests shorter to write, but sometimes I find plain Go easier to read...

samsalisbury: Hmm, reading the docs I think it fails the test by adding an error, but does not halt execution of the test. That seems reasonable.

sous/rectifier.go
samsalisbury: Before reading the whole method, yes I think you're right... An unbuffered channel (i.e. one like this with no explicit size set, equivalent to size zero), will block any goroutine trying to write to it until another goroutine tries to read. If multiple goroutines are trying to write at the same time, when another gr reads, one of the writing grs is picked _at random_ to write to the channel and continue, whilst the other gr is still blocked. Each read from the channel picks a random goroutine to write/continue until they're all unblocked.

So, being unbuffered here means that any goroutine that errs and tries to write to this channel will be blocked, but others not erring will continue.

sous/rectifier.go
```diff
@@ -43,66 +43,257 @@ type (
...
+
+func (e *ChangeError) ExistingDeployment() *Deployment {
+	return e.Deployments.prior
+}
+
+func (e *ChangeError) IntendedDeployment() *Deployment {
+	return e.Deployments.post
+}
+
+func Rectify(dcs DiffChans, s RectificationClient) (chan RectificationError, chan struct{}) {
```
samsalisbury: I'm not sure channels are the best return types for this call, unless there's some clever optimisation I'm not seeing. Assuming we want to run a rectification as a single action (on some human time-scale in the order of seconds/minutes), then make decisions afterwards based on the errors, I would probably favour just returning a slice of error, whose length can be checked (both zero length slice and nil have length zero) to see if an error occurred. That would certainly be simpler for the caller to handle, at least.

sous/rectifier.go
```diff
@@ -43,66 +43,257 @@ type (
...
 	rect := rectifier{s}
 	done := make(chan struct{})
+	errs := make(chan RectificationError)
 	wg := &sync.WaitGroup{}
 	wg.Add(3)
-	go rect.rectifyCreates(dcs.Created, wg)
-	go rect.rectifyDeletes(dcs.Deleted, wg)
-	go rect.rectifyModifys(dcs.Modified, wg)
+	go rect.rectifyCreates(dcs.Created, errs, wg)
```
samsalisbury: Rather than pass the WaitGroup around, I would favour wrapping these calls in anonymous functions, which call done inside this method. That should make the synchronisation flow a bit easier to read.

sous/rectifier.go
```diff
@@ -43,66 +43,257 @@ type (
...
 	go func(r *sync.WaitGroup) {
+		log.Println("")
 		r.Wait()
+		log.Println("")
 		close(done)
+		log.Println("")
 	}(wg)
 
-	return done
+	return errs, done
```
samsalisbury: Returning the done channel along with the errors channel gives the caller a lot of information to deal with. If we did keep this signature (i.e. returning channels), then we don't really need the done channel, we can simply close the error channel since we already know that no more errors can be written there after waiting for the WaitGroup. Reads from a closed channel always succeed, getting the zero type for the channel (nil in the case of error).

However, I think we will end up changing the method signature to return a slice of error, but just something to bear in mind.

sous/rectifier.go
samsalisbury: Re leaking goroutines: this is actually one of the most common types of leak in Go, and whilst it can be self-limiting in cases where those goroutines terminate, it's always best to avoid them.

sous/rectifier.go
samsalisbury: E.g.

```go
go func() { rect.rectifyCreate(dcs.Created, errs); wg.Done() }()
```

sous/rectifier.go
```diff
@@ -43,66 +43,257 @@ type (
...
+		return cl
+	}
 }
 
-func (r *rectifier) rectifyCreates(cc chan Deployment, done *sync.WaitGroup) {
+func (r *rectifier) rectifyCreates(cc chan Deployment, errs chan RectificationError, done *sync.WaitGroup) {
 	defer done.Done()
+	log.Println("")
 	for d := range cc {
+		log.Println("")
```
samsalisbury: I've noticed a lot of empty log lines, were you intending to add messages?

samsalisbury:
Was going to merge this and send a subsequent PR to resolve issues, but I think there's enough going on here that it's worth talking about it first :)

sous/image_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+	set := make(map[string]struct{})
+	for _, s := range left {
+		set[s] = struct{}{}
+	}
+
+	for _, s := range right {
+		set[s] = struct{}{}
+	}
+
+	res := make([]string, len(set))
```
samsalisbury: `res := make([]string, 0, len(set))` would set capacity instead of length, I think that's what you meant?

sous/image_mapping.go
nyarly: In this case, this *is not* the SourceLocation. Docker containers can be referred to by multiple names. e.g. anyone can add a new tag for any image at any time, and latest is always changing... The v2 docker registry design (this may be true of v1 as well) treats the digest-indexed reference as preferred and their code and documentation refers to <host>/<reponame>:<digest> form as the "canonical name", which is reasonable since the digest-reference form is guaranteed to uniquely and immutably refer to a particular instance. 

The docker lookup responses includes the digest form, so this is cheap to include, and in fact moving requests towards the canonical form as much as possible is probably the best idea. 

For instance, a running instance should probably be running from a digest-reference docker image, not from a tagged one - using tags to track deployment intent might make sense as a practice, but it should be Sous's prerogative to select images, not the happenstance of tags and when Singularity boots an instance.

sous/image_mapping_test.go
nyarly: Yes - it's both a test for nil (and a test fail if it's not nil) and returns a boolean so that the `if` is useful. Since the guarded test doesn't make sense in the presence of an error...

samsalisbury:
Looks great, thanks

PR #54, opened 2016-06-15
Storage: correct hy tags and test data.

samsalisbury:
- Using flattened defs file containing all non-manifest data

See comments in DCOPS-7430


nyarly:
I think this merits more discussion. I'll follow up on Jira

nyarly:
Merging this, but I'm going to push the Jira ticket back into the backlog.

PR #58, opened 2016-06-22
Implements the "harvest" feature

nyarly:
All requests now go through the name cache. All running Source Versions are
therefore pulled into the cache by doing a `sous query adc` (or during
rectify).

Docker repos are paired with source locations as they're seen. If a request
is made for the docker name of an unknown source version, we can back up to
its source location, find related repos (almost always just one), pull all
those tags and populate the cache from the registry.


samsalisbury:
Thanks, I think we may revisit the need for name cache in future, but in the short term this will get us moving with deployments.

PR #61, opened 2016-06-23
Single manifest

nyarly:
This is effectively "yes merge, but..."

Two variations from your original, @samsalisbury - I created a type for the predicate function, and changed the Deployments.Filter function to return the Deployments itself if the function is nil. Mostly a little efficiency, but I can see the argument for returning a copy. Let me know what you think.

### DIFF
--- a/cli/sous_rectify.go
+++ b/cli/sous_rectify.go
@@ -14,7 +14,8 @@ type SousRectify struct {
 	Config       LocalSousConfig
 	DockerClient LocalDockerClient
 	flags        struct {
-		dryrun string
+		dryrun,
+		manifest string
 	}
 }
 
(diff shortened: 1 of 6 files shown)


nyarly:
One actually practical change here: you (mostly correctly, I think) put a defer close(errs) at the toplevel for Rectify - but Rectify uses the close of the errs channel to signal the completion of work, so it was closing errs as well, which led to a close-of-closed panic.

Still,closing the channel where it was made makes sense, so I've pushed the errs channel into the Rectify function, and return it as opposed to receiving it as an argument.

Upshot is: if nothing else, CI passes here.

samsalisbury:
Thanks for fixing, somehow I missed that failure when running tests locally.

Re returning the `Deployments` itself for nil filter, I like that change. The important thing is that you changed the receiver to not be a pointer any more, so you can always get a new `Deployments` instance returned. Having inconsistency there (sometimes a copy, sometimes the same object) would have been bad, which is why I wrote it that way initially, but your version is way cleaner.

Re strongly-typed predicate type: I'm in two minds about this because the original method signature is so short, and contains a lot more information than an opaque type name. I think I preferred it as a bare `func(*Deployment) bool` than `DeploymentPredicate`. Yes, it's 3 characters longer, but I think the extra exposition it gives makes this worth it... Or am I missing some other benefit of the strong type?

Re closing the channel near where it's opened, I find that much easier to follow, cheers.

PR #63, opened 2016-06-29
Bug was: the docker registry client didn't include the hostname

nyarly:
Result was that we tried to pull OT images from dockerhub.

This fixes that.


samsalisbury:
This is still failing, `opentable/two no longer deployed after resolve` ... I'll take a look this morning and see if I can figure out what's wrong.

nyarly:
On the trail of the integration fail - got it working here, and the result will be a lot more solid. In the meantime it's not going to work so...

PR #65, opened 2016-07-01
Build makes both :x.y.z version and :123abcdef revision tags

nyarly:
The main result here is that the build be "pinned" at the registry, so that its
digest will continue to be available.

This is essentially a docker implementation detail, and the tags being pushed
should properly be the responsiblity of a docker registry package

The requirement Sous is making of the registry component is that the deploy
artifacts have a stable, immutable identifier: that there is a name for the
artifact that will change if the artifact does, and that will remain in the
registry for as long as this clean build is needed by Sous.

The Docker Registry provides the former (digest image names change whenever a
new image is pushed), but doesn't guarantee the latter (the digests become
unavailable if no tag name refers to the image anymore.) So the detail here is
to use a revision tag exclusively for pinning purposes.


samsalisbury:
Thanks, I'm holding off merging this until I've got the rearrangement finished, at that point I will apply these changes to the refactored codebase.

PR #66, opened 2016-07-05
Sketch of docs about interfaces

nyarly:
Ultimately, this should document the required interfaces and functionality of 
future container image formats, image registry and cluster services.

In the meantime, I'm hoping to address the content-addressibility issue and the 
search cache. I feel like we're at an impasse there, and I'd like to find a 
common understanding of the issues involved.

### DIFF
--- a/doc/cluster_and_container_iface.md
+++ b/doc/cluster_and_container_iface.md
@@ -1,0 +1,108 @@
+# Cluster and Container Interface
+
+Sous uses an abstraction of a running service called a Source Version.
+This incorporates
+the official repository for the source code,
+and offset within that repository (usually a relative path from the root)
+the [version](http://semver.org) of the software
+and the revision of the software.
+The Source Versions are composed out of the manifests in the GDM
+to determine what should be deployed where.
+
+Sous has to be able to record and recover metadata sufficient to construct a source version
+from both the running cluster
+and from the container registry.
+
+Sous maintains the state of deployments by comparing
+the data in the container registry
+and the running clusters.
+It therefore must be able to make an accurate comparisons
+between what's actually running
+and the images available in the registry.
+
+## An implementation
+
+Using Docker and Singularity as an example,
+Docker allows for labels on its images.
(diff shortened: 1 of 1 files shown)


samsalisbury:
Hi, thanks for doing this, we definitely need to get these concepts solid v soon. I've not spent enough time digesting this today, but from a single read-through, I think it's in the right direction. I'll reply properly/suggest edits tomorrow.

samsalisbury:
OK, replies below. I can't merge this yet, still fighting on the other branch, but here are my comments...

> While meaningful image names were considered, there are a number of drawbacks. First, changes to design in Sous would mean changes in the format in the names. (Note that as of this writing, the idea of "advisory metadata" is being considered, which would need to be recorded in the image name somehow.)

The first issue (changes to Sous) isn't a reason not to do this to my mind, we would have options to either write a migration for these names, introduce backwards-compatibility, or as a last resort simply re-build everything considered "current".

Advisory metadata may not need to be part of the image _name_, assuming we can flag to the user early that it's there (so they don't expect to deploy it to prod), and that production would refuse to deploy it anyway.

As an aside, quite apart from docker repositories and tags, we should also remember we can use the "username" path component of docker repositories. E.g. images built with no advisories could be prefixed `production/`those for ci `ci/` etc. This would not be a failsafe (production would still need to check the labels before launching) but would provide a convenient separation between different image kinds in the absence of an attacker/joker who decides to try and be clever and deploy to prod using this method.

> Second, Docker does not itself attempt to provide reproducible builds, and even if it did, no platform we know of provides reproducibility out of the box. Consequently, builds performed on separate nodes (a decided possibility with a globally distributed Sous system) would appear to be same image, which would could result in deployments of a service in some clusters behaving differently than others, with no indication that they weren't identical.

This is true, and I agree we do need to store the image digest for this reason, probably inside the server-managed manifest. However we also need to be able to look images based on broader information about the build environment they were built on, and possibly trust that 2 different images built in strikingly similar environments are highly likely not to diverge on behaviour.

PR #67, opened 2016-07-06
Split docker singularity

samsalisbury:


nyarly:
Added to this in #68, comments are there.

Found some issues with how this was processing against a fresh compose, but am now getting a much different error in travis. c.f.

PR #68, opened 2016-07-06
Companion to #67

nyarly:
This is only a PR so that I can get Travis to run against it


ext/docker/image_mapping_test.go
```diff
@@ -1,25 +1,42 @@
...
+	}
+	return db
+}
+
 func TestRoundTrip(t *testing.T) {
 	assert := assert.New(t)
 
 	dc := docker_registry.NewDummyClient()
-	nc := NewNameCache(dc, "sqlite3", InMemoryConnection("roundtrip"))
+	nc := NewNameCache(dc, inMemoryRoundtripDB())
```
nyarly: Every test that uses an in memory database *has* to have a unique database, so it really just makes sense to have a single function for this.

ext/docker/image_mapping.go
```diff
@@ -175,11 +167,11 @@ func (nc *NameCache) GetCanonicalName(in string) (string, error) {
 }
 
 // Insert puts a given SourceVersion/image name pair into the name cache
-func (nc *NameCache) Insert(sv SourceVersion, in, etag string) error {
+func (nc *NameCache) Insert(sv sous.SourceVersion, in, etag string) error {
```
nyarly: This method really should get removed - I don't think that change makes sense for this PR, but wanted to call out that the NameCache is spitting distance from being exactly that: only getting data as a side effect of responding to queries for SV or Artifacts

ext/docker/image_mapping.go
```diff
@@ -145,7 +137,7 @@ func (nc *NameCache) GetSourceVersion(in string) (SourceVersion, error) {
 }
 
 // GetImageName returns the docker image name for a given source version
-func (nc *NameCache) GetImageName(sv SourceVersion) (string, error) {
+func (nc *NameCache) GetImageName(sv sous.SourceVersion) (string, error) {
```
nyarly: This, seems to me, should 100% not be exported.

ext/docker/image_mapping.go
```diff
@@ -17,8 +17,8 @@ type (
 	// NameCache is a database for looking up SourceVersions based on
 	// Docker image names and vice versa.
 	NameCache struct {
-		registryClient docker_registry.Client
-		db             *sql.DB
+		RegistryClient docker_registry.Client
+		DB             *sql.DB
```
nyarly: Why is this exported? Users of the NameCache shouldn't ever manipulate the cache DB themselves.

ext/singularity/rectifier.go
```diff
@@ -0,0 +1,159 @@
...
+
+		// DeleteRequest instructs Singularity to delete a particular request
+		DeleteRequest(cluster, reqID, message string) error
+
+		//ImageLabels finds the (sous) docker labels for a given image name
+		ImageLabels(imageName string) (labels map[string]string, err error)
+	}
+
+	// DTOMap is shorthand for map[string]interface{}
+	DTOMap map[string]interface{}
```
nyarly: DTOMap is a kludge around the way the generated DTO objects work for go-singularity. Maybe there's a better way to approach this, but again I don't think this type should be exported.

integration/integration_test.go
```diff
@@ -224,18 +253,11 @@ func findRepo(deps sous.Deployments, repo string) int {
...
-func manifest(nc sous.ImageMapper, drepo, containerDir, sourceURL, version string) *sous.Manifest {
-	//	sv := sous.SourceVersion{
-	//		RepoURL:    sous.RepoURL(sourceURL),
-	//		RepoOffset: sous.RepoOffset(""),
-	//		Version:    semv.MustParse(version),
-	//	}
-
-	in := buildImageName(drepo, version)
-	buildAndPushContainer(containerDir, in)
+func manifest(nc sous.Registry, drepo, containerDir, sourceURL, version string) *sous.Manifest {
```
nyarly: This should probably be exported since the docker tests need to make use of it. That way we can remove the dupe there.

lib/manifest.go
```diff
@@ -140,17 +140,7 @@ func (dc *DeployConfig) Equal(o DeployConfig) bool {
...
-func (r Resources) SingMap() dtoMap {
-	return dtoMap{
-		"Cpus":     r.cpus(),
-		"MemoryMb": r.memory(),
-		"NumPorts": int32(r.ports()),
-	}
-}
-
-func (r Resources) cpus() float64 {
+func (r Resources) Cpus() float64 {
```
nyarly: Ugh. This is where we get into trouble. Independence from the scheduler is going to be hard because of the intimate connection between GDM configuration and the capabilities of the scheduler.

lib/registry.go
```diff
@@ -0,0 +1,12 @@
+package sous
+
+type (
+	Registry interface {
+		// GetArtifact gets the build artifact address for a named version.
+		// It does not guarantee that that artifact exists.
+		GetArtifact(SourceVersion) (*BuildArtifact, error)
+		// GetSourceVersion gets the source version associated with the
+		// artifact, regardless of the existence of the artifact.
```
nyarly: I don't think this makes sense. I've been thinking a lot about the artifact name issue, and I really do not think that we can transform SV <-> BA transparently.

lib/resolve.go
```diff
@@ -1,8 +1,19 @@
...
+import (
+	"fmt"
+	"strings"
+	"sync"
+)
 
 type (
+	// Resolver is responsible for resolving intended and actual deployment
+	// states.
+	Resolver struct {
```
nyarly: I think Resolver is fine as is (re your email) - I don't see why we should be concerned with creating a new object per resolve.

lib/resolve.go
```diff
@@ -14,8 +25,95 @@ type (
...
+func NewResolver(d Deployer, r Registry, intended State) *Resolver {
+	return &Resolver{
+		Deployer:      d,
+		Registry:      r,
+		IntendedState: intended,
+	}
+}
+
+// Rectify takes a DiffChans and issues the commands to the infrastructure to reconcile the differences
+func Rectify(dcs DiffChans, d Deployer, reg Registry) chan RectificationError {
```
nyarly: Why not make this a private method of Resolver?

ext/docker/image_mapping.go
samsalisbury: In the process of moving things around I had to export a lot, in order to make incremental changes whilst being still able to compile. You're right, we need to comb through this again to tidy up our exports.

ext/singularity/rectifier.go
samsalisbury: Again, this was a necessity of the refactor, and should be unexported again.

integration/integration_test.go
samsalisbury: Ah yes, the dupes were a quick way to get things moved, but we need to remove them. The problem with consuming integration from docker is circular references iirc. I think some of the docker and singularity tests should probably be moved into the integration package...

lib/manifest.go
samsalisbury: This is definitely a problem area. Resources is strictly speaking a dictionary with no other constraints, they're defined ad-hoc by arguments to the mesos agent start script. This is why Resources need to be defined somewhere in the State, and are potentially unique to each organisation. Having said that, there are some (CPU/Mem/Ports/Disk) which are likely to turn up frequently. For simplicity towards v1.0 let's constrain resources to some well-defined `sous` types, and have Singularity do the mapping based on these. We can save custom resource definitions for v2.

lib/resolve.go
samsalisbury: There was no Resolver type 'as is' previously, just a global `Resolve` func. Making the resolver into an object means we shorten method signatures significantly for ease of reading.

My mistake was in making `IntendedState` a field of this new Resolver type. That particular field should be passed in to the Resolve method, so Resolvers can be re-used.

lib/resolve.go
samsalisbury: Why not indeed :)

ext/docker/image_mapping_test.go
samsalisbury: Not sure what you mean here? Are you agreeing with the diff or saying it should be changed?

I wonder if these tests can be made more readable by having a larger function that assembles the `NameCache` and registry and returns them in one go...

ext/docker/image_mapping.go
samsalisbury: Probably right, after the great rearrangement there are a number of things that should be unexported. I'm also considering moving some of the more integrationy tests to `integration` and refactoring them to refer only to sous core interfaces, where that is possible. (Where it's not possible, it's clear those tests must live alongside the lib with unexported functions that need to be called.)

lib/registry.go
samsalisbury: I agree in fact, but I also think getting to the stage of having an a priori mapping between SV (plus the env, see below) and Artifact name would be extremely valuable.

I wonder if, by passing more information to GetArtifact, we could get to the point of having 1-1 mappings (for CI and Prod builds only), without the need for an external source to be queried? At a minimum, we would need to pass in a hash of: (Sous Version, Buildpack Version, BaseImage version [in the case of Docker], and possibly Docker version). Perhaps we should construct a structured signature representing "the entire build system" which gets added to the manifest, and labels, and the hash of which is used in constructing artifact names themselves. It would not quite be guaranteed reproducible builds, but quite possibly a big enough stride in that direction to allow us to circumvent querying external sources. That data would also be invaluable in tracking down security vulnerabilities in packages installed in underlying base images etc.

The (Name, Type) pair is totally up for discussion as well... I think that may be enough information to identify an Artifact, and was thinking we might be better using established media type names for the Type field, and the name would simply be a unique identifier of a resource of that type.

samsalisbury: I've followed this up with an email exploring this idea in more depth, the email supersedes my comment, above.

samsalisbury: Also note that what happens behind the call to GetArtifact can be literally anything, including going via Registry/NameCache/Database/whatever.

lib/resolve.go
samsalisbury: Done in d368bdcdeabca682a0dbb7db51c56a045f4e4361

lib/resolve.go
samsalisbury: Ah, because it's referenced by `ext/singularity` ...

samsalisbury:
Further commits including these back on #67

ext/docker/image_mapping.go
samsalisbury: Done in ef211f7b1b22ca006276a99bcec4be3650d1ba6f

ext/docker/image_mapping.go
samsalisbury: Unexported it for now in 37e6123 to avoid any accidental external usage

ext/singularity/rectifier.go
samsalisbury: Done in f141abff0768ab8ff4548ea46e55ec2ebfec2821

PR #72, opened 2016-07-12
Add bin/test-sample script

samsalisbury:
- This allows us to run the tests multiple times to get a sample
  of failure rates.
- Output is placed in samples/{date}/ for review.


samsalisbury:
@nyarly well, Travis is using a version of bash that definitely supports pipefail.. https://travis-ci.org/opentable/sous/builds/144408560#L297

bin/test-sample
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
samsalisbury: This change worked (allows `set -o pipefail`) so I guess sh doesn't have this option. [Now we know](https://www.youtube.com/watch?v=MKEUujz12S4) cc @nyarly

PR #77, opened 2016-07-20
factor disk-based state storage to injectable type

samsalisbury:
- It's now injected directly into commands that need it
- The GDM loaded from disk is now directly injectable as well
- This is a breaking change to the command line, you can not
  specify a state dir right now, except via configuration.
- It's apparent we need to use CLI flags in injection.
- Adds a couple of readability tweaks.


samsalisbury:
**UPDATE:** State directory can now be specified via environment, e.g.:

```shell
    SOUS_STATE_LOCATION=$HOME/.local/share/sous/state sous [init|rectify|query_adc|query_gdm]
```

samsalisbury:
**UPDATE2:** State directory also has a default location, following XDG conventions (i.e. `XDG_DATA_HOME` and default value for that), so no need to specify it unless it's somewhere else.

PR #79, opened 2016-07-21
Fix todos

samsalisbury:
This is a bunch of unrelated commits that fix some of the issues flagged up by:

```shell
grep -rEin TODO . | grep -vE '^\./(vendor)|(\.git)|(doc)'
```

There are 2 TODOs left, which relate to bigger stories.


samsalisbury:
Note that this PR includes #78

PR #82, opened 2016-07-22
Solves several issues with existing CI process

nyarly:


.travis.yml
```diff
@@ -27,12 +27,13 @@ before_install:
...
 - docker version
 - docker-compose version
+- env | grep '^DOCKER_'
 - (cd test-registry && docker-compose pull && docker-compose build --no-cache registry)
 
 install:
 - echo "Default install step disabled by .travis.yml"
 script:
-- "./bin/test-until-success 10"
+  - "./bin/test-until-success 3"
```
samsalisbury: Just noticed this WS diff between here and next line. I think that makes it invalid YAML, and may be causing Travis not to recognise the file, although if that's the case a parse error might be nice..

nyarly:
Ready to merge

PR #83, opened 2016-07-22
Travis fix 3

samsalisbury:
Tiny whitespace change compared with #82


samsalisbury:
It's building! 💃

bin/check-compilation
```diff
@@ -1,18 +1,47 @@
...
+: > "$RESULTS_FILE"
+
+trap 'rm $RESULTS_FILE; kill $PIDS > /dev/null 2>&1 || true' EXIT
+
+check_comp() {
+	P="$1"
+	if ! TESTBUILDOUT=$(go test -c -o /dev/null "$P" 2>&1); then
+		printf '\n%s\n' "$TESTBUILDOUT"
+		echo "1" >> "$RESULTS_FILE"
+	else
```
samsalisbury: Is there a race condition here? Or are these piped writes somehow serialised? I must find out...

samsalisbury:
This isn't a real fix, as the integration tests are disabled @nyarly.

However, I've  found the place it's hanging, at least on my machine...

```
2016/07/22 12:30:25 integration_setup.go:243: Attempting connection: 192.168.99.100:7099
2016/07/22 12:30:25 integration_setup.go:243: Attempting connection: 192.168.99.100:5000
2016/07/22 12:30:25 integration_setup.go:253:   192.168.99.100:5000 up and running
2016/07/22 12:30:25 until_ready.go:94: <nil> <nil>
```

samsalisbury:
@nyarly I think due to the complexity of these tests, it may be worth rewriting parts of them with all the knowledge built up over the time, and we can probably simplify them a little. There are a few issues with using sudo, or not, as well as with working with the docker for Mac beta, which would be worth addressing. So, I have used half my day on this, I understand a little more of the issues, thanks for all your work on it. I'm going to leave it for the rest of the day, but let's talk about how to address the issues later.

samsalisbury:
Superseded by #85

PR #87, opened 2016-07-27
Build manager and build config TECHOPS-122

nyarly:
This supports most of the build-config idea, as well as guards for strict
compliance and registration

Note that in the process I've adjusted the scope: `sous build` doesn't (and
probably shouldn't) take any action with git: it doesn't do shallow clones, and
it doesn't do checkouts.


samsalisbury:
This failed due to gofmt error, merging anyway and I'll fix in follow-up

PR #88, opened 2016-07-28
Marks advisories into image metadata

nyarly:
So that the unclean might be recognized and shunned


samsalisbury:
This failed due to gofmt error, merging anyway and I'll fix in follow-up.

PR #92, opened 2016-08-01
Single cluster rectify TECHOPS-108

nyarly:
Adds -cluster flag to restrict rectify to considering only one cluster.
When used, also adds configuration to the GDM to only report that cluster,
so that we save queries targetting other clusters.

Also: renamed "-manifest" to "-repo" and added "-offset" so that manifests can
be properly constrained to.

Finally, adds "-all" because once Sous begins to be used in a "disjoint" mode,
gobal rectifies become hazardous.

### DIFF
--- a/cli/rectify_test.go
+++ b/cli/rectify_test.go
@@ -1,0 +1,14 @@
+package cli
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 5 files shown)


samsalisbury:
> samsalisbury added the in progress label 10 hours ago

did I? Strange.. I did sign up for waffle.io to take a look, so I assume this came from there.. It looks quite nice: https://waffle.io/opentable/sous

PR #94, opened 2016-08-03
Sous deploy

samsalisbury:
Add 'sous deploy' command, e.g.:

    $ sous deploy -cluster pp-sf -repo github.com/opentable/cuisines-api -version 1.0.2-beta.2

This updates the manifest for the repo to use the new version. You must run `sous rectify` afterwards to initiate the deployment.


cli/sous_deploy.go
```diff
@@ -0,0 +1,134 @@
...
+
+usage: sous deploy -cluster <name> -version <semver>
+`
+
+// Help returns the help string for this command
+func (su *SousDeploy) Help() string { return sousInitHelp }
+
+// AddFlags adds the flags for sous init.
+func (su *SousDeploy) AddFlags(fs *flag.FlagSet) {
+	fs.StringVar(&su.flags.RepoURL, "repo-url", "",
```
nyarly: Given that rectify already has a similar flag (and in fact this value passes through to rectify), shouldn't the names of the flags be the same? i.e. either both -repo or -repo-url?

cli/sous_deploy.go
```diff
@@ -0,0 +1,134 @@
...
+	var repoURL, repoOffset string
+	if err := firsterr.Parallel().Set(
+		func(e *error) { repoURL, *e = su.resolveRepoURL() },
+		func(e *error) { repoOffset, *e = su.resolveRepoOffset() },
+	); err != nil {
+		return EnsureErrorResult(err)
+	}
+
+	if su.flags.Cluster == "" {
+		return UsageErrorf("You must a cluster using the -cluster flag.")
```
nyarly: Lost word - "select"?

cli/sous_deploy.go
```diff
@@ -0,0 +1,134 @@
...
+	m := su.GDM.GetManifest(sl)
+	if m == nil {
+		return UsageErrorf("update failed: manifest %q does not exist", sl)
+	}
+
+	clusterName := su.flags.Cluster
+	cluster, ok := m.Deployments[clusterName]
+	if !ok {
+		return UsageErrorf("Cluster %q does not exist.")
+	}
```
nyarly: The checks for the manifest and the cluster are a good idea which should get carried across into `rectify` - which implies that they should be extracted to their own file for reuse.

cli/sous_deploy.go
```diff
@@ -0,0 +1,134 @@
...
+	repoURL := su.flags.RepoURL
+	if repoURL == "" {
+		repoURL = su.SourceContext.PossiblePrimaryRemoteURL
+		if repoURL == "" {
+			return "", fmt.Errorf("no repo URL found, please use -repo-url")
+		}
+		sous.Log.Info.Printf("using repo URL %q (from git remotes)", repoURL)
+	}
+	if !strings.HasPrefix(repoURL, "github.com/") {
+		return "", fmt.Errorf("repo URL must begin with github.com/")
```
nyarly: I suppose we can cross this bridge when we come to it

cli/sous_deploy.go
samsalisbury: Shared flags are now unified, see #96

cli/sous_deploy.go
samsalisbury: Addressed in #96

cli/sous_deploy.go
samsalisbury: Resolved in #96 see e980df78a8a92da2cf1e2a7e44f50312d5e7b0a5 and fd21a2de73df90ca9f4d5e86778012b8f4b8dd04

PR #95, opened 2016-08-04
Adding improved coverage scripts, and reports coverage to codecov.io

nyarly:


nyarly:
If nothing else, this gets us the codecov check, so we can try to keep coverage up.

PR #97, opened 2016-08-08
Constrains assemblers to limited pool TECHOPS-491

nyarly:
This should resolve the file descriptors exhaustion problem.
Needs to be tested in the field though.

The ultimate fix here involves limiting the number of
simultaneous build assemblers there can be at once.
The underlying Go http.Clients should only be holding
2 idle connections each (based on Go docs)
and there should be 1 client per server per GetActualDeploymentSet().

The maps that restrict the clients are local to the deployer
and registry managers, which are created at the CI level as singletons.

There's the possibility that in a long running system,
the idle sockets might back up behind garbage collection,
and we'll need to implement access to close them in a defer.


ext/singularity/actual_deployment_set.go
```diff
@@ -143,6 +148,26 @@ func catchAndSend(from string, errs chan error) {
 	}
 }
 
+func logFDs(when string) {
+	defer func() { recover() }() // this is just diagnostic
```
samsalisbury: Can this method panic? Looks like you've checked every error. At a minimum, it might be better to log whatever is recovered in case there's a weird panic emanating from `os`. In general Go is very predictable when it will and won't panic, so swallowing them seems like a code smell.

PR #99, opened 2016-08-11
Bugfix for verbosity flags

nyarly:
This started out as "hey, why doesn't -v work?" 
and grew into an overhaul of how the CLI interacts with DI.

Comments inline


cli/add_flags.go
```diff
@@ -35,14 +35,18 @@ func AddFlags(fs *flag.FlagSet, target interface{}, help string) error {
 		name := strings.ToLower(f.Name)
 		u, ok := usage[name]
 		if !ok {
-			return errors.Errorf("no usage text for flag -%s", name)
+			continue
 		}
 		switch field := fp.(type) {
 		default:
-			return errors.Errorf("target field %s.%s is %s; want string, int",
+			return errors.Errorf("target field %s.%s is %s; want string, int, or bool",
```
nyarly: Needed this so that -all would work. Besides, it seemed nice from the standpoint of working to the flag interfaces to have more types available.

cli/add_flags.go
```diff
@@ -35,14 +35,18 @@ func AddFlags(fs *flag.FlagSet, target interface{}, help string) error {
 		name := strings.ToLower(f.Name)
 		u, ok := usage[name]
 		if !ok {
-			return errors.Errorf("no usage text for flag -%s", name)
+			continue
 		}
 		switch field := fp.(type) {
```
nyarly: Briefly contemplated using ft.Kind() here...

cli/cli.go
```diff
@@ -50,12 +61,32 @@ func NewSousCLI(v semv.Version, out, errout io.Writer) (*cmdr.CLI, error) {
...
+	cli.Hooks.PreExecute = func(cmd cmdr.Command) error {
+		// Create the CLI dependency graph.
+
+		for _, c := range chain {
+			if r, ok := c.(Registrant); ok {
+				r.RegisterOn(g)
+			}
+		}
+
+		for _, c := range chain {
```
nyarly: This is in two cycles because the simple stupid timing is easier to reason about - it could be "Register, then inject" per command, but this decouples that interaction a bit more.

cli/flags_source.go
```diff
@@ -1,15 +1,75 @@
...
 
-// SourceFlags are CLI flags used to set the source context for a given command
-// invocation.
-type SourceFlags struct {
+import "github.com/opentable/sous/lib"
+
+// DeployFilterFlags are CLI flags used to configure the underlying deployments
+// a given command will refer to
+// N.b. that not every command will use every filter
+type DeployFilterFlags struct {
```
nyarly: This is the biggest digression from the original here - I'm abstracting from SourceFlags into a "deployment filter", and then controlling which flags get added to the FlagSet.

Starting to think this becomes a more central datatype, and we add methods to it to e.g. generate SourceID and SourceLocations.

cli/graph.go
```diff
@@ -35,6 +36,14 @@ type (
 	// with their dependencies.
 	SousCLIGraph struct{ *psyringe.Psyringe }
 	// Version represents a version of Sous.
+
+	// OutWriter is an alias on io.Writer to disguish "stderr"
+	OutWriter io.Writer
+	// ErrWriter is an alias on io.Writer to disguish "stderr"
+	ErrWriter io.Writer
```
nyarly: There's a bit of mismatch between this and the cmdr.Output stuff. We should keep an eye out for some kind of unification.

cli/graph.go
```diff
@@ -74,25 +79,28 @@ type (
 
 // BuildGraph builds the dependency injection graph, used to populate commands
 // invoked by the user.
-func BuildGraph(s *Sous, c *cmdr.CLI) *SousCLIGraph {
-	return &SousCLIGraph{psyringe.New(s, c,
+func BuildGraph(c *cmdr.CLI, out, err io.Writer) *SousCLIGraph {
+	return &SousCLIGraph{psyringe.New(
+		c,
+		func() OutWriter { return out },
+		func() ErrWriter { return err },
```
nyarly: Wrapper here was necessary to get psyringe to recognize the interface. There's an issue filed for memory's sake.

cli/graph.go
```diff
@@ -111,8 +119,34 @@ func newErrOut(c *cmdr.CLI) ErrOut {
 	return ErrOut{c.Err}
 }
 
-func newSourceFlags(c *cmdr.CLI) (*SourceFlags, error) {
-	sourceFlags := &SourceFlags{}
+func newLogSet(s *Sous, err ErrWriter) *sous.LogSet { // XXX temporary until we settle on logging
```
nyarly: This still manipulates a global object for the time being, with a dep factory method just to put it into the DI resolution process.

cli/graph.go
```diff
@@ -74,25 +79,28 @@ type (
...
 		newLocalWorkDirShell,
 		newScratchDirShell,
 		newLocalGitClient,
 		newLocalGitRepo,
-		newSourceFlags,
 		newGitSourceContext,
-		newSourceContextFunc,
-		newBuildContextFunc,
+		newSourceContext,
+		newBuildContext,
```
nyarly: This felt like a win, though.

util/cmdr/cli.go
```diff
@@ -37,9 +37,17 @@ type (
...
 	// it cancels execution and the error is displayed to the user.
 	Hooks struct {
+		// Parsed is run on a command when it's found on the command line
+		Parsed func(Command) error
 		// PreExecute is run on a command before it executes.
 		PreExecute func(Command) error
 	}
+
+	// A PreparedExecution collects all the information needed to execute a command
+	PreparedExecution struct {
```
nyarly: Invocation gets broken, now, into two pieces: preparing the execution, and actually running the execution. The nice result is that PE object can be used in testing regardless of what side effects actually running the command would have.

util/cmdr/cli.go
```diff
@@ -175,18 +195,15 @@ func (c *CLI) invoke(base Command, args []string, ff []func(*flag.FlagSet)) Resu
...
+				return nil, EnsureErrorResult(err)
 			}
-			return c.invoke(subcommand, args, ff)
+			return c.prepare(subcommand, args, ff)
 		}
 	}
 	// If the command can itself be executed, do that now.
 	if command, ok := base.(Executor); ok {
 		c.init()
-		if err := c.runHook(c.Hooks.PreExecute, base); err != nil {
```
nyarly: Note also that the timing of PreExecute has been pushed to after all the flags are parsed.

nyarly:
One of the overall outcomes here is that flag parsing happens before DI runs, which means the user can actually manipulate the parameters of object construction.

cli/sous.go
```diff
@@ -68,6 +67,11 @@ func (s *Sous) AddFlags(fs *flag.FlagSet) {
 		"debug: output detailed logs of internal operations")
 }
 
+// RegisterOn adds the Sous object itself to the Psyringe
+func (s *Sous) RegisterOn(psy Addable) {
+	psy.Add(s)
```
nyarly: This replaces the *Sous argument to BuldGraph.

cli/sous.go
```diff
@@ -85,27 +89,5 @@ func (s *Sous) usage() cmdr.ErrorResult {
 
 func (s *Sous) Subcommands() cmdr.Commands {
 	//s.CLI.SetVerbosity(s.Verbosity())
-	s.Verbosity()
```
nyarly: This also felt like a win: if we've got a DI system, let's use it.

cli/sous_build.go
```diff
@@ -29,24 +31,25 @@ path, it will instead build the project at that path.
 args: [path]
 `
 
+func (sb *SousBuild) AddFlags(fs *flag.FlagSet) {
+	err := AddFlags(fs, &sb.DeploymentConfig, sourceFlagsHelp)
```
nyarly: As for common-but-not-global flags, I really like this pattern. A) common execution as a single function B) using the flags help as parameterization on the interface.

cli/sous_context.go
```diff
@@ -22,9 +25,5 @@ func (*SousContext) Help() string { return sousContextHelp }
 
 // Execute prints the detected sous context.
 func (sv *SousContext) Execute(args []string) cmdr.Result {
-	sc, err := sv.SourceContextFunc()
-	if err != nil {
-		return EnsureErrorResult(err)
-	}
-	return SuccessYAML(sc)
+	return SuccessYAML(sv.SourceContext)
 }
```
nyarly: This is what an Execute() method should look like :)

cli/graph.go
samsalisbury: Nice one, another way to do this is to wrap the type you want as an anonymous struct field, depends if you want it to be able to be nil or not. You can see this method used throughout `cli/graph.go`.

cli/sous_context.go
samsalisbury: w00t!!

samsalisbury:
Looks great, just that missing vendored package and I'll merge...

PR #100, opened 2016-08-11
Pulling out setup to a script

nyarly:


nyarly:
To clarify: this *does not* pull out the integration setup into its own script. It pulls the bash stuff from .travis.yml into a bin/ script. I needed this mostly in order to debug, but this seems to have some nice value - just not editing a giant YAML document for CI is nice.

PR #104, opened 2016-08-16
Fixes sous, sous build, sous rectify

nyarly:
I did notice but don't have time to fix: sous context is broken ATM.

Gist of the fixes:

`sous` was broken by the second invoke: when we invoked the second time on the same CLI,
the Sous command object got added again, which is how it got double-registered.

The address there was to add a new "Startup" hook, which actually creates a new
graph (and chain) for each invocation. I waffled on this for a bit, but the
ultimate argument I came to was that a sequence of invocations within a single
execution should work as much as possible as if the user had run them each from
the command line. I'd be open to re-examining this, but one key observation is
that we invoke commands in order to change the global state, so caching that at
the start of a command sequence is likely to have confusing results.

`sous build` was broken because the build context wasn't set up properly. The
neat solution I found was to migrate the contruction of the BuildManager into
the DI, which further reduced the tricky-to-test Execute method.

Also note that both of the above were added to `cli/cli_test.go` so that their
setup could be tested. `sous context` _et al_ should likewise get entries in
`cli_test.go` so that we can confirm that they at least parse and inject
properly.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -61,9 +61,15 @@ func NewSousCLI(v semv.Version, out, errout io.Writer) (*cmdr.CLI, error) {
 		HelpCommand: os.Args[0] + " help",
 	}
 
-	g := BuildGraph(cli, out, errout)
-
+	var g *SousCLIGraph
 	var chain []cmdr.Command
+
+	cli.Hooks.Startup = func(*cmdr.CLI) error {
+		g = BuildGraph(cli, out, errout)
+		chain = make([]cmdr.Command, 0)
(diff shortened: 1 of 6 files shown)


samsalisbury:
> The address there was to add a new "Startup" hook, which actually creates a new
graph (and chain) for each invocation. I waffled on this for a bit, but the
ultimate argument I came to was that a sequence of invocations within a single
execution should work as much as possible as if the user had run them each from
the command line.

This is totally reasonable, good idea.

PR #103, opened 2016-08-15
Prep for mutate state from CLI + cleanup.

samsalisbury:
This PR:

- Cherry-picks work from an old branch for state mutation via CLI
- Removes the RepoURL and RepoOffset types.
- Renames Deployer.GetRunningDeployment to Deployer.RunningDeployments
- Tidies up the generated concurrent maps.
- Moves test-registry to integration/test-registry.

cc @nyarly I consider this ready to merge if you get a chance today

PR #106, opened 2016-08-16
Supersedes #103

samsalisbury:
See #103


samsalisbury:
Most of the uncovered code is generated code which I don't think should have its own tests.

nyarly:
I think we need to try to isolate type changes into their own PRs in future. They wind up being so pervasive that it makes it hard to track other changes to the code. i.e. in retrospect I wish that the transition to concurrent maps and the re-stringing of RepoURL and RepoOffset (all good things!) were here as their own PRs

bin/ggen/ggen.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	for in, out := range m {
+		goinlineArgs = append(goinlineArgs, fmt.Sprintf("%s->%s", in, out))
+	}
+
+	do("goinline", goinlineArgs...)
+	do("mv", in.File, out.File)
+	do("sed", "-i", ".bak", fmt.Sprintf("s/%s/%s/g", in.Typ, out.Typ), out.File)
+	for k, v := range m {
+		do("sed", "-i", ".bak", fmt.Sprintf("s/%s/%s/g", k, v), out.File)
```
nyarly: This part makes me extremely nervous. We'll need to keep an eye out for places where this clobbers things. Also worried about punning between Go type defs and regexp. (* and [] in particular)

Perfect world, we fork go-inline and extend it to do regexp replacements of variable names and in comment blocks.

.travis.yml
```diff
@@ -9,9 +9,10 @@ before_install:
 install:
 - echo "Default install step disabled by .travis.yml"
 script:
-- "./bin/cover"
-- "./bin/test-until-success 3"
 - "./bin/check-gofmt"
+- "go test -short ./..."
```
nyarly: It doesn't make sense to me to have Travis run the -short tests: Travis doesn't short-circuit a build on a test fail, so running test -short is just extra time spent in CI

bin/ggen/ggen.go
samsalisbury: Yup this is a pretty nasty hack, fixing up goinline would be better, but not worth the time right now. Fortunately this code is only manually at dev time and the generated code committed, so it isn't a risk to sous itself.

.travis.yml
samsalisbury: Hmm, you're right... I would like to be able to short-circuit travis test runs, I'll look into it.

PR #114, opened 2016-08-18
SourceID + SourceLocation tests

samsalisbury:
Includes #113 

Adds tests for SourceLocation, tidies up tests for SourceID.

### DIFF
--- a/lib/source_id_test.go
+++ b/lib/source_id_test.go
@@ -4,27 +4,52 @@ import (
 	"testing"
 
 	"github.com/samsalisbury/semv"
-	"github.com/stretchr/testify/assert"
 )
 
+var parseSourceIDTests = map[string]SourceID{
+	"github.com/opentable/sous,1,": {
+		RepoURL: "github.com/opentable/sous",
+		Version: semv.MustParse("1"),
+	},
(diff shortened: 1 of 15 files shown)


lib/source_id_test.go
```diff
@@ -4,27 +4,52 @@ import (
 	"testing"
 
 	"github.com/samsalisbury/semv"
-	"github.com/stretchr/testify/assert"
 )
 
+var parseSourceIDTests = map[string]SourceID{
```
nyarly: For discussion: I don't find this form more readable than the other: IOW, I'd rather define a test function and call it with literal args than add a Go for-range over a test list.

I suspect you can find guidance towards exactly the opposite, however, so this is a strong opinion lightly held.

samsalisbury: I find separating them out like this more readable, because you're listing literal inputs and literal outputs, rather than literal function calls, which in this case were really long lines.

This form can also be useful when writing groups of tests that test a mapping in both directions, as you can use the same source data. I also use it for error conditions, e.g.: https://github.com/opentable/hy/blob/master/analyse_test.go#L34

PR #115, opened 2016-08-18
Renames

samsalisbury:
Includes #114 

- Shortens SourceID and SourceLocation field names.
- "RepoURL" was a misnomer, renamed to the more ambiguous "Repo".
- "RepoOffset" was a long way of saying "Directory", renamed "Dir".

### DIFF
--- a/cli/flags_source.go
+++ b/cli/flags_source.go
@@ -23,13 +23,13 @@ func (f *DeployFilterFlags) buildPredicate() sous.DeploymentPredicate {
 
 	if f.Repo != "" {
 		preds = append(preds, func(d *sous.Deployment) bool {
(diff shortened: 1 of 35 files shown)


nyarly:
Not going to stop me merging but: Offset != Directory. For instance, it may be that we need to allow the offset to refer to a particular Dockerfile (I have one project in mind particularly), or even a build target name. I'll add an issue somewhere to revisit though.

samsalisbury:
Yeah, I'm happy to revisit this. I really want to make SourceID a pair (SourceLocation, Version) I had another quick go at this today, would rather leave that name for now until I can finish that work.

If we allow non-directories that's going to significantly affect the design of buildpacks, so we need to explore this idea further.

PR #117, opened 2016-08-18
Deployments <-> Manifests mapping

samsalisbury:
Includes #116 

Provides incomplete but correct two way mapping from manifests to deployments.

### DIFF
--- a/lib/manifest.go
+++ b/lib/manifest.go
@@ -41,7 +41,7 @@ func (m *Manifest) ID() SourceLocation {
 
 // FileLocation returns the path that the manifest should be saved to.
 func (m *Manifest) FileLocation() string {
-	return filepath.Join(string(m.Source.Repo), string(m.Source.Offset))
+	return filepath.Join(string(m.Source.Repo), string(m.Source.Dir))
 }
 
 const (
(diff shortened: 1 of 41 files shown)


lib/manifest.go
```diff
@@ -61,6 +61,36 @@ const (
...
+	if len(m.Deployments) != len(o.Deployments) {
+		return false
+	}
+	for clusterName, deploySpec := range m.Deployments {
+		if !o.Deployments[clusterName].Equal(deploySpec) {
+			return false
+		}
+	}
+	return true
+}
```
samsalisbury: This can definitely be made shorter in future.

lib/map_state_to_deployments.go
```diff
@@ -32,6 +32,40 @@ func (s *State) Deployments() (Deployments, error) {
...
+				delete(spec.DeployConfig.Env, k)
+			}
+		}
+
+		m.Deployments[d.Cluster.Name] = spec
+		ms.Set(sl, m)
+	}
+	return ms, nil
+}
+
```
samsalisbury: Above method is fugly and inefficient, I'll send a follow-up to refactor some of this tomorrow.

PR #119, opened 2016-08-20
Quick command to warm up local cache

nyarly:


nyarly:
There are (probably) trivial test failures on this. Hold until next week, or fix yourself, either way is fine.

PR #120, opened 2016-08-20
Quickie: there was a problem with how query artifacts printed

nyarly:


nyarly:
See #119

samsalisbury:
Tried this locally but I'm getting the same error as travis: https://travis-ci.org/opentable/sous/builds/153715626#L838

PR #132, opened 2016-08-25
Fixes a deadlock because a WaitGroup didn't decrement

nyarly:
On the one hand, we were limiting the number of active
SingularityRequest capturing goroutines to the number of unique
Singularity URLs.

BUT, we were counting all the clusters and waiting for them all to
finish - so if there were two "clusters" with the same URL, (i.e.
because different deployment enviroments (ci vs. pp), one would get
skipped and never be marked `Done().`

Also handled: the stray "cannot retry: <nil>" message that popped up
because the error channel closed to indicate completion.

Finally, some extra debug logging.

### DIFF
--- a/ext/singularity/actual_deployment_set.go
+++ b/ext/singularity/actual_deployment_set.go
@@ -53,10 +53,12 @@ func (sc *deployer) RunningDeployments(clusters sous.Clusters) (deps sous.Deploy
 
 	var singWait, depWait sync.WaitGroup
 
+	Log.Vomit.Printf("Setting up to wait for %d clusters", len(clusters))
 	singWait.Add(len(clusters))
 	for _, url := range clusters {
 		url := url.BaseURL
 		if _, ok := sings[url]; ok {
+			singWait.Done()
 			continue
 		}
 		//sing.Debug = true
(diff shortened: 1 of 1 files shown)


nyarly:
...in today's adventures in low hours of labor:lines of code ratios.

ext/singularity/actual_deployment_set.go
```diff
@@ -53,10 +53,12 @@ func (sc *deployer) RunningDeployments(clusters sous.Clusters) (deps sous.Deploy
 
 	var singWait, depWait sync.WaitGroup
 
+	Log.Vomit.Printf("Setting up to wait for %d clusters", len(clusters))
 	singWait.Add(len(clusters))
 	for _, url := range clusters {
 		url := url.BaseURL
 		if _, ok := sings[url]; ok {
+			singWait.Done()
```
nyarly: This is the actual fix.

ext/singularity/actual_deployment_set.go
```diff
@@ -70,19 +72,25 @@ func (sc *deployer) RunningDeployments(clusters sous.Clusters) (deps sous.Deploy
...
 
 	for {
+		var cont bool
 		select {
 		case dep := <-depCh:
 			deps.Add(dep)
 			Log.Debug.Printf("Deployment #%d: %+v", deps.Len(), dep)
 			depWait.Done()
-		case err = <-errCh:
+		case err, cont = <-errCh:
```
nyarly: This is just nice.

PR #135, opened 2016-08-26
Adds a String() method for BuildResult

nyarly:
Motivation was so that `sous build` would have nicer output

### DIFF
--- a/lib/build_result_test.go
+++ b/lib/build_result_test.go
@@ -1,0 +1,24 @@
+package sous
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/nyarly/testify/assert"
+)
+
+func TestBuildResultString(t *testing.T) {
+	assert := assert.New(t)
+
+	br := &BuildResult{
+		Elapsed:     time.Second * 5,
+		Advisories:  []string{"ephemeral tag"},
+		VersionName: "something-something-2.3.4",
+	}
+
+	str := fmt.Sprintln(br)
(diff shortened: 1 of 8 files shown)


lib/buildpack.go
```diff
@@ -64,3 +68,11 @@ type (
 func (s *EchoSelector) SelectBuildpack(c *BuildContext) (Buildpack, error) {
 	return s.Factory(c)
 }
+
+func (br *BuildResult) String() string {
+	str := fmt.Sprintf("Built: %q", br.ImageID)
+	if len(br.Advisories) > 0 {
+		str = str + "\nAdvisories:\n  " + strings.Join(br.Advisories, "  \n")
+	}
+	return fmt.Sprintf("%s\nElapsed: %v", str, br.Elapsed)
```
samsalisbury: `%s` may be better for br.Elapsed, as time.Duration implements fmt.Stringer

lib/build_result_test.go
```diff
@@ -0,0 +1,24 @@
...
+	br := &BuildResult{
+		Elapsed:     time.Second * 5,
+		Advisories:  []string{"ephemeral tag"},
+		VersionName: "something-something-2.3.4",
+	}
+
+	str := fmt.Sprintln(br)
+	assert.Regexp(`ephemeral`, str)
+	assert.Regexp(`2.3.4`, str)
+	assert.Regexp(`5`, str)
```
samsalisbury: Could maybe assert "5s" here?

PR #137, opened 2016-08-26
Adds tag validation to sous build TECHOPS-658

nyarly:
validates that tags given on the command line are formatted correctly


lib/build_manager.go
```diff
@@ -20,6 +20,7 @@ func (m *BuildManager) Build() (br *BuildResult, e error) {
 	var bc *BuildContext
 
 	e = firsterr.Returned(
```
samsalisbury: `firsterr.Set` may be more succinct here.

PR #138, opened 2016-08-26
Documentation update TECHOPS-559

nyarly:


doc/getting_started.md
```diff
@@ -0,0 +1,67 @@
...
+the IP and port you'll be on,
+and you should bind to the network accordingly.
+
+While there are many many options about how Singularity works,
+Sous provides a limited subset of these options.
+For instance, automatic port mapping is not available,
+on the grounds that a Singularity based service
+will need to announce itself somehow,
+and will need to know its outside port regardless.
+In the default networking mode, the outside and inside ports are the same.
```
samsalisbury: I'm not sure this is right, or at least may be confusing. The "outside" port on the host will certainly not be the same as the port the application process listens on, or may be occasionally through pure coincidence. Therefore, e.g. for discovery purposes, in the way we do it at OpenTable, the application needs to be passed extra information about the "outside port" in order to announce correctly, it cannot just rely on the inside listening port.

doc/getting_started.md
```diff
@@ -0,0 +1,67 @@
...
+is that your application will provide its own Dockerfile.
+
+# Summary
+
+Consequent to the above,
+applications intending to be deployed via Sous
+should address the following points:
+
+* The binary artifact is a Docker container, described by a Dockerfile you write.
+* All configuration for the application should come through environment variables.
```
samsalisbury: I think some people will take issue with this... Although I believe that written would be ideal, I think we should really specify that "all configuration that _changes per environment_ should come through environment variables".

doc/getting_started.md
```diff
@@ -0,0 +1,67 @@
...
+  * One pattern to address this requirement is to build "flavors" of configuration, and select based on ENV
+* You'll need to get your network addresses from `ENV[PORT0]` and bind accordingly.
```
samsalisbury: Hmm, think I was confusing TASK_HOST with PORT0 earlier in https://github.com/opentable/sous/pull/138/commits/c17b9d84648bcd540ccab9f53c61d67a68435cd9#r76763385 

Either way though, the real advice here should be to bind on "0.0.0.0:$PORT0" and announce on "$TASK_HOST:$PORT0". However, this all seems quite specific to OpenTable's discovery system, so not sure the documentation belongs here entirely.

doc/getting_started.md
samsalisbury: Actually, I might be wrong about this... See https://github.com/opentable/sous/pull/138/commits/c17b9d84648bcd540ccab9f53c61d67a68435cd9#r76764185

doc/getting_started.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Summary
+
+Consequent to the above,
+applications intending to be deployed via Sous
+should address the following points:
+
+* The binary artifact is a Docker container, described by a Dockerfile you write.
+* All configuration for the application should come through environment variables.
+  * One pattern to address this requirement is to build "flavors" of configuration, and select based on ENV
```
samsalisbury: "Flavour" or "ENV"? Yeah, flavours are basically a way to escape from environment variables and pick a different configuration set based on an arbitrary short string. They are something which should be avoided where possible, but I think having flavours as a fallback to support those apps which have odd requirements is definitely a useful tool.

doc/getting_started.md
```diff
@@ -0,0 +1,67 @@
...
+Sous (currently) only deploys into Singularity clusters.
+There are assumptions about
+how your application works
+that it therefore carries over
+from Singularity.
+
+More specifically,
+we assume the default networking mode in Singularity:
+you'll be provided with environment variables for
+the IP and port you'll be on,
```
samsalisbury: Suggest:

    you'll be provided with two environment variables you must use to configure your server:
        `TASK_HOST` contains the host name of the node you're running on, and
        `PORT0` contains the number of an available port you must listen on.
    If you asked for more than one port in your resources configuration, you will also have
        `PORT1, PORT2, ... PORTN-1` also indicating free ports you can listen on.

doc/getting_started.md
nyarly: This reflects my best understanding: the design at the moment, and the goal as expressed by Platform, is to use bridge networking with the default port mappings - which is outside<->inside. I'm obliquely referencing discussions I've had where it was clear that engineers wanted more complicated mappings to simplify their run scripts (this is the nginx-and-listen-port problem), but that they'd need to announce anyway, so all they're avoiding is templating a very simple nginx config.

doc/getting_started.md
nyarly: For the record, this use of "flavor" predates the idea you were talking about with Sous flavors. I had in mind the approach I've seen of having "config/{dev,prod}.properties" and then ROLE=dev => `File.new("config/"+Env.get("ROLE"))` ...

nyarly:
@samsalisbury For the life of me, I cannot remember where we left this. Can you do a counter-PR?

doc/getting_started.md
samsalisbury: Yeah that is quite a common usage of the concept. "Flavours" at OpenTable/sous are distinct from this meaning though, they are about multiple deployments of the same application in the same environment. I.e. it effectively multiplies the number of virtual environments an application can run in beyond the number of clusters we have. In my follow-up PR I'm removing reference to flavours here, since they are more of an advanced topic for special use cases.

All that's required to satisfy the "alternative configuration set per environment" from your example above is an environment variable in the deployment for that environment.

samsalisbury:
@nyarly I've opened a PR continuing from this one at https://github.com/opentable/sous/pull/143

PR #141, opened 2016-08-30
Several small bugfixes

nyarly:
The two biggest are:

* the name cache was preventing some ADS collections because
  it was too restrictive in its constraints
* the rectify predicate wasn't being applied to the ADS - so the result was
  that filtered deployments were being deleted

Fixed the usage of "ADC" which is nonsense for ADS for Actual Deployment Set.
Pointed to the defs.yaml if an undefined cluster name was used.


nyarly:
Incidentally, I'm thrilled to be ironing out little bugs and fixing docs, instead of laying down huge swaths of engine.

PR #143, opened 2016-09-01
Getting started tweaks.

samsalisbury:
- Mainly to be specific about networking options satisfying OpenTable
  in its current state, rather than for satisfying service discovery
  in general. (Since SD has many solutions, often out-of-process.)
- Also some more detail about the Dockerfile and how it is passed
  environment variables on docker run.


samsalisbury:
Note this is a follow-up PR from #138

PR #145, opened 2016-09-01
Update deployment-workflow.md

nyarly:
Little text tweaks to the example deploy


samsalisbury:
Ah, thanks, that makes it really clear.

PR #147, opened 2016-09-01
Recreates DB when schema changes, refreshing data as needed  TECHOPS-677

nyarly:
Biggest change is creation of a GroomDatabase method on NameCache.

### DIFF
--- a/ext/docker/image_mapping.go
+++ b/ext/docker/image_mapping.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"strings"
 	"text/tabwriter"
+	"time"
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/reference"
@@ -83,7 +84,9 @@ func (e NotModifiedErr) Error() string {
 
 // NewNameCache builds a new name cache
 func NewNameCache(cl docker_registry.Client, db *sql.DB) *NameCache {
-	return &NameCache{cl, db}
+	nc := &NameCache{cl, db}
+	nc.GroomDatabase()
+	return nc
 }
 
 // ListSourceIDs implements Registry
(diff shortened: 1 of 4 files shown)


ext/docker/image_mapping.go
```diff
@@ -281,77 +350,74 @@ func GetDatabase(cfg *DBConfig) (*sql.DB, error) {
...
-		"repo_name_id references docker_repo_name"+
-		"    not null"+
-		", location_id references docker_search_location"+
-		"    not null"+
-		",  primary key (repo_name_id, location_id)"+
-		");"); err != nil {
-		return nil, err
-	}
+func clobber(db *sql.DB) {
+	Log.Debug.Print("DB Clobbering time!")
```
samsalisbury: This made me laugh out loud, I do like software with personality, I just wonder if a bit more information might be helpful for the user, e.g. "Updating Docker Name Cache DB Schema" (although that would belong in the GroomDB method I guess...)

ext/docker/image_mapping_test.go
```diff
@@ -23,6 +24,113 @@ func inMemoryDB(name string) *sql.DB {
 	return db
 }
 
+func BenchmarkFPSchema(b *testing.B) {
```
samsalisbury: Cool, benchmarks :) you know what I would really like? A tool like CodeCov that would compare benchmark results from one build to another... Have you some across an online tool to do that? I have not despite a quick Google, but it would be nice. There is a CLI tool (benchcmp) that does these comparisons... Startup idea?

ext/docker/image_mapping_test.go
```diff
@@ -23,6 +24,113 @@ func inMemoryDB(name string) *sql.DB {
...
+	for i := 0; i < b.N; i++ {
+		dbName := fmt.Sprintf("inmemory%d.db", i)
+		_, err := GetDatabase(&DBConfig{"sqlite3", InMemoryConnection(dbName)})
+		if err != nil {
+			b.Log(err)
+		}
+	}
+}
+
+func TestReharvest(t *testing.T) {
```
samsalisbury: I wonder if this test should be called "TestGroom"?

PR #149, opened 2016-09-04
Flavor implementation.

samsalisbury:
Includes #148 

This PR implements the concept of "Flavor" as part of a manifest's and deployment's ID.

Flavor is a way to multiply deployments of a single codebase per cluster. The name "Flavor" is already in use at OpenTable to describe short strings used for this purpose, so we have kept the name.

The original spec we imagined for ManifestID.String was to use a colon to separate the SourceLocation from the Flavor, however this makes for file paths in the GDM that are difficult to work with (colon being the default path list separator on gnu/linux type operating systems). Therefore, the implementation uses tilde `~` to separate the two. This means that tilde is a special character in Sous, and for now will not be usable in repository names. Follow-up PR coming to make sure we validate incoming source locations, tags (versions), and cluster names to ensure none of them make use of tilde or comma, which are both special characters for Sous.

It would be possible to use alternative encoding for file paths based on ManifestID, however that would add significant complexity, which we don't need to tackle, at least not yet.

Similarly, we originally intended for the comma separator used in SourceIDs to be optional, by specifying an alternative separator at the start of the line. We can still accept this for user input for now, but regardless will validate that source locations and tags do not contain commas or tildes in their particles, to avoid the cost of allowing escape chars. If this becomes a problem later (e.g. we need to support SourceLocations containing commas and tildes), then we can address it by adding escaping at that point. I doubt this will ever become an issue for more than a couple of oddball repositories though.

### DIFF
--- a/cli/add_flags_test.go
+++ b/cli/add_flags_test.go
@@ -13,10 +13,12 @@ func TestAddFlagsForRectify(t *testing.T) {
 				all deployments should be considered
 	-cluster string
 				target deployment cluster
-  -offset string
-        source code relative repository offset
-  -repo string
-        source code repository location
+	-flavor string
+				flavor is a short string used to differentiate alternative deployments
+	-offset string
+				source code relative repository offset
+	-repo string
+				source code repository location
 `
 
 	fs := flag.NewFlagSet("rectify", flag.ContinueOnError)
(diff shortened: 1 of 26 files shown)


samsalisbury:
Relates to TECHOPS-678

PR #157, opened 2016-09-10
Baseline server, with first endpoint: a GDM lister TECHOPS-53

nyarly:
Line comments by way of tour

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -148,22 +148,24 @@ func newRegistryDumper(r sous.Registry) *sous.RegistryDumper {
 }
 
 func newLogSet(v *config.Verbosity, err ErrWriter) *sous.LogSet { // XXX temporary until we settle on logging
+	sous.Log.Info.SetOutput(err)
+
 	if v.Debug {
 		if v.Loud {
 			sous.Log.Vomit.SetOutput(err)
 		}
 		sous.Log.Debug.SetOutput(err)
-		sous.Log.Info.SetOutput(err)
-
 	}
 	if v.Loud {
-		sous.Log.Info.SetOutput(err)
 	}
 	if v.Quiet {
+		sous.Log.Info.SetOutput(ioutil.Discard)
 	}
 	if v.Silent {
+		sous.Log.Info.SetOutput(ioutil.Discard)
 	}
 
+	//sous.Log.Warn.Println("Normal output enabled")
 	sous.Log.Vomit.Println("Verbose debugging enabled")
 	sous.Log.Debug.Println("Regular debugging enabled")
 	return &sous.Log
--- a/server/handle_gdm.go
+++ b/server/handle_gdm.go
@@ -1,19 +1,36 @@
 package server
 
-import "net/http"
+import (
+	"net/http"
+
+	"github.com/opentable/sous/graph"
+	"github.com/opentable/sous/lib"
+)
 
 type (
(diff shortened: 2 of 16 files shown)


graph/graph.go
```diff
@@ -148,22 +148,24 @@ func newRegistryDumper(r sous.Registry) *sous.RegistryDumper {
 }
 
 func newLogSet(v *config.Verbosity, err ErrWriter) *sous.LogSet { // XXX temporary until we settle on logging
+	sous.Log.Info.SetOutput(err)
```
nyarly: Changed default logging somewhat: "warning" or "info" levels of logging need to be explicitly turned off now. Found this very useful for server exploratory testing, and expect it to be part of the changes we were discussing for rectify output

server/handle_gdm.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// Exchange implements the Handler interface
+func (h *GDMHandler) Exchange() (interface{}, int) {
+	data := gdmWrapper{Deployments: make([]*sous.Deployment, 0)}
+	for _, d := range h.GDM.Snapshot() {
+		data.Deployments = append(data.Deployments, d)
+	}
+	return data, 200
+}
```
nyarly: This winds up being a very simple endpoint handler to write. Future endpoints will likely be more complicated.

server/router.go
```diff
@@ -0,0 +1,126 @@
...
+		ph.LogSet.Warn.Printf("%+v", r)
+		ph.LogSet.Warn.Print(string(debug.Stack()))
+		ph.LogSet.Warn.Print("Recovered, returned 500")
+		// XXX in a dev mode, print the panic in the response body
+		// (normal ops it might leak secure data)
+	}
+}
+
+// Handling (sometimes) updates the local copy of the GDM and formats it
+func Handling(graphFac func() *graph.SousGraph, factory ExchangeFactory) httprouter.Handle {
```
nyarly: The returned function here is where most of the action is.

server/router.go
```diff
@@ -0,0 +1,126 @@
...
+		g.Add(&ResponseWriter{ResponseWriter: w}, r, p)
+
+		ph := &PanicHandler{}
+		g.Inject(ph)
+		defer ph.Handle()
+
+		h := factory()
+		g.Inject(h)
+
+		data, status := h.Exchange()
```
nyarly: The actual injected handler for the route returns data and status code.

server/router.go
```diff
@@ -0,0 +1,126 @@
...
+
+		w.Header().Add("Content-Type", "application/json")
+		w.WriteHeader(status)
+
+		e := json.NewEncoder(w)
+		e.Encode(data)
```
nyarly: When we work up an HTML version, I forsee checking for Accept headers and switching based on them, perhaps only if there's an template associated with the route. Alternatively, we could build an SPA to present the raw JSON.

server/server_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/nyarly/testify/assert"
+	"github.com/opentable/sous/config"
+)
+
+func TestOverallRouter(t *testing.T) {
```
nyarly: This test drives most of the machinery of the overall server and router, leaving very tight, fast unit tests for the individual handlers.

Issue #159, opened 2016-09-12
Unit Tests and Integration Tests


alrs:
Unit tests on travis-ci take ~20 minutes, and they're hazardous to run in a local environment because they require sudo access and start manipulating the local Docker configuration.

If integration tests and unit tests are separated, the unit tests run in 16 seconds. One way to do this is to move the test_with_docker package out of util and move it into the "integration" directory. Here's a Makefile that would work with that layout: https://gist.github.com/alrs/864caa0d29876012fa7f1e4afba071aa

Alternately, build tags could be configured to separate unit tests from integration tests. Soundcloud recommends:

> Package testing is geared around unit testing, but for integration tests, things are a little trickier. The process of spinning up external services typically depends on your integration environment, but we did find one nice idiom to integrate against them. Write an integration_test.go, and give it a build tag of integration. Define (global) flags for things like service addresses and connect strings, and use them in your tests.
http://peter.bourgon.org/go-in-production/#testing-and-validation

PR #160, opened 2016-09-12
Small change to code to humanize things.

nyarly:
Addresses part of #159. Specifically, implements the advice in
http://peter.bourgon.org/go-in-production/#testing-and-validation so that `go
test` ./... just executes unit tests


samsalisbury:
I hadn't noticed this pattern before but I really like it, nice find.

PR #162, opened 2016-09-13
GitStateManager

nyarly:
Pulls on read, pushes on write. On conflict, it rolls back and cleans up


ext/git/remote_test.go
```diff
@@ -10,6 +10,15 @@ func TestCanonicalRepoURL_GoodInput(t *testing.T) {
...
 		"git@github.com:user/project",
+		"ssh://git@github.com:user/project.git",
+		"ssh://git@github.com:user/project",
+		"ssh://github.com:user/project.git",
+		"ssh://github.com:user/project",
+		"git://git@github.com:user/project.git",
+		"git://git@github.com:user/project",
+		"git://github.com:user/project.git",
+		"git://github.com:user/project",
+		//"github.com/user/project",
```
samsalisbury: This was failing a test if I remember correctly... I figure we should handle this eventually, but happy to merge without.

ext/storage/git_state_manager.go
```diff
@@ -0,0 +1,96 @@
...
+
+// GitStateManager wraps a DiskStateManager and implements transactional writes
+// to a Git remote. It also polls the Git remote for changes
+//
+// Methods of GitStateManager are serialised, and thus safe for concurrent
+// access. No two GitStateManagers should have DiskStateManagers using the same
+// BaseDir.
+type GitStateManager struct {
+	sync.Mutex
+	*DiskStateManager //can't just be a StateReader/Writer: needs dir
```
samsalisbury: Yeah this tight coupling isn't very nice, but also probably not bad enough to prioritise fixing for now.

PR #163, opened 2016-09-13
Unit tests and helper scripts for the shell package.

alrs:
I'm happy to reorganize or do whatever it takes to make this Mac-friendly.

### DIFF
--- a/util/shell/shell_test.go
+++ b/util/shell/shell_test.go
@@ -1,0 +1,134 @@
+package shell
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+	"time"
+)
+
+func TestDefault(t *testing.T) {
+	sh, err := Default()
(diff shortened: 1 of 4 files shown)


util/shell/shell_test.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if output == testValue {
+		t.Logf("%s ==  %s\n", output, testValue)
+	} else {
+		t.Fatalf("%s != %s\n", output, testValue)
+	}
+
+	falsePath := "/bin/false"
```
nyarly: My inclination it to divide tests like these into several smaller test functions - some of that is habit and superstition from Ruby.

Reasoning, though, is that a) there's no way to screw up test independence (of local variables) that way and that b) failures are more localized - i.e. "TestCommandExpectsError" fails, instead of "TestCommand."

PR #164, opened 2016-09-15
Shell tests, continued.

alrs:
Improvements as suggested in https://github.com/opentable/sous/pull/163.


util/shell/shell_test.go
```diff
@@ -97,34 +108,52 @@ func TestCommand(t *testing.T) {
...
+	sh := &Sh{}
 	failCmd := sh.Cmd("/bin/false")
-	err = failCmd.Fail()
+	err := failCmd.Fail()
 	if err == nil {
 		t.Log("Fail() correctly returns nil on a command that exited with an error status.")
 	} else {
 		t.Fatal("Fail() should have returned nil from a command that exited with an error status.")
 	}
+}
```
samsalisbury: Nitpick: I've just noticed that we have a mixture of (error case first, then success case) in some tests, and (success case first, then error case) in other tests. In Go code, especially in the standard lib, the convention is to always deal with failure cases first, and success cases second, so I would rather see Sous code also follow this convention throughout. (The reason for this ordering is that when a failure case is unrecoverable and you `return` or otherwise terminate execution, this means you don't need to indent the success case in an else block, aiding readability.)

I also wonder if there is any value in logging success cases in tests. I'm not against this for a first pass of a test where you're trying to observe what's going on, but generally they increase the amount of code to maintain which in tests can frustrate refactoring, unless they are providing some real value.

Also, I find it useful to print the error that occurred when it was unexpected so we know why it failed. Since test failures print their file and line number, adding long descriptions of the error case might not be necessary either, although I'm not against that if it adds useful context or information that may not be obvious from the 2 or 3 lines preceding the failure.

Using the above observations, I would usually write lines 132 to 136 above like this:

```go
if err != nil {
    t.Fatal(err)
}
```

samsalisbury: CC @nyarly @alrs ^ in case that didn't show up in your messages.

PR #165, opened 2016-09-16
Autoresolver: runs every 60 seconds TECHOPS-53

nyarly:

PR #167, opened 2016-09-21
Autoresolver updated

samsalisbury:
This PR includes #165, merges in master, and makes some changes to the shell tests to make them pass on a Mac. As noted in the commit message, I don't know if that change is good or not, let's see what Travis thinks.

UPDATE: Regardless of whether the above mentioned thing works, I've reverted to using the true/false stubs @nyarly added in 6ef6518e88606f5abf83b8d67d36a486785315c6 as that's more robust.


samsalisbury:
Oh, I just saw 6ef6518e88606f5abf83b8d67d36a486785315c6 which should robustly fix the true/false issue I addressed more poorly in a2c9675315d2a5f44e27c452df7c7f8abe6279ad and which was undone by merging master into this branch. I will rectify that back to using @nyarly's stubs...

samsalisbury:
...which should probably be in `util/shell/testdata` as well, I'll move them.

PR #169, opened 2016-09-24
Yo creo complete -

nyarly:
The server should now accept PUT updates to add and modify manifests.
Next step: `sous update` actually makes those PUT requests


server/router.go
```diff
@@ -36,38 +44,58 @@ type (
...
-		Name     string
-		Methods  []string
-		Path     string
-		Exchange ExchangeFactory
+		Name, Method, Path string
+		Exchange           ExchangeFactory
 	}
 
+	// EmptyReader is an empty reader - returns EOF immediately
+	EmptyReader struct{}
```
samsalisbury: I wonder if an empty byte buffer inside an `io.Reader` variable is all that's needed for usages of EmptyReader?

Perhaps instead of a type, we could have a package-level variable:

    var EmptyReader io.Reader = &bytes.Buffer{}

Just an idea.

server/router.go
```diff
@@ -86,53 +114,139 @@ func (rm *RouteMap) PathFor(name string, params map[string]string) (string, erro
...
+
+func (mh *MetaHandler) injectedHandler(factory ExchangeFactory, w http.ResponseWriter, r *http.Request, p httprouter.Params) Exchanger {
+	h := factory()
+
+	mh.exchangeGraph(w, r, p).Inject(h)
+
+	return h
+}
+
+func (mh *MetaHandler) writeHeaders(status int, w http.ResponseWriter, r *http.Request) {
```
samsalisbury: I assume this will do more than wrap w.WriteHeader eventually?

server/router.go
```diff
@@ -86,53 +114,139 @@ func (rm *RouteMap) PathFor(name string, params map[string]string) (string, erro
...
+	mh.router.ServeHTTP(gw, gr)
+	return gw.Response()
+}
 
-		ph := &PanicHandler{}
-		g.Inject(ph)
-		defer ph.Handle()
+func (*EmptyReader) Read(p []byte) (n int, err error) {
+	return 0, io.EOF
+}
```
samsalisbury: See comment above, not sure a whole type is necessary to represent empty reader.

server/router.go
```diff
@@ -86,53 +114,139 @@ func (rm *RouteMap) PathFor(name string, params map[string]string) (string, erro
...
 
-		h := factory()
-		g.Inject(h)
+func emptyBody() io.ReadCloser {
+	return ioutil.NopCloser(&EmptyReader)
```
samsalisbury: Ah, this doesn't compile?

samsalisbury: Ah, this doesn't compile?

server/router.go
```diff
@@ -86,53 +114,139 @@ func (rm *RouteMap) PathFor(name string, params map[string]string) (string, erro
...
+}
+
+func copyRequest(req *http.Request) *http.Request {
+	new := &http.Request{}
```
samsalisbury: `new` is a builtin function, may be worth considering a name like `newRequest`?

server/handle_manifest_test.go
```diff
@@ -0,0 +1,69 @@
...
+		return &QueryValues{v}
+	}
+	ev := func(x interface{}, e error) error {
+		return e
+	}
+
+	assert.Error(ev(manifestIDFromValues(pq(""))))
+	assert.Error(ev(manifestIDFromValues(pq("repo=gh1&repo=gh2"))))
+	assert.Error(ev(manifestIDFromValues(pq("repo=gh1&offset=o1&offset=o2"))))
+	assert.Error(ev(manifestIDFromValues(pq("repo=gh1&offset=o1&flavor=f1&flavor=f2"))))
```
samsalisbury: Not sure we should be using query params to locate manifests, since they are the central piece of data in Sous, I feel they should their own paths. This will be easier for users, avoiding them having to type all the `?&=` chars, and will allow them to prefix their GitHub repo URL to navigate to their deployment configuration for that code.

samsalisbury: I am not against having a `/search` endpoint, with these kinds of query params being used to return search results that match the provided params, however.

config/config.go
```diff
@@ -21,7 +21,7 @@ type (
 		// considers the master. If this is not set, this node is considered
 		// to be a master.
 		Server string `env:"SOUS_SERVER"`
-		// BuildStateLocation is a directory where information about builds
+		// BuildStateDir is a directory where information about builds
```
samsalisbury: Thanks, that's clearer.

server/router.go
```diff
@@ -31,96 +29,157 @@ type (
...
+		new.URL = &url.URL{}
+		*new.URL = *req.URL
+	}
+	new.Body = emptyBody() //users must copy body themselves
+	return new
+}
 
-		w.Header().Add("Content-Type", "application/json")
-		w.WriteHeader(status)
+func (mh *MetaHandler) synthResponse(req *http.Request) *http.Response {
```
samsalisbury: consider "response", "makeResponse" or "newResponse" for simplicity/uniformity? Unless I'm misreading the purpose of "synthResponse" (I'm reading it as "synthesize response")

server/server_test.go
```diff
@@ -1,19 +1,96 @@
...
+	"encoding/json"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"testing"
 
+	"github.com/julienschmidt/httprouter"
 	"github.com/nyarly/testify/assert"
```
samsalisbury: I've noticed that this `assert` library escapes from the type system by using `interface{}` params everywhere, and frequently makes refactoring more difficult (we only find out that types are mismatched after running the tests rather than at compile time for example). I think there are cases where this may cause tests to _pass_ erroneously as well, I saw this at least once but can't remember exactly how it happened. Have you considered trying out the bare `testing` package from standard lib to write tests? It's slightly more verbose to type, but tests are more readable since they're in plain Go, so any Go developer can understand them easily, whereas using this assert package one has to also understand how that package works, and how to interpret ordering of arguments, the meaning of different numbers of parameters, etc, which slows down reading.

server/testdata/state/README.md
```diff
@@ -0,0 +1,110 @@
...
+(e.g. to update the version in different clusters)
+and save the manifest.
+The manifest will be sent to the server and updates will happen.
+
+## Related Projects
+
+The primary Sous repo is at opentable/sous.
+
+There's a demo project (forked from opentable/service-demo) at opentable/sous-service-demo.
+The differences may be instructive.
```
samsalisbury: I would rather not add OpenTable-specific instructions, URLs and code to this open source repository.

server/testdata/state/defs.yaml
```diff
@@ -0,0 +1,49 @@
...
+  Desc: The primary TCP port your app should listen on.
+  Scope: ""
+  Type: int
+Resources:
+- Name: memory
+  Type: Float
+- Name: cpu
+  Type: Float
+- Name: ports
+  Type: Integer
```
samsalisbury: I would rather not add OpenTable-specific instructions, URLs and code to this open source repository, because it leaks information about our internal systems like the internal URL of our discovery and singularity servers, names of our clusters, and email addresses. These things together may be useful to a bad actor, and don't provide anything useful to understanding Sous. For tests, can we at least change the strings to avoid this?

server/testdata/state/manifests/github.com/opentable/aws-stats.yaml
```diff
@@ -0,0 +1,22 @@
...
+    Env:
+      CONFIG_YAML: config-qa.yaml
+      GRAPHITE_HOST: graphite-qa.otenv.com
+      OT_ENV_TYPE: qa
+    NumInstances: 1
+    Volumes:
+    - Host: /opt/ot
+      Container: /opt/ot
+      Mode: RO
+    Version: 0.0.0-unversioned
```
samsalisbury: I would rather not add OpenTable-specific instructions, URLs and code to this open source repository, because it leaks information about our internal systems like the internal URL of our discovery and singularity servers, names of our clusters, and email addresses. These things together may be useful to a bad actor, and don't provide anything useful to understanding Sous. For tests, can we at least change the strings to avoid this?

server/testdata/state/manifests/github.com/opentable/cuisine-nodejs.yaml
```diff
@@ -0,0 +1,44 @@
...
+  pp-uswest2:
+    Resources:
+      cpus: "0.003"
+      memoryMb: "160"
+      numPorts: "1"
+    Env:
+      NODE_ENV: pp-uswest2
+    NumInstances: 2
+    Volumes: []
+    Version: 0.0.0
```
samsalisbury: I would rather not add OpenTable-specific instructions, URLs and code to this open source repository, because it leaks information about our internal systems like the internal URL of our discovery and singularity servers, names of our clusters, and email addresses. These things together may be useful to a bad actor, and don't provide anything useful to understanding Sous. For tests, can we at least change the strings to avoid this?

server/testdata/state/manifests/github.com/opentable/service-demo.yaml
```diff
@@ -0,0 +1,16 @@
...
+  ci-sf:
+    Resources:
+      cpus: "0.11"
+      memory: "352"
+      ports: "2"
+    Env:
+      OT_ENV: ci-sf
+    NumInstances: 1
+    Volumes: []
+    Version: 0.7.6-1472248433
```
samsalisbury: I would rather not add OpenTable-specific instructions, URLs and code to this open source repository, because it leaks information about our internal systems like the internal URL of our discovery and singularity servers, names of our clusters, and email addresses. These things together may be useful to a bad actor, and don't provide anything useful to understanding Sous. For tests, can we at least change the strings to avoid this?

server/server_test.go
samsalisbury: btw, not blocking over this obvs, just a general observation

server/router.go
samsalisbury: I see this was rectified in a later commit, so ignore that comment.

samsalisbury: Not sure why that comment is appearing doubled-up to me. Anyway, I see this was rectified in a later commit, so ignore that comment.

nyarly:
Regarding the use of query parameters:

First and foremost, at the moment these interfaces are an API. I don't anticipate human beings casually constructing PUT requests. If someone wants to curl some data around, they'll deal with the same format of URL that the machines do. Likewise, I'm leery of using POST instead of PUT e.g. to allow for a browser based form to work. A browser interface to Sous has been discussed several times, but I'd like to keep it as a separate concern.

Second, the idea that query parameters are "for search" is a consequence of their use (and motivation) in browser forms needing to be able to construct a URL. From the point of view of a truly RESTful API, the content of the URLs is effectively opaque. So, I contend that the use here is valid.

I justify their use thus: we do in fact have a multi-part identifier for manifests, and packaging those parts into query params dodges questions of how to parse them out. As a for instance, using `/manifest/:repo/:offset,:flavor` means effectively discarding the work that `httprouter` and `url` (because `httprouter` cannot differentiate the / in "opentable/sous" (or really "github.com/opentable/sous") from the one between "...sous/offset". The best way to solve that would be to url-encode the parts, at which point I think you lose all the ergonomics you're after.

In the near term, a machine will produce the URL and a machine will parse it, so it's form for humans isn't relevant. Part of the value of the router (and eventually, having a bidirectional router) is that the details of the URLs themselves are abstracted, and we can change them later.

But I do hold strongly that one of the worst things that's happened in the last decade or so, which I blame Rails for, mostly, is the association of "pretty" URLs with REST, when the two things are completely unrelated. And for the moment, REST is foremost in my mind.

server/handle_manifest.go
```diff
@@ -29,30 +34,77 @@ ManifestID{
...
+		Flavor: f,
+	}, nil
 }
 
 // Exchange implements Exchanger
 func (gmh *GETManifestHandler) Exchange() (interface{}, int) {
 	mid, err := manifestIDFromValues(gmh.QueryValues)
 	if err != nil {
-		return nil, http.StatusNotFound
+		return nil, http.StatusGone // Gone because we know this URL will always be wrong
```
nyarly: Gone was incorrect for a number of reasons. I'm not sure why this was in the PR at all, since it's been changed.

server/router.go
nyarly: I just double checked this: the Body field of http.Request is an io.ReadCloser, not just a Reader, which is the reasoning behind the EmptyReader. It is true that the underlying Reader can just be a Buffer, so I've made that amendment, at which point there isn't much point to keeping the Empty impl.

server/router.go
nyarly: That was the idea, yes.

server/router.go
nyarly: Good point.

server/router.go
nyarly: synthResponse produces a synthetic response: it short circuits the handlers inside the router in order to provide certain RESTful behaviors automatically. In this case, conditional PUT is handled by synthesizing the corresponding GET, and comparing the headers returned. Individual ResourceFamilies can override this behavior, of course, but it turns out to be so common as to bear handling automatically.

Likewise it may be useful to provide PATCH as "GET, apply the patch, PUT"

It's this kind of automation that can make a really RESTful API extremely productive.

The final observation here is that a future iteration should provide Context (and context-alikes) so that the synthetics can share data as appropriate. For instance, I'd like to build a StateManager that decorates the GitSM (or indeed, any SM) to cache reads until there's a write so that this sort of synthesis can save some roundtrips.

server/handle_manifest_test.go
nyarly: c.f. top-level comment. In short: I agree as far as a UI is applicable, but this isn't yet a UI.

server/server_test.go
nyarly: Conversely, I find the bare testing package far less readable, since there's a lot of on-the-spot interpretation that needs to happen. Most of the assert calls are "Equal", where the order doesn't matter.

Conversely, and again this is a larger, less solveable discussing, a lot of the table-testing I've seen fails to delight me because it embeds a lot of conditionals into the test code itself. Where it's things like "all these should be errors" make sense, or "all of these should be equal" likewise. (And I can see the case for getting "the equality you meant" as opposed to the admittedly handwavy equality of assert.Equal), but there are places where the tables get a lot more complicated, and their meaning gets quickly lost.

I start to thing that the best case would actually be something like 

```go
assert := func(pred bool, fmt... interface{}) {
  if !pred {
    t.Logf(fmt...)
    t.Fail()
  }
}
```
(I think there's some jiggery-pokery needs to happen to make the fmt work there, but...)

The thing that I get out of testify that I like most is that there's clear and consistent failure messages without a lot of consideration for each case, which helps a lot in the TDD cycle. Hints about what was tested is useful for producing those. Coming from testing in Ruby, the stdlib's testing is extremely anemic from this perspective, and I think it's often underestimated in a testing library.

One thing that I really like that I just discovered in testify (which 1.7's testing has but 1.6 does not) is that it provides suites with common setup and teardown. Leaving that out of stdlib testing was... courageous, but I think their utility is pretty well vindicated by their inclusion in 1.7's testing.

Ultimately, there are a number of things in 1.7's testing story that make me want to advance onto it as soon as is reasonable... but I don't know what the installed base is like. I guess once we have reliable release builds we don't need to care anymore.

server/testdata/state/README.md
nyarly: Yeah, all of this was ultimately a mistake to add, and it's been pulled from the repo. I'm going to look at squashing the commits that added it out as well.

server/handle_manifest.go
samsalisbury: Yeah, I reviewed this commit by commit, since all the vendoring looked noisy in the total PR. I'll bite the bullet and review entire PRs at once to avoid outdated comments in future. It may be worth squashing groups of related commits as well to make reviewing a bit easier.

server/router.go
samsalisbury: Cool, I'll have to read that code a bit more carefully to fully grok this.

server/handle_manifest_test.go
samsalisbury: Right, that's a fair point, but would eventually be good for usability/discoverability if the UI and API shared URLs where possible, with output controlled by content-type headers or whatever.

server/server_test.go
samsalisbury: > Conversely, and again this is a larger, less solveable discussing, a lot of the table-testing I've seen fails to delight me because it embeds a lot of conditionals into the test code itself. Where it's things like "all these should be errors" make sense, or "all of these should be equal" likewise.

Yeah, I agree that with table-driven tests it's preferable to have a single relationship between the things in the rows of the table, since that makes the test function easier to read. I think I've been doing this at least half the time, but there are some which could be simplified by applying that idea.

One nice pattern I've discovered with table-driven tests recently is to define a single table with columns labelled simply for their type (when possible) and multiple test functions that run against that table, each making a specific assertion or related group of assertions. To facilitate reading, each loop iteration labels one of the fields as "input", one "expected", then generates "actual", and compare with a nice error message listing all the pieces iff the test fails (and silence otherwise, you can always use `go test -v` if you want to see a list of run tests, but that's not relevant to this particular discussion).

samsalisbury:
Regarding query parameters: I wasn't making an argument from REST, but simply from conventions and following the principle of least surprise. I agree that whilst machines are the only consumers of these URLs it doesn't matter, but that case will only hold true for a short time since it strikes me as a reasonable goal of a REST API that different representations of the same underlying resource be available at the same URL, e.g. in JSON or HTML formats. This means humans using the UI or writing integrations using the API need learn only one taxonomy/URL format.

> I justify their use thus: we do in fact have a multi-part identifier for manifests, and packaging those parts into query params dodges questions of how to parse them out.

I agree on this point, always good to dodge making decisions not directly related to the task at hand :)

> As a for instance, using /manifest/:repo/:offset,:flavor means effectively discarding the work that httprouter and url (because httprouter cannot differentiate the / in "opentable/sous" (or really "github.com/opentable/sous") from the one between "...sous/offset". The best way to solve that would be to url-encode the parts, at which point I think you lose all the ergonomics you're after.

I've thought about this in terms of the web UI, and there is a simple solution that will provide us with natural URLs to display information at different "zoom levels". E.g. We handle `/github.com` as a single route, which leads to the GitHub handler, possibly providing a list of all the GitHub repos we have manifests for, enabling switching them on and off, etc. `/github.com/<user>` shows all repos owned by that user, with an indicator of whether they're currently in Sous or not, and `/github.com/<user>/<repo>` showing all manifests related to that repo. Since `/github.com` is its own handler, and because we know a GitHub repo path is exactly in the form `github.com/<user>/<repo>` any further slashes indicate the offset (offsets themselves containing slashes would require some additional thought, but I can't see it being a blocker).

samsalisbury:
> Second, the idea that query parameters are "for search" is a consequence of their use (and motivation) in browser forms needing to be able to construct a URL. From the point of view of a truly RESTful API, the content of the URLs is effectively opaque. So, I contend that the use here is valid.

Yes, strictly the content of URLs is opaque, and this scheme is technically valid, but in reality URLs which present a hierarchical taxonomy of the system are nicer for humans to use, since humans are good at understanding hierarchies. (Thus most programming languages provide "namespaces" or "packages" to logically group code.) There is not really any such thing as a system which is _only_ used by machines, at a minimum, human developers will write integrations against the API, it seems like a feature that they would need to learn the default hierarchy/taxonomy only once to use both.

PR #173, opened 2016-10-04
Deployment diff concentrated into manifest diffs

nyarly:
Using the existing deployment differ: concentrates deployment hunks into
manifest hunks. This is a precursor to doing an HTTP state manager: we'll check
for differences to the local understanding of the GDM, and push those diffs as
PUTs and DELETEs to manifests.

### DIFF
--- a/lib/diff_concentrator.go
+++ b/lib/diff_concentrator.go
@@ -1,6 +1,7 @@
 package sous
 
 import (
+	"log"
 	"sort"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


lib/diff_concentrator.go
```diff
@@ -0,0 +1,305 @@
...
+	ms, err := db.before.Manifests(defs)
+	if err != nil {
+		return nil, err
+	}
+	switch ms.Len() {
+	default:
+		return nil, errors.Errorf(
+			"bundled deployments produced multiple manifests:\n%#v\n%#v",
+			db.before, ms)
+	case 0:
```
samsalisbury: Maybe a comment `// do nothing` would aid reading for those not familiar with go's no-fallthrough switches?

lib/diff_concentrator.go
```diff
@@ -0,0 +1,305 @@
...
+	ms, err = db.after.Manifests(defs)
+	if err != nil {
+		return nil, err
+	}
+	switch ms.Len() {
+	default:
+		return nil, errors.Errorf(
+			"bundled deployments produced multiple manifests:\n%#v\n%#v",
+			db.after, ms)
+	case 0:
```
samsalisbury: Same here re comment about doing nothing for zero.

lib/diff_concentrator.go
```diff
@@ -0,0 +1,305 @@
...
+	case 1:
+		p, got := ms.Get(ms.Keys()[0])
+		if !got {
+			panic("Non-empty Manifests returned no value for a reported key")
```
samsalisbury: Often useful to include the specific key that was missing in the error message.

PR #181, opened 2016-10-13
HTTP StateManager

nyarly:
This is a complete state manager, but rather than read/write to the drives, it
issues HTTP requests to the Sous server. ReadState is completely statisfied
with GETs, and WriteState issues the actual PUTs and DELETEs to update the
remote state. Note that it returns an error if the desired WriteState is based
on a ReadState that no longer agrees with the server's. (The differences are
confined to the changes known to the StateManager, though.) Consumers of this
particular service will need to, conceivably, retain the changes they intend to
make and retry if there's an error.

The level at which that happens (i.e. the Execute of `sous update` vs. the
shell script doing the `sous update`) is an open question. For the moment I
lean toward the latter, as I expect the error case to be uncommon.

Following this PR, expect an update to the graph that selects the
GitStateManager or this HTTPStateManager based on the presence of a ServerURL
or StateLocation configuration.


nyarly:
There's a package name conflict in integration/. Not sure if I'll get to it tonight but it should be a simple fix

PR #184, opened 2016-10-14
Yo creo this completes the `sous build` client, etc.

nyarly:
As now the CLI commands that alter the Sous State now act as HTTP clients of the server transparently.

Still need e.g. `sous rectify` triggers, possibly, hey presto.


samsalisbury:
Just google translated "Yo creo". It's prob easier if we stick with simple English for work on Sous, as we already have to battle with being in different time zones etc, so one less piece of friction in our ability to comprehend one another is probably helpful.

Looks good, tx, merging.

PR #187, opened 2016-10-18
Adds a image name inserter interface and HTTP implementation

nyarly:
The Docker registry search component needs to be informed by anything that does
`sous build` of the association betwen an image name and the associated source
version (repo,offset,version) triple.  This requirement is the result if how
the Docker v2 registries work. Until now, the assumption was that the server
would be doing the builds, but it's increasingly clear that this won't always
be the case. In line with the HTTPStateManager, this PR adds an "Inserter"
interface, and a second implementation that sends assocations over HTTP to the
server. The existing, direct Sqlite Inserter or the new HTTP Inserter are
selected based on whether a Server URL is configured.

If a Server is configured, the HTTPNameInserter is used transparently by `sous
build` to notify the server of the docker images that Sous builds

### DIFF
--- a/server/router.go
+++ b/server/router.go
@@ -19,7 +19,7 @@ type (
 	// The MetaHandler collects common behavior for route handlers
 	MetaHandler struct {
 		router        *httprouter.Router
-		graphFac      GraphFactory
+		graphFac      GraphFactory //XXX This is a workaround for a bug in psyringe.Clone()
 		statusHandler *StatusHandler
 	}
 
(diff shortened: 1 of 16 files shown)


server/router.go
```diff
@@ -19,7 +19,7 @@ type (
 	// The MetaHandler collects common behavior for route handlers
 	MetaHandler struct {
 		router        *httprouter.Router
-		graphFac      GraphFactory
+		graphFac      GraphFactory //XXX This is a workaround for a bug in psyringe.Clone()
```
samsalisbury: That issue is now fixed, sorry I didn't get to it sooner, for some reason GH didn't email me your issue so I missed it... https://github.com/samsalisbury/psyringe/issues/2

PR #188, opened 2016-10-20
Fixes missing environment variables

nyarly:
There's a variety of little fixes in here; some tweaks to how advisories are
rendered and parsed that was preventing rectification on the server.

But the **real** change here, and it was a bear is in the manifest<->deployment
mappings. Because the `Deployments` is a map of pointers to Deployment, when a
`Manifests` was built from it, the step that deleted keys from the `DeploySpec`
in order to fold them out appropriately (i.e. to make sure that the Manifest
doesn't have redundant and eventually stale environment variables), those
deletes were changing the original Deployment DeploySpec. Most of the effort
involved in this PR comes down to tracking down that one fact and making the
one-line change to fix it.

I'd say that what hit us here was (effectively) global mutable state, and that
the solution would be to have the State, its continuents and related data
structures all be methodless immediate values, rather than pointers to
anything. This is certainly coming in the wake of hunting down this issue, and
I may be overstating the severity of the problem. I tend to thing that where we
turn over a rock and find a bug, there's more hiding, though.

One thing that's missing is a regression test. It's the end of my day, or I'd
write one that took a test state, copied a deployment from it, computed the
manifests from the deployments and tested that the deployment remained
unchanged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,5 @@ artifacts/
 samples/
 sous
 .agignore
-samples/
 check-compile.results
 testdata/
--- a/lib/map_state_to_deployments.go
+++ b/lib/map_state_to_deployments.go
@@ -69,7 +69,7 @@ func (ds Deployments) Manifests(defs Defs) (Manifests, error) {
 		}
 		spec := DeploySpec{
 			Version:      d.SourceID.Version,
-			DeployConfig: d.DeployConfig,
+			DeployConfig: d.DeployConfig.Clone(),
 		}
 		for k, v := range spec.DeployConfig.Env {
 			clusterVal, ok := d.Cluster.Env[k]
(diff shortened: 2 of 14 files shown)


lib/map_state_to_deployments.go
```diff
@@ -69,7 +69,7 @@ func (ds Deployments) Manifests(defs Defs) (Manifests, error) {
 		}
 		spec := DeploySpec{
 			Version:      d.SourceID.Version,
-			DeployConfig: d.DeployConfig,
+			DeployConfig: d.DeployConfig.Clone(),
```
nyarly: **THIS** right here, is the solution. If you get a chance to write a test, roll this change back to confirm that it's broken without it.

samsalisbury: Thanks for finding this! I'll get a test written for it today.

.gitignore
```diff
@@ -3,6 +3,5 @@ artifacts/
 samples/
 sous
 .agignore
-samples/
```
samsalisbury: Heh, does this really fix missing environment variables? :P

samsalisbury: Sorry, ignore that comment, I thought I was looking at a different commit.

lib/map_state_to_deployments.go
samsalisbury: See #189

PR #192, opened 2016-10-20
More attempts to make the server more robust

nyarly:
In this case, adding a recover so that panics during rectify are reduced to an
error for the respective change.


samsalisbury:
Thanks for this @nyarly, closing as superseded by #193 (which just refactors and adds an extra test)

PR #193, opened 2016-10-21
Resolver panics

samsalisbury:
Supplants #192, rebased onto master and factoring out, adding tests for panic recovery.


samsalisbury:
Looks like Travis has gone AWOL :/

PR #194, opened 2016-10-21
Do not delete

samsalisbury:
Includes #193; comments out line deleting Singularity requests.

This is not at all pretty; but I would like to get server back up-and-running ASAP, and people are relying on Sous in production now, so I want to avoid the possibility of catastrophe from this source.

For deleting things, we need to alert the owner of the deleted manifest that they have an unowned request that needs cleaning up, or else they need to add their manifest back.

### DIFF
--- a/ext/singularity/deployer.go
+++ b/ext/singularity/deployer.go
@@ -2,10 +2,12 @@ package singularity
 
 import (
 	"fmt"
+	"runtime/debug"
 	"strings"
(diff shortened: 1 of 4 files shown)


samsalisbury:
cc @wilwade

samsalisbury:
Travis seems to have hung; but there's a failing test so looking at that now.

PR #199, opened 2016-10-27
Rejects null volumes in manifests

nyarly:
Primarily, the addition here is adding the concept of Validation to States and
their components.  Flawed.Validate() returns a list of Flaws, which can then
each be Flaw.Repaired.

The disk state manager does a full attempt to repair the state on load and
store, so existing GDMs should be fixed up automatically.

The HTTP state manager checks for flaws but rejects changes that would
introduce them, forcing the user to take considered action. Maybe this is worth
relaxing, but I wanted to aviod "fixing" bad input automatically when we have a
human being at hand.

The server component that receives HTTP state manager manifest updates
Validates the manifests, and rejects with 400 if the update is invalid. It does
not, at present, render an error body describing the error, however. I wanted
to carefully consider the data leaked in such an error against the potential
value for UI. Also: less work not to provide the error body.

### DIFF
--- a/ext/storage/disk_state_manager.go
+++ b/ext/storage/disk_state_manager.go
@@ -39,6 +39,23 @@ func NewDiskStateManager(baseDir string) *DiskStateManager {
 	return &DiskStateManager{Codec: c, BaseDir: baseDir}
 }
 
+func repairState(s *sous.State) error {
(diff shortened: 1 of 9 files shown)


samsalisbury:
This is a great idea, as we'll need this pattern for other things as well, good thinking!

PR #203, opened 2016-10-28
Resource value validations

nyarly:
There were some awkward false starts because I'd misremembered that the ResDefs
were defaults, not a schema.

There's also the addition of validation "contexts" - as validation gets
contructed, various pieces can tag themselves in to be used either as part of
the repair or as part of the String() method. In the end I didn't wind up using
this, but it seemed like a good enough feature to leave, as I suspect there'll
be more validations to come. One the other hand, it might just be code to
delete later.

### DIFF
--- a/lib/deploy_config.go
+++ b/lib/deploy_config.go
@@ -50,7 +50,24 @@ func (dc *DeployConfig) Validate() []Flaw {
 			break
 		}
 	}
+	rezs := dc.Resources
+	if dc.Resources == nil {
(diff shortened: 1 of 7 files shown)


samsalisbury:
I was thinking about adding contexts/flaw wrapping yesterday, definitely useful, tx I'll review this properly later.

samsalisbury:
I've fixed the tests in #204, so closing this one.

PR #206, opened 2016-10-31
Changes to Singularity client

nyarly:
Removes the Scale method, because POSTs to the /api/requests endpoint will
update an existing Singularity Request if one exists.

Adds parameters to the PostRequest methods for the kind and owners list, so
that they can likewise be updated.

All in all, this should close the loop in terms of state we care to record in
Singularity.

### DIFF
--- a/ext/singularity/deployer.go
+++ b/ext/singularity/deployer.go
@@ -33,10 +33,7 @@ type (
 		Deploy(cluster, depID, reqID, dockerImage string, r sous.Resources, e sous.Env, vols sous.Volumes) error
 
 		// PostRequest sends a request to a Singularity cluster to initiate
-		PostRequest(cluster, reqID string, instanceCount int) error
-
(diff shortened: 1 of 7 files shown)


ext/singularity/deployer.go
```diff
@@ -135,12 +132,14 @@ func (r *deployer) RectifySingleModification(pair *sous.DeploymentPair) (err err
 	Log.Debug.Printf("Rectifying modified %q: \n  %# v \n    =>  \n  %# v", pair.ID(), pair.Prior, pair.Post)
 	defer rectifyRecover(pair, "RectifySingleModification", &err)
 	if r.changesReq(pair) {
-		Log.Debug.Printf("Scaling...")
-		if err := r.Client.Scale(
+		Log.Debug.Printf("Updating Request...")
```
samsalisbury: Just a general idea; log messages might be even more useful if they contain e.g. the DeployID, or some other context.

PR #208, opened 2016-11-04
Adds `sous manifest get,set`

nyarly:
Get prints the named manifest to stdout
Set reads a manifest from stdin and updates the State.

`edit` should be easy as a plumbing of each to a tempfile with a exec.Command($EDITOR) in between


nyarly:
Not only is coverage too low, discovered a whole class of bug in here...

PR #213, opened 2016-11-16
Bugfix: allowed advisories empty strings.

samsalisbury:
- Resolves TECHOPS-1238

I've found a series of related probable bugs in the way we (rather I :P) have been using the `copy` builtin. I've been using it backwards. The correct usage is `copy(dest, source)`. Going to get more tests in around all the Clone methods soon.

### DIFF
--- a/lib/state.go
+++ b/lib/state.go
@@ -119,7 +119,7 @@ func (cs Clusters) Clone() Clusters {
 // Clone returns a deep copy of this Cluster.
 func (c Cluster) Clone() *Cluster {
 	allowedAdvisories := make([]string, len(c.AllowedAdvisories))
-	copy(c.AllowedAdvisories, allowedAdvisories)
+	copy(allowedAdvisories, c.AllowedAdvisories)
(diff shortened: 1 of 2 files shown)


lib/state.go
```diff
@@ -119,7 +119,7 @@ func (cs Clusters) Clone() Clusters {
 // Clone returns a deep copy of this Cluster.
 func (c Cluster) Clone() *Cluster {
 	allowedAdvisories := make([]string, len(c.AllowedAdvisories))
-	copy(c.AllowedAdvisories, allowedAdvisories)
+	copy(allowedAdvisories, c.AllowedAdvisories)
```
samsalisbury: @alrs @nyarly this is the crux of this issue, I wouldn't be surprised if all the other places where this mistake is made are also causing odd things to happen. Going to look into them v soon.

samsalisbury:
@alrs @nyarly I am self-merging this as well in order to make progress, I would appreciate a post-review of these changes though, thanks.

PR #214, opened 2016-11-17
Bugfix build offset

samsalisbury:
- Build offset was not correctly propagated from source context, now it is.
- The Dockerfile buildpack had a silly error preventing it from building with offsets, now it's less silly.

### DIFF
--- a/ext/docker/dockerfile_buildpack.go
+++ b/ext/docker/dockerfile_buildpack.go
@@ -48,9 +48,9 @@ var successfulBuildRE = regexp.MustCompile(`Successfully built (\w+)`)
 // Build implements Buildpack.Build
 func (d *DockerfileBuildpack) Build(c *sous.BuildContext, dr *sous.DetectResult) (*sous.BuildResult, error) {
 	start := time.Now()
-	offset := "."
-	if c.Source.OffsetDir != "" {
-		offset = c.Source.OffsetDir
+	offset := c.Source.OffsetDir
+	if offset == "" {
+		offset = "."
 	}
 
 	cmd := []interface{}{"build"}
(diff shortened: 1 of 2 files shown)


ext/docker/dockerfile_buildpack.go
```diff
@@ -85,10 +85,11 @@ func (d *DockerfileBuildpack) Build(c *sous.BuildContext, dr *sous.DetectResult)
...
 // Detect detects if c has a Dockerfile or not.
 func (d *DockerfileBuildpack) Detect(c *sous.BuildContext) (*sous.DetectResult, error) {
-	if !c.Sh.Exists(filepath.Join(c.Source.OffsetDir, "Dockerfile")) {
-		return nil, fmt.Errorf("Dockerfile does not exist")
+	dfPath := filepath.Join(c.Source.OffsetDir, "Dockerfile")
+	if !c.Sh.Exists(dfPath) {
+		return nil, fmt.Errorf("%s does not exist", dfPath)
 	}
-	df, err := c.Sh.Stdout("cat", "Dockerfile")
+	df, err := c.Sh.Stdout("cat", dfPath)
```
alrs: Instead of shelling out here, maybe we could use os.Stat() and os.IsNotExist, a la https://github.com/alrs/fatcp/blob/master/main.go#L17-L30

samsalisbury: Yup, the only difference is that using the Sh package spits out the contents of the Dockerfile as well depending on how it's configured.

Is that what we want? Well, one of the goals of Sous is to provide radical transparency and high levels of useful output to users, so this does not seem to go against that. Also, buildpacks were always intended to be written as a collection of shell scripts, hence a lot of shelling out in this fake buildpack. The main trade off is that this (shelling out) makes Sous more fragile and non-portable (but it isn't anyway for a multitude of other reasons).

So the question is: "Is printing the contents of the Dockerfile and the command used to print it helpful to the user?" In all honesty I'm not sure, I would happily hear arguments either way...

samsalisbury: Note that the change here was only to the arguments to `c.Sh.Stdout` and not the call itself, so usually I wouldn't block a merge over this as the reported issue was not introduced by this PR. 

I appreciate you calling it out in general though, maybe this line should have a comment explaining why shelling out to cat is being used rather than the `os` lib.

samsalisbury: We could of course still print the contents of the Dockerfile without using cat. But we're then into many more lines of code to achieve the same thing.

nyarly:
Lars and I chatted about the `cat` command. I'm going to put in a story to enhance the Sh package, but otherwise we agree this is good.

PR #217, opened 2016-11-22
Remove Global DeploySpec concept

samsalisbury:
A package-by-package removal of the Global DeploySpec concept.

Whilst Global has potential to avoid repetition in configurations, and help every team think with a more global mindset, its current implementation was flawed and slowing down progress in more important areas.

We should look to implementing a feature like this in future, but for now let's remove it to make progress.


nyarly:
I wonder if we actually do want Global long term

PR #219, opened 2016-11-22
Config comments.

alrs:


alrs:
This is only comment and doc changes, so I'll self-merge.

PR #228, opened 2016-11-23
cmdr Hello World

alrs:
Hi, @samsalisbury.

What more does this program need to be a functional "hello world" example for the cmdr package?


samsalisbury:
Hey @alrs, this is a good idea. What you have there runs fine, so it's a good minimal example as is. The only thing I would ask is that you make this into a testable example (see below). This will mean putting it in a file ending `_test.go`, and can be put in the main `cmdr` dir. A good filename might be `example_minimal_test.go` 

The main thrust of cmdr was to make composing larger CLI apps with multiple subcommands and flags easier. IMO it achieves this, but only just, and actually makes some things harder than they should be. However, whilst we're using it, it's worth having examples that demonstrate usage of these features.

So, my other suggestion would be to add a nested command or two, and some flags, using `AddFlags`. Although each of those features (subcommands and flags) would probably be better in their own example files, and a final example pulling everything together would also be useful.

The go tool supports example code as a first class entity, and will run examples when you run tests. This ensures examples stay up to date, and the example code is inserted into rendered godoc, so it's really worth using. See this blog post for an intro to Go examples: https://blog.golang.org/examples 

The function naming scheme required for examples applies pretty well to tests as well, so I often use it there too. cc @nyarly I think you were asking about why I named some tests a particular way a while back...

PR #232, opened 2016-11-28
Print Command Help

alrs:
The help messages embedded in each of the command line utilities are now printing. I removed parseHelp() and the Help struct from cmdr. I removed the Usage() function that gave Help as its receiver. I removed the unused PrintHelp() function.


alrs:
Sending this to Travis revealed where the testing is happening. I'll send a new PR when it is fixed.

PR #233, opened 2016-11-28
Command Help Fixes

alrs:
The help messages embedded in each of the command line utilities are now printing. I removed parseHelp() and the Help struct from cmdr. I removed the Usage() function that gave Help as its receiver. I removed the unused PrintHelp() function. The help string for the top-level sous command was modified to match what the tests expect. Subcommands have individual short descriptions when described in the help message of the parent command. Modified the tests to expect much longer help messages.

### DIFF
--- a/cli/sous.go
+++ b/cli/sous.go
@@ -39,8 +39,6 @@ type Sous struct {
 var TopLevelCommands = cmdr.Commands{}
 
 const sousHelp = `
-invoke sous
-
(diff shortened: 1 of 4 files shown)


alrs:
I consider this to be "better," not "fixed." I'm happy to keep going. This is passing what tests we have and no longer losing the bulk of help messages, so I'd like to get these changes into master sooner rather than later.

samsalisbury:
Totally agree! Cheers!

PR #236, opened 2016-12-01
Image guard moved to per-deployment

nyarly:
Now, as we collect the build artifact (= image name) for proposed
deployments, we also e.g. check for advisories and strip out undeployable
deployments before they're sent to the deployer.

This kind of exploded in scope. The idea was to push the guard for the
availability of images from the top level (and blocking the whole
rectification if it fails) to a per-deployment scope. We need to resolve
the image name to do that, and since that can involve a docker-registry
roundtrip, I wanted to store the retrieved value.

The result is that the deployer now works on "Deployables" as opposed to
"Deployments" - a Deployable is a pair of a Deployment and a BuildArtifact.
The top level "resolver" farms out the resolution of the BuildArtifacts.

One very nice result is that the Singularity Deployer is now decoupled from
the Docker NameCache.


nyarly:
@samsalisbury - I've just improved the tests covering this change (and not coincidentally that code itself). With this merged and server redeployed, I think we're ready to take another run at getting webhooks deployed.

cli/sous_query_ads.go
```diff
@@ -13,9 +13,10 @@ type SousQueryAds struct {
 	Deployer     sous.Deployer
 	Config       graph.LocalSousConfig
 	DockerClient graph.LocalDockerClient
-	GDM          graph.CurrentGDM
-	State        *sous.State
-	flags        struct {
+	sous.Registry
```
samsalisbury: This is an anonymous field. In general, it seems better to make fields non-anonymous, unless there is a good reason for that anonymity. This is because later, if we add another dependency (in the form of a struct field) that happens to have some of the same field or method names as another anonymous field, we will have to refactor usages of these fields to avoid ambiguity. There are good cases to use anonymous fields, e.g. as mixins to implement interfaces on behalf of larger objects, but in this case I can't see a strong reason for it.

samsalisbury:
I'm just waiting for Travis to pass the tests before merging. For some reason it failed overnight, but the integration + unit tests all passed on my machine.

samsalisbury:
#238 supersedes this, fixing travis build.

PR #239, opened 2016-12-05
Adds request debug logging for all HSM HTTP traffic

nyarly:
This was done to try to address TECHOPS-1316, but once it was completed, could not reproduce an issue.
Nevertheless, the extra output is worthwhile, so I'm issuing this PR for it.

### DIFF
--- a/lib/auto_resolver_test.go
+++ b/lib/auto_resolver_test.go
@@ -51,7 +51,7 @@ func TestAfterDone(t *testing.T) {
 	ar.afterDone(tc, done, ac)
 	select {
 	case <-tc:
-	case <-time.After(time.Duration(4)):
+	case <-time.After(time.Duration(8)):
 		t.Error("Trigger channel took too long")
 	default:
 		t.Error("Trigger channel not triggered")
--- a/lib/http_state_manager.go
+++ b/lib/http_state_manager.go
@@ -22,7 +22,47 @@ type (
 	gdmWrapper struct {
 		Deployments []*Deployment
 	}
+
+	// ReadDebugger wraps a ReadCloser and logs the data as it buffers past.
+	ReadDebugger struct {
+		wrapped io.ReadCloser
+		logged  bool
+		count   int
+		read    []byte
+		log     func([]byte, int, error)
+	}
 )
+
+// NewReadDebugger creates a new ReadDebugger that wraps a ReadCloser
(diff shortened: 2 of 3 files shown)


lib/auto_resolver_test.go
```diff
@@ -51,7 +51,7 @@ func TestAfterDone(t *testing.T) {
 	ar.afterDone(tc, done, ac)
 	select {
 	case <-tc:
-	case <-time.After(time.Duration(4)):
+	case <-time.After(time.Duration(8)):
```
samsalisbury: Do you really mean 8 nanoseconds here? Seems a bit too short a time to be meaningful...

samsalisbury: You could use, e.g. `8 * time.Second` for 8 seconds. 8 being an untyped constant and `time.Second` being a second's worth of nanoseconds.

lib/http_state_manager.go
```diff
@@ -281,6 +284,71 @@ func (hsm *HTTPStateManager) modifies(mc chan *ManifestPair, ec chan error, done
 	}
 }
 
+func (hsm *HTTPStateManager) getDefs() (Defs, error) {
+	ds := Defs{}
+	url, err := hsm.serverURL.Parse("./defs")
```
samsalisbury: General observation; I tend to avoid the use of `url` as an identifier, because it's the name of a stdlib package. Obvs doesn't really matter, just usually when I'm using URLs I have to import `net/url` and then have to come up with other names, so am unaccustomed to seeing it not refer to `net/url`.

lib/http_state_manager.go
```diff
@@ -352,32 +398,12 @@ func (hsm *HTTPStateManager) del(m *Manifest) error {
...
-		return errors.Errorf("%s: %#v", grz.Status, bf)
-	}
-	rm := hsm.jsonManifest(grz.Body)
-	different, differences := rm.Diff(bf)
-	if different {
-		return errors.Errorf("%q: Remote and prior manifests don't match: %#v", mp.name, differences)
-	}
-	etag := grz.Header.Get("Etag")
 
+	// XXX I don't think the URL should be *able* to be different here.
```
samsalisbury: Yeah that makes sense, but I wonder if it's worth asserting on anyway, as if that happened it would urgently need fixing.

samsalisbury:
Oh, TECHOPS-1316, do you mean 1365?

PR #242, opened 2016-12-06
Bugfix rectify cluster confusion

samsalisbury:
Includes #241 

Resolves issue where sous thought requests belonged to clusters which they did not belong to.

### DIFF
--- a/ext/singularity/deployer.go
+++ b/ext/singularity/deployer.go
@@ -195,6 +195,37 @@ func MakeRequestID(mid sous.DeployID) string {
 	return fmt.Sprintf("%s:%s:%s", sl, mid.ManifestID.Flavor, mid.Cluster)
 }
 
+// ParseRequestID parses a DeployID from a Singularity request ID created by
+// Sous.
+func ParseRequestID(id string) (sous.DeployID, error) {
+	parts := strings.Split(id, ":")
+	if len(parts) != 3 {
+		return sous.DeployID{}, fmt.Errorf("request ID %q should contain exactly 2 colons", id)
+	}
+	if len(parts[0]) == 0 {
+		return sous.DeployID{}, fmt.Errorf("request ID %q has an empty SourceLocation", id)
+	}
+	if len(parts[2]) == 0 {
+		return sous.DeployID{}, fmt.Errorf("request ID %q has an empty Cluster name", id)
+	}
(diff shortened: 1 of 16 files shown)


samsalisbury:
NOTE: This is now deployed as ot-sous-server 0.1.41 in pp-sf and ci-sf

ext/singularity/deployer.go
```diff
@@ -195,6 +195,37 @@ func MakeRequestID(mid sous.DeployID) string {
 	return fmt.Sprintf("%s:%s:%s", sl, mid.ManifestID.Flavor, mid.Cluster)
 }
 
+// ParseRequestID parses a DeployID from a Singularity request ID created by
+// Sous.
+func ParseRequestID(id string) (sous.DeployID, error) {
```
nyarly: This seems very much like a silent reintroduction of the request id parsing proposal. I remain deeply opposed to this approach, and pushing it through in the name of expediency of a bug fix, and especially without calling it out strikes me as bad faith.

samsalisbury: Absolutely no bad faith was intended here. Overnight, someone had left Sous running in a state where it was redeploying most things ad nauseam. It was very important to stabilise our deployments and to undo the regression that had been introduced. I think you will agree that having a broken Sous is probably the worst thing possible in terms of driving up adoption, which is quite an important goal for our team.

This change, which we discussed in the standup, was the simplest way to achieve this, and is entirely transparent. That’s what pull requests are for. Thank you for your review. Should this implementation cause actual problems, it would be very easy to replace with an alternative, less visible, less obvious method. I take issue with your accusation, following up in a private email.

samsalisbury:
I'm working on fixing the broken test and covering this change with one presently.

The ResolveFilter requires that we know the name of the Sous-defined cluster of the deployment. However, the way we are collecting deployments from Singularity does not tell us which Sous-defined cluster they belong to, only which Singularity cluster they belong to, which is irrelevant. Therefore, in the context of reading requests from Singularity, there exists no prior art to achieve what this change achieves.

samsalisbury:
That last commit ac0f53090c18dd095ca71ae174e44865f74f84d4 fixes the integration tests broken by this fix. I will follow with tests ensuring the change introduced does not regress tomorrow. I am happy for this to be merged as-is, assuming Travis is happy with tests and coverage...

PR #246, opened 2016-12-13
Cache fixes

samsalisbury:
This PR first adds a duplicate (failing) test to demonstrate that we get different results depending on if we get docker image metadata from the cache or not (e5db6c2). Then it demonstrates that one of the assertions in the tests is not satisfiable by commenting it out, making the tests pass (86169bc). Next we change the way semantic versions are compared in cache database queries to align that with the semver spec (fc3cde9). Then we add revision data to the cache via alteration of the text format used (6b279bd). Finally, the broken assertions are reinstated, and the tests pass (3f183be).

### DIFF
--- a/ext/docker/image_mapping.go
+++ b/ext/docker/image_mapping.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"regexp"
 	"strings"
+	"sync"
 	"text/tabwriter"
 	"time"
 
@@ -364,9 +365,11 @@ var schema = []string{
 
 var schemaFingerprint = fingerPrintSchema(schema)
 
+var registerSQLOnce = &sync.Once{}
+
 // GetDatabase initialises a new database for a NameCache.
 func GetDatabase(cfg *DBConfig) (*sql.DB, error) {
-	driver := "sqlite3"
+	driver := "sqlite3_sous"
 	conn := InMemory
 	if cfg != nil {
 		if cfg.Driver != "" {
@@ -377,13 +380,19 @@ func GetDatabase(cfg *DBConfig) (*sql.DB, error) {
 		}
 	}
 
-	sql.Register(driver, &sqlite.SQLiteDriver{
-		ConnectHook: func(conn *sqlite.SQLiteConn) error {
-			if err := conn.RegisterFunc("semverEqual", semverEqual, true); err != nil {
-				return err
-			}
-			return nil
-		},
+	if driver == "sqlite3" {
+		driver = "sqlite3_sous"
+	}
+
+	registerSQLOnce.Do(func() {
+		sql.Register(driver, &sqlite.SQLiteDriver{
+			ConnectHook: func(conn *sqlite.SQLiteConn) error {
+				if err := conn.RegisterFunc("semverEqual", semverEqual, true); err != nil {
+					return err
+				}
+				return nil
+			},
+		})
 	})
 
 	db, err := sql.Open(driver, conn) //only call once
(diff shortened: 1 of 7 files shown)


samsalisbury:
@nyarly one question remaining from this PR is: do we need the SQLite Driver name to be in the configuration at all? This PR basically ignores the configuration and always uses the same name, which is hard-coded all over the place anyway. I wonder if a follow-up PR should remove this option from the config?

ext/docker/image_mapping.go
```diff
@@ -377,10 +380,40 @@ func GetDatabase(cfg *DBConfig) (*sql.DB, error) {
...
 	}
 
+	if driver == "sqlite3" {
+		driver = "sqlite3_sous"
+	}
+
+	registerSQLOnce.Do(func() {
+		sql.Register(driver, &sqlite.SQLiteDriver{
+			ConnectHook: func(conn *sqlite.SQLiteConn) error {
+				if err := conn.RegisterFunc("semverEqual", semverEqual, true); err != nil {
```
nyarly: I had no idea you could do this, which seems pretty cool. Kind of a big hammer to swing here.

integration/integration_test.go
```diff
@@ -45,7 +45,40 @@ func newInMemoryDB(name string) *sql.DB {
 	return db
 }
 
-func TestGetRunningDeploymentSet(t *testing.T) {
+func TestGetRunningDeploymentSet_all(t *testing.T) {
```
nyarly: As I understand this, we're duplicating a test here, which relies on an actual failure of the test infrastructure - our "reset" doesn't properly clear the cache. If it did, these tests would incorrectly pass in the face of this bug.

It'd be better maybe to have extracted the body of the test into its own function, and simply run it twice under the original test name.

integration/integration_test.go
```diff
@@ -45,7 +45,40 @@ func newInMemoryDB(name string) *sql.DB {
...
+	ds, which := deploymentWithRepo(nc, assert, d, "github.com/opentable/docker-grafana")
+	deps := ds.Snapshot()
+	if assert.Equal(3, len(deps)) {
+		grafana := deps[which]
+		assert.Equal(SingularityURL, grafana.Cluster.BaseURL)
+		assert.Regexp("^0\\.1", grafana.Resources["cpus"])    // XXX strings and floats...
+		assert.Regexp("^100\\.", grafana.Resources["memory"]) // XXX strings and floats...
+		assert.Equal("1", grafana.Resources["ports"])         // XXX strings and floats...
+		assert.Equal(17, grafana.SourceID.Version.Patch)
+		assert.Equal("91495f1b1630084e301241100ecf2e775f6b672c", grafana.SourceID.Version.Meta)
```
nyarly: Reviewing this, I'm led to question our design decision to embed the commit ID into the Meta field of the version. I'd want to review, but I wonder if there's anywhere that we need or want to present the commitID that it couldn't appear alongside the version, instead of pushing it into/pulling it out of the version string.

samsalisbury: Possibly. One thing I think we discussed briefly a while back was to make the commitID present everywhere, including the manifest Version field, for reasons noted below. I would be happy to go one way or the other: either SourceIDs always have the Meta field populated with the commitID, or we simply record commitID completely separately. However, always using CommitID, with the semver component more of a human-friendly decoration in the version string seems like a powerful idea, enabling teams with no versioning policy yet to start deploying stuff.

Adding the Meta field = commitID everywhere, including in manifests, may be the simplest thing, as this would obviate the need for teams to start using semver initially, and allow them to just start deploying `0.0.0+{commitID}` straight away. That way we could use the "nearestTag" plus commitID as the version. It would be possible to also check whether this commitID exactly matches the semver Git tag, and if not add an advisory. I suspect doing it this way will be less code to write, as well as just supporting new projects out of the box which may be good for adoption. From a data-model perspective it also exactly encodes the concept that what gets deployed is code from a repo at a particular commit, which is core to Sous.

PR #247, opened 2016-12-13
Separates QA setup into a standalone tool.

nyarly:
Apologies in advance: this branch was pushed before being rebased, then the
rebase got clobbered fixing "no-fast-forward" issues.

The interesting commits are: 55f7502 3ec4900 ff70f6b c6689af

The changes here are entirely around extracting the QA setup into a command
line tool, and adjusting the config to accommodate that change. There shouldn't
be any changes to actual test or code under test in this PR.


samsalisbury:
Hi, I ran `./bin/test -tags integration` on this branch but got `2016/12/14 06:48:10 docker_helpers.go:47: Panic: SOUS_QA_DESC is unset! Integration tests now require a description file generated by sous_qa_setup.`; so far this makes sense.

Then I found `util/sous_qa_setup` and wanted to run that to set up the env but I couldn't build it...

```
sam-mac:sous_qa_setup ssalisbury$ go build 
# github.com/opentable/sous/util/sous_qa_setup
./certificates.go:20: undefined: EnvDesc
```

I also tried running `main.go` in case that was the intention...

```
sam-mac:sous_qa_setup ssalisbury$ go run main.go 
# command-line-arguments
./main.go:138: undefined: registryCerts
```

I think this is the same thing Travis is seeing.

samsalisbury:
I just looked at this again and saw that it needs some code generation in certificates.go so I tried `go generate`, and found that I need to install `github.com/nyarly/inlinefiles`. I installed that, and then ran `go generate` in the sous_qa_setup directory. This appeared to succeed, although I did not see any generated files, which likely means the generated file is already checked in, as it should be, so no surprises there. I tried to run `go build` again, and got the same error as before. Giving up on this for now, would like a little extra documentation on getting it working if possible.

samsalisbury:
One other thing... This should have been a single review, but to late now... The `util/` directory is intended for completely generic utility stuff that has nothing to do with Sous specifically, but that we have not yet split out into their own repos. I would suggest that the `sous_qa_setup` package belongs in `integration` rather than `util`, unless I'm missing something?

nyarly:
I'm going to argue that the coverage failure here should be ignored. It's support code for the integration setup, not actual application code. I wonder if there's a way to exclude that from coverage consideration easily without actually removing the code to a different repo?

Related note, in looking at moving sous_qa_setup, started to see how the things that are specific to it could easily be extracted into config (i.e. which ports to find the services our CI depend on, so we can test liveness - **almost** determinable from the docker-compose.yaml) so I was tempted to make that change and really extract test_with_docker... Saves a couple hours not to though...

samsalisbury:
Hi, nice one for getting this to run on Travis. I think we may still have a docs/usability problem, but happy to merge this and follow-up with another PR, after standup.

I couldn't find any docs explaining how to run the integration tests in this PR, apologies if I missed them. Through trial and error, I have worked out that to run integration tests, I now need to do this:

```shell
sam-mac:sous_qa_setup ssalisbury$ go build
sam-mac:sous_qa_setup ssalisbury$ ./sous_qa_setup --compose-dir ../../integration/test-registry --out-path ./env-desc.json
sam-mac:sous_qa_setup ssalisbury$ export SOUS_QA_DESC="$PWD/env-desc.json"
sam-mac:sous_qa_setup ssalisbury$ cd ../..
sam-mac:sous ssalisbury$ ./bin/test -tags=integration
```

Well, that certainly worked for me anyway, is there a quicker way? I would suggest that a script, e.g. `./bin/integration` that performs these additional setup steps, and then runs `./bin/test --tags=integration` would be a really helpful addition to this, to demystify it and speed up its use a bit.

Before doing this, I thought I would try running `./bin/ci-setup` as I saw it was changed in this PR, but that was wanting me to sudo my life away so I quickly retreated.

PR #248, opened 2016-12-13
CLI Ideas

alrs:
I'm exercising CLI commands. Here are some proposed improvements to the help text.

### DIFF
--- a/cli/sous_context.go
+++ b/cli/sous_context.go
@@ -18,7 +18,8 @@ func init() { TopLevelCommands["context"] = &SousContext{} }
 
 const sousContextHelp = `show the current build context
 
-context prints out sous's view of your current context
+sous context describes Sous's understanding of the state of your
+Git respository.
 
 args:
 `
--- a/cli/sous_query_gdm.go
+++ b/cli/sous_query_gdm.go
@@ -19,9 +19,10 @@ type SousQueryGDM struct {
 
 func init() { QuerySubcommands["gdm"] = &SousQueryGDM{} }
 
-const sousQueryGDMHelp = `Loads the current deployment configuration and prints it out
+const sousQueryGDMHelp = `The attempted state of deployment for every project and every cluster known to Sous.
 
-This should resemble the manifest that was used to establish the intended state of deployment.
+The results of 'sous query gdm' and 'sous query ads' will not be identical if
+a problem is preventing sous from modifying the current state of Singularity.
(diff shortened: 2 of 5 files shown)


lib/build_config.go
```diff
@@ -127,7 +127,7 @@ func (c *BuildConfig) Resolve() {
 // Validate checks that the Config is well formed
 func (c *BuildConfig) Validate() error {
 	if _, ve := semv.Parse(c.Tag); ve != nil {
-		return fmt.Errorf("build config: tag format %q invalid: %s", c.Tag, ve)
+		return fmt.Errorf("semver git tag required. build config: tag format %q invalid: %s", c.Tag, ve)
```
samsalisbury: Thanks for this, I think this demostrates the general rule we should follow: all error messages should explain how to make things better, for faster fixing of problems by end-users, and fewer support requests :)

I would suggest we can make this message even shorter and more pithy, something like:

```
git tag %q is not valid semver: %s
```

But this is a definite improvement on what we had, so still merging, above is just a suggestion for giving our users less text to read whilst still conveying the essential info.

cli/sous_context.go
```diff
@@ -18,7 +18,8 @@ func init() { TopLevelCommands["context"] = &SousContext{} }
 
 const sousContextHelp = `show the current build context
 
-context prints out sous's view of your current context
+sous context describes Sous's understanding of the state of your
+Git respository.
```
samsalisbury: I like it, no point clouding this fact. In future we may add support for other VCSs, but for now it makes sense to be as specific as we can.

cli/sous_query_gdm.go
```diff
@@ -19,9 +19,10 @@ type SousQueryGDM struct {
 
 func init() { QuerySubcommands["gdm"] = &SousQueryGDM{} }
 
-const sousQueryGDMHelp = `Loads the current deployment configuration and prints it out
+const sousQueryGDMHelp = `The attempted state of deployment for every project and every cluster known to Sous.
```
samsalisbury: I would use "intended" rather than "attempted" here, as that's more consistent with the docs. An intended state will not necessarily have been attempted, depending on the existence of a rectifying process somewhere...

samsalisbury:
Ignoring the coverage fail here, as it doesn't make much sense to require additional coverage for a few textual tweaks.

PR #251, opened 2016-12-14
Test stability (follow up from cluster-confusion test).

samsalisbury:
Includes #249 because it would result in merge conflicts otherwise.

**UPDATE: Note that only the last two commits in the PR are new (e5b80bc and 2b32d5f)**

Addresses issue noted in https://github.com/opentable/sous/pull/246#discussion_r92236804 where we had a duplicated test that additionally relied on lack of isolation between tests for the intended failure condition to cause a failure. Now the failure condition will hold even when separate `Test` functions are completely isolated.


samsalisbury:
@nyarly @alrs Bump? :D

PR #253, opened 2016-12-15
Fix for skipped multi-cluster test

nyarly:
This builds on #251, removing the Skip on the multiple logical cluster test


samsalisbury:
Great, thanks for fixing :)

PR #254, opened 2016-12-15
New integration usability

samsalisbury:
Expands on #247, adds a few conveniences and minimal docs.


samsalisbury:
cc @nyarly in case you didn't get a notification ... see also https://github.com/opentable/sous/pull/247#discussion_r92667453

PR #256, opened 2016-12-15
Changes to Git configuration lookup

nyarly:
Sketchy quickfix which should now be (properly) handled by our sous-server
wrapper conflicting with my git config in testing

### DIFF
--- a/ext/storage/git_state_manager.go
+++ b/ext/storage/git_state_manager.go
@@ -36,7 +36,14 @@ func (gsm *GitStateManager) git(cmd ...string) error {
 	}
 	git := exec.Command(`git`, cmd...)
 	git.Dir = gsm.DiskStateManager.BaseDir
-	//git.Env = []string{"GIT_CONFIG_NOSYSTEM=true", "HOME=none", "XDG_CONFIG_HOME=none"}
+
+	// This had been commented out, with a comment to the effect that it was fixing a bug.
+	// However, it could lead to come confusion because the behavior of Sous
+	// server would change based on the configuration of git outside of the GDM
+	// repo.
+	// It's also definitely causing problems in testing for me (JL) because my
+	// local git configuration effects testing behaviors.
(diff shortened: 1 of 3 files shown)


ext/storage/git_state_manager.go
```diff
@@ -36,7 +36,14 @@ func (gsm *GitStateManager) git(cmd ...string) error {
...
 	git.Dir = gsm.DiskStateManager.BaseDir
-	//git.Env = []string{"GIT_CONFIG_NOSYSTEM=true", "HOME=none", "XDG_CONFIG_HOME=none"}
+
+	// This had been commented out, with a comment to the effect that it was fixing a bug.
+	// However, it could lead to come confusion because the behavior of Sous
+	// server would change based on the configuration of git outside of the GDM
+	// repo.
+	// It's also definitely causing problems in testing for me (JL) because my
+	// local git configuration effects testing behaviors.
+	git.Env = []string{"GIT_CONFIG_NOSYSTEM=true", "HOME=none", "XDG_CONFIG_HOME=none"}
```
samsalisbury: Ah, didn't see this before merging #257, great! Out of curiosity, is "none" a special string in XDG_CONFIG? Or if not, just wondering why that's necessary on top of `GIT_CONFIG_NOSYSTEM`?

We should also add `GIT_CONFIG_NOGLOBAL=true`as that's where a lot of users (including me) keep the majority of their configuration (in the global ~/.gitconfig rather than the system ~/$XDG_CONFIG_HOME/git/config) see https://git-scm.com/docs/git-config#FILES Maybe with this change, we wouldn't need to set `HOME` and `XDG_CONFIG_HOME` ?

nyarly: Did not get a notification, for either.

XDG_... is a prefix used in more recent Linux distros, and respected by git. The HOME=none should be enough to prevent global config, since ~/.gitconfig will expand to none/.gitconfig. Still, adding NOGLOBAL can't hurt.

samsalisbury: Added `GIT_CONFIG_NOGLOBAL=true` in #266

PR #257, opened 2016-12-15
Fixes a DI oversight in `sous query artifacts`

nyarly:
When the DryRunOption was added, `sous query artifacts` got missed. I found a
test that had been commented out from before that time, which is probably why
we had this issue.

DryRunOption became an input to the (docker) Registry injection - in certain
circumstances, a dummy registry gets injected. Because `s q a` doesn't update
the registry, I'm simply injecting a fixed "DryRunNeither" so that a live
registry client becomes available.

Along with this change, I ran across a couple of other places where local git
configuration impacted testing. Tried a different approach to addressing that.
Not sure which I like better.

### DIFF
--- a/cli/sous_server_test.go
+++ b/cli/sous_server_test.go
@@ -43,25 +43,22 @@ func TestEnsureGDMExists(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Create the empty source repo.
-	c := exec.Command("git", "init")
-	c.Dir = gdmSourceRepo
-	if err := c.Run(); err != nil {
-		t.Fatal(err)
+	git := func(args ...string) {
+		c := exec.Command("git", args...)
+		c.Dir = gdmSourceRepo
+		if err := c.Run(); err != nil {
+			t.Fatal(err)
+		}
 	}
+
+	// Create the empty source repo.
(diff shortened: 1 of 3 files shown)


cli/sous_server_test.go
```diff
@@ -43,25 +43,22 @@ func TestEnsureGDMExists(t *testing.T) {
...
+		c := exec.Command("git", args...)
+		c.Dir = gdmSourceRepo
+		if err := c.Run(); err != nil {
+			t.Fatal(err)
+		}
 	}
+
+	// Create the empty source repo.
+	git("init")
+	git("config", "commit.gpgSign", "false")
```
samsalisbury: Thanks, this is definitely a bit less noisy now. It seems to me that any tests which involve shelling out, or that test code which itself shells out, should ideally be done inside a container so we have zero chance of doing anything nasty to the user's own configuration. This would probably impact huge swathes of our tests, but would be a big plus for cleanliness. Further down the road, it would let us test against a matrix of external tools, e.g. git versions, docker versions, etc. Not something to tackle right now, but I wonder if the work you're doing in integration tests would help us here as well eventually @nyarly ?

PR #262, opened 2016-12-16
Confirms that the DOCKER_MACHINE_NAME is known to docker-machine

nyarly:
I really like the change to sniffing the environment (I think when I was first
writing this, I wasn't as clear what was a necessary input of docker-machine
and what was mere convenience...) but I also want to fail early if the named
machine isn't built.


samsalisbury:
Superseded by #263

PR #267, opened 2016-12-20
Add ResolveStatus

samsalisbury:
This is a big pull request which removes `Resolver.Resolve`, and replaces it with `Resolver.Begin`, which immediately returns a `ResolveStatus`. You can call `Wait` on the returned `ResolveStatus` in order to get the same error the old `Resolve` method would have given.

This pull request includes some incidental cleanup, for example making comments into sentences (without this, I find reading comments a little jarring, not sure if there was meant to be more or not); I tend to find the advice at https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences reasonable and helpful. These cleanup steps are separated into their own commits, so should be easy to review independently.

This pull request also includes a preliminary change to `server` which adds a blank `/status` endpoint, which always returns a 404. I will submit a follow-up pull request to flesh out this endpoint, using data from ResolveStatus. I would rather have this work so far merged asap as it touches a number of files.


samsalisbury:
This is failing a test, I've found the issue, update coming soon...

nyarly:
I don't have a strong objection here, but what's missing for me is the motivation. In other words, how will Resolver.Begin().Done() be used, that Resolver.Resolve() didn't satisfy?

server/statusmiddleware.go
```diff
@@ -9,15 +9,15 @@ import (
 
 type (
 
-	// A StatusHandler processes panics into 500s and other status codes
-	StatusHandler struct {
+	// A StatusMiddleware processes panics into 500s and other status codes.
```
nyarly: Maybe we want to go one step further with this to "HTTPStatusMiddleware" or something? The distinction is between the HTTP request/response status and Sous' resolve status, after all.

samsalisbury:
In reply to https://github.com/opentable/sous/pull/267#issuecomment-268403311 the motivation here is to be able to return a status immediately, for usage by the HTTP /status endpoint, so we can report on intermediate statuses. That status can be long-lived (between requests since it is safe for concurrency) and therefore provide the core data for this endpoint.

PR #273, opened 2016-12-27
Adding /status endpoint

nyarly:
This adds 'GET /status' that can be queried to get: the current GDM, the last stable resolve status (i.e. at the end of a resolve cycle) and the current (in progress) resolve status.

This should be enough data that clients can poll it for updates. If load is unacceptable, we can add a variation that omits the live status and rely on etagging for caching.

### DIFF
--- a/lib/auto_resolver.go
+++ b/lib/auto_resolver.go
@@ -24,7 +24,7 @@ type (
 		*LogSet
 		listeners []autoResolveListener
 		sync.RWMutex
-		status *ResolveStatus
+		status *ResolveRecorder
 	}
 )
 
@@ -90,7 +90,7 @@ func (ar *AutoResolver) Kickoff() TriggerChannel {
 }
 
 // Status returns the current status of the resolution underway.
-func (ar *AutoResolver) Status() ResolveStatus {
+func (ar *AutoResolver) Status() ResolveRecorder {
 	ar.RLock()
 	defer ar.RUnlock()
 	return *ar.status
--- a/lib/auto_resolver.go
+++ b/lib/auto_resolver.go
@@ -25,7 +25,7 @@ type (
 		listeners []autoResolveListener
 		sync.RWMutex
 		stableStatus, liveStatus *ResolveStatus
-		status                   *ResolveRecorder
+		currentRecorder          *ResolveRecorder
 	}
 )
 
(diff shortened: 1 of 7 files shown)


lib/auto_resolver.go
```diff
@@ -89,11 +90,22 @@ func (ar *AutoResolver) Kickoff() TriggerChannel {
...
+		return
+	}
+	ar.Lock()
+	defer ar.Unlock()
+	ls := ar.status.CurrentStatus()
+	ar.liveStatus = &ls
+}
+
+// Statuses returns the current status of the resolution underway.
+func (ar *AutoResolver) Statuses() (*ResolveStatus, *ResolveStatus) {
```
alrs: Might this be a spot to use a time.Ticker and return an error if it times out?

lib/resolvestatus.go
```diff
@@ -1,146 +1,170 @@
...
-	sync.RWMutex
-}
+import (
+	"sync"
+)
+
+type (
+	// ResolveStatus captures the status of a Resolve
+	ResolveStatus struct {
+		Phase string
```
alrs: Phase lost its comment

lib/resolvestatus.go
```diff
@@ -1,146 +1,170 @@
...
+		Log   []DiffResolution
```
alrs: Log lost its comment.

lib/resolvestatus.go
```diff
@@ -1,146 +1,170 @@
...
+		Errs  ResolveErrors
```
alrs: Errs lost its comment.

lib/resolvestatus.go
```diff
@@ -1,146 +1,170 @@
...
+		err error
+		sync.RWMutex
+	}
 
-// NewResolveStatus creates a new ResolveStatus and calls f with it as its
-// argument. It then returns that ResolveStatus immediately.
-func NewResolveStatus(f func(*ResolveStatus)) *ResolveStatus {
-	rs := &ResolveStatus{
+	// DiffResolution is the result of applying a single diff.
+	DiffResolution struct {
```
alrs: DiffResolution has uncommented fields. Their roles are pretty obvious, but we could probably pick up some go vet points here.

lib/auto_resolver.go
nyarly: The only way this should time out is if there's a deadlock, which is a much worse problem. I'd rather make it clearer that it needs to be addressed.

PR #275, opened 2016-12-29
Future-proofing the /servers API

nyarly:
In the future, we'll likely want to give cluster names to the returned
servers. We can change the configs pretty easily, but the deployed
client base will be harder to change. Wrapping the URL string in a
struct (-> a JSON object) lets us add e.g. a Name field when that's
needed.

### DIFF
--- a/server/handle_servers.go
+++ b/server/handle_servers.go
@@ -3,14 +3,20 @@ package server
 import "github.com/opentable/sous/config"
 
 type (
+	// ServerListResource dispatches /servers
 	ServerListResource struct{}
 
+	// ServerListHandler handles GET for /servers
 	ServerListHandler struct {
 		Config *config.Config
 	}
 
+	server struct {
+		URL string
+	}
+
 	serverListData struct {
-		Servers []string
+		Servers []server
 	}
 )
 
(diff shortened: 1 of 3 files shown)


server/handle_servers.go
```diff
@@ -3,14 +3,20 @@ package server
...
 type (
+	// ServerListResource dispatches /servers
 	ServerListResource struct{}
 
+	// ServerListHandler handles GET for /servers
 	ServerListHandler struct {
 		Config *config.Config
 	}
 
+	server struct {
```
alrs: This is likely more for my education than a real review question: Why make a single-element struct instead of just declaring a server to be a string? Is it a coin-toss of style to consider storing this value as a url.URL type instead of as a string?

server/handle_servers.go
```diff
@@ -3,14 +3,20 @@ package server
...
+		URL string
+	}
+
 	serverListData struct {
```
alrs: Same as the last question, roughly. Why not declare the serverListData type to be an alias to []servers?

alrs: Judson explained JSON de-serialization is the reason for these decisions, which made sense to me.

PR #278, opened 2016-12-31
sous status + StatusPoller

nyarly:


lib/httpclient.go
```diff
@@ -0,0 +1,254 @@
...
+// the grounds that the client is going to clobber a value it doesn't know
+// about.) Then it issues a PUT with "If-Match: <etag of from>" so that the
+// server can check that we're changing from a known value.
+func (client *HTTPClient) Update(urlPath string, qParms map[string]string, from, qBody Comparable) error {
+	return errors.Wrapf(func() error {
+		url, err := client.buildURL(urlPath, qParms)
+		etag, err := client.getBodyEtag(url, from, err)
+		rq, err := client.buildRequest("PUT", url, ifMatch(etag), qBody, err)
+		rz, err := client.sendRequest(rq, err)
+		return client.getBody(rz, nil, err)
```
nyarly: To my eye, this is actually *more* readable, but I admit that's subjective. There's 3 or 4 occurrences of very similar (but not identical) skeleton methods, so I wanted to highlight how they differ, and strip out as much as possible code that isn't about those differences.

Alternatives like firsterr or `if x, err ...; err != nil` would both require var lines and either 3 times as many lines of code or the `function()` wrappers.

I'd be better with that if I could extra a single method for all of these, but etag handling already would require a function argument to that method (to wire the URL into getting the etag).

Anyway, I'd like to try living with the pattern for a little while and seeing how it works out. I suspect we'll be back to this code in the near term, which'll be a good test for the read-and-reasonability here.

lib/status_poller.go
```diff
@@ -0,0 +1,11 @@
+package sous
+
+type (
+	// StatusPoller polls servers for status
+	StatusPoller struct {
+	}
+)
+
+func (sp *StatusPoller) Start() {
```
nyarly: It's a stub that needs to be filled in. Hopefully by EOD

nyarly:
This is very much quick and dirty. There are a few things here I'm nervous about but cover our immediate use case.

First, if a deploy resolution never completes, there's not yet a timeout or whatever. It should be easy to add, but I figured we'd iterate that when we have call for it.

Second, and more significantly, the StatusPoller as written will get badly confused if a single sous-server is managing more than one cluster. For our current use case, that doesn't seem to be a problem: we'll just deploy one server per cluster. But if we did something like server-per-singularity (i.e. multiple logical clusters) or server-per-network, we'd need to revisit the polling logic.

Third, the DiffResolution Desc value "unchanged" is significant, and we change it at our peril. This is a necessary coupling between client and server, though, and I don't see how to avoid it or something equivalent.

All that said, quick theory of operation:

`sous plumbing status` takes some deploy filter flags and sets up a StatusPoller with the resulting ResolveFilter. The Status poller gets the /servers, and creates a subPoller for each listed URL. The subPollers query their respective server on /status every 500ms, filtering the results based on the filter flags and looking for the case where:

+ that server's GDM includes a matching deployment
+ its list of current diff resolutions includes a "diff unchanged" entry or
+ its list of stable resolutions (i.e. the previous resolve) includes "diff unchanged"

Once all the subpollers have been satisfied, the StatusPoller returns.

**not yet addressed**

First, each subpoller should be considering, probably, a different version tag.

Conversely, it should be possible to watch only a particular cluster (or list of clusters.) 

Finally, this plumbing command needs to be folded into `sous deploy` in order to actually achieve the feature we're trying to accomplish. How that gap closes will dictate which the the previous two items needs to be addressed.

Also, there's still no good tests for the overall polling of status, which will require, I think, a full integration to set up.

lib/httpclient.go
```diff
@@ -0,0 +1,254 @@
...
+	return map[string]string{"If-None-Match": "*"}
+}
+
+func ifMatch(etag string) map[string]string {
+	return map[string]string{"If-Match": etag}
+}
+
+// ****
+
+func (client *HTTPClient) buildURL(urlPath string, qParms map[string]string) (urlS string, err error) {
```
nyarly: Here, my intention was to clarify to purpose of the returned string (i.e. that it's a url), which ties me to naming the error value. Note that the string doesn't stutter.

lib/httpclient.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	qry := url.Values{}
+	for k, v := range qParms {
+		qry.Set(k, v)
+	}
+	URL.RawQuery = qry.Encode()
+	return client.serverURL.ResolveReference(URL).String(), nil
+}
+
+func (client *HTTPClient) getBodyEtag(url string, body Comparable, ierr error) (etag string, err error) {
```
nyarly: Likewise

samsalisbury:
I've spent some time trying to get this to compile @nyarly but I'm having a bit of trouble with it, the easy way to get compilation working breaks some tests because of injection failures. It seems to me that we are expecting the `sous.Client` to satisfy `sous.StateManager` ... or something? Anyway I'll keep trying...

lib/httpclient.go
samsalisbury: To me the 'URL' in the method name 'buildURL' is all that's needed to document the first return value as a URL, as the other returned value is an error and so unlikely to contain the URL.

lib/httpclient.go
samsalisbury: Like above, the method name seems to document what it returns, so I'm not sure the named return parameter adds to understanding of the method. But doesn't hurt other than brevity, so not that important.

samsalisbury:
Addressed in #281 wasn't that difficult in the end.

nyarly:
Merged in #281, fixed a merge-conflict with master.

PR #280, opened 2017-01-03
cmdr cli variable names

alrs:
I'm not looking to merge this yet, I'm looking for review that my understanding of the variable names is correct.


util/cmdr/help.go
```diff
@@ -75,20 +77,28 @@ func commandTable(cs Commands) [][]string {
 // findBottomCommand exists to satisfy this rule: "The arguments to a command
 // can either be values or indicative of a subcommand." It traverses the list
 // of command arguments to find the subcommand furthest down the tree.
-func findBottomCommand(cmd Command, cmdArgs []string) *Command {
+func findBottomCommand(cmd Command, cmdArgs []string) (*Command, int) {
+	var pos int
 	bottomSubCmd := &cmd
-	for _, a := range cmdArgs {
+	for pos = 0; pos < len(cmdArgs); pos++ {
```
samsalisbury: Just trying to work out why you didn't simply replace this line with:

    for pos, bottomSubCmd = range cmdArgs {

That would allow you to remove lines 91-95, and you would also use `bottomSubCmd` in place of `nextToken`. However, I'm not 100% sure about that so I'll pull the code and take a proper look offline...

util/cmdr/help.go
```diff
@@ -75,20 +77,28 @@ func commandTable(cs Commands) [][]string {
 // findBottomCommand exists to satisfy this rule: "The arguments to a command
 // can either be values or indicative of a subcommand." It traverses the list
 // of command arguments to find the subcommand furthest down the tree.
-func findBottomCommand(cmd Command, cmdArgs []string) *Command {
+func findBottomCommand(cmd Command, cmdArgs []string) (*Command, int) {
```
samsalisbury: What is the use case for that second return value? Apologies if I've missed it, just couldn't spot it in this PR. Also, worth adding a note in the godoc about what the second return value means.

util/cmdr/cli.go
```diff
@@ -196,36 +198,36 @@ func (c *CLI) ListSubcommands(base Command) []string {
 	return list
 }
 
-func (c *CLI) prepare(base Command, args []string, ff []func(*flag.FlagSet)) (*PreparedExecution, error) {
-	if len(args) == 0 {
-		return nil, InternalErrorf("command %T received zero args", base)
+func (c *CLI) prepare(cmd Command, cmdArgs []string, flagAddFuncs []func(*flag.FlagSet)) (*PreparedExecution, error) {
```
alrs: Bouncing around the codebase I've been pretty confused as to when I'm looking at os.Args, os.Args[1:], or whatever remains of flagSet.Args()

util/cmdr/cli.go
```diff
@@ -196,36 +198,36 @@ func (c *CLI) ListSubcommands(base Command) []string {
...
 	}
-	name := args[0]
-	args = args[1:]
+	cmdName := cmdArgs[0]
+	cmdArgs = cmdArgs[1:]
 	// Add and parse flags for this command.
-	if command, ok := base.(AddsFlags); ok {
-		if ff == nil {
-			ff = []func(*flag.FlagSet){}
+	if cmdHasFlags, ok := cmd.(AddsFlags); ok {
```
alrs: @Nyarly also mentioned that "has" gets the mind ready for a bool. I was pretty tangled up in the multiple appearances of things that read like AddFlags/AddsFlags/MustAddFlags in the codebase, so I wanted to make it really obvious that this was something else. I've changed the CmdHasInterface variables to CmdWithInterface.

util/cmdr/help.go
alrs: I'm hoping to use this function for cli.go, as well. I'm going to put that ambition aside for the meantime, but having the position of a command in the list was helpful there.

util/cmdr/help.go
nyarly: The pos needs to be captured outside of the `for{}` scope to return it. Otherwise `for range` would be clearer.

util/cmdr/help.go
```diff
@@ -75,20 +77,28 @@ func commandTable(cs Commands) [][]string {
...
 		// check if the command has any subcommands
 		testCmd := *bottomSubCmd
 		hasSubCmd, ok := testCmd.(Subcommander)
 		if !ok {
-			return bottomSubCmd
+			// there are no more subcommands, this loop is done.
+			return bottomSubCmd, pos
 		}
-		childCmd, ok := hasSubCmd.Subcommands()[a]
+		if pos == len(cmdArgs) {
```
nyarly: If `pos<len(cmdArgs)` (loop condition) then it `pos == len(cmdArgs)` can never be true. I think you can pull this out.

util/cmdr/help.go
nyarly: I think if we're not going to use that return value, better to pull it out (which then simplifies your loop further down.)

util/cmdr/help.go
samsalisbury: You can still capture a range index outside of the range, that's why my example code used `=` rather than `:=`. See this example, or [run it in Go playgrounds.](https://play.golang.org/p/LJm_tzjmoW)

```go
package main

import (
	"fmt"
)

func main() {
	var a, b, c = 1, "", []string{"die", "bart", "die"}
	for a, b = range c {
		fmt.Printf("%d: %q\n", a, b)
	}
	fmt.Printf("Printed %d strings.", a+1)
}
```

PR #282, opened 2017-01-05
Sous status

nyarly:
Accomplished here: config/serving named servers.

There's some work towards a test of the StatusPoller happy path - it's the only
test currently failing. :)

Goal would be to finish that test (mostly around constructing the right
statusJSON, I think) and confirm that the statuspoller itself works. Incidentally, constructing the JSON by hand is something of an experiment. It may be more effective simply to arrange the appropriate structs and serialize them... We lose some "does it handle incomplete or weird input" there, but I'm unsold on the value.

After that, I'd like to proceed to adding some tests for intermediate states,
making calls directly to StatusPoller/subPoller methods.

### DIFF
--- a/cli/sous_update_test.go
+++ b/cli/sous_update_test.go
@@ -3,45 +3,21 @@ package cli
 import (
 	"testing"
 
-	"github.com/opentable/sous/config"
 	"github.com/opentable/sous/graph"
 	sous "github.com/opentable/sous/lib"
 )
 
-var getIDsTests = []struct {
-	Flags       config.DeployFilterFlags
+type getIDsTestCase struct {
+	Flags       sous.ResolveFilter
 	SL          sous.ManifestID
 	ExpectedSID sous.SourceID
 	ExpectedDID sous.DeployID
 	ExpectedErr string
-}{
-	{
-		ExpectedErr: "You must select a cluster using the -cluster flag.",
-	},
-	{
-		Flags:       config.DeployFilterFlags{Cluster: "blah"},
-		ExpectedErr: "You must provide the -tag flag.",
-	},
-	{
-		Flags:       config.DeployFilterFlags{Cluster: "blah", Tag: "nope"},
-		ExpectedErr: `Version "nope" not valid: unexpected character 'n' at position 0`,
-	},
-	{
-		Flags:       config.DeployFilterFlags{Cluster: "blah", Tag: "nope"},
-		ExpectedErr: `Version "nope" not valid: unexpected character 'n' at position 0`,
-	},
-	{
-		Flags:       config.DeployFilterFlags{Cluster: "blah", Tag: "1.0.0"},
-		SL:          sous.MustParseManifestID("github.com/blah/blah"),
-		ExpectedSID: sous.MustParseSourceID("github.com/blah/blah,1.0.0"),
-		ExpectedDID: sous.DeployID{Cluster: "blah",
-			ManifestID: sous.ManifestID{Source: sous.SourceLocation{Repo: "github.com/blah/blah"}}},
-	},
 }
 
 func TestGetIDs(t *testing.T) {
-	for _, test := range getIDsTests {
-		sid, did, err := getIDs(test.Flags, test.SL)
+	test := func(test getIDsTestCase) {
+		sid, did, err := getIDs(&test.Flags, test.SL)
 		if err != nil {
 			if test.ExpectedErr == "" {
 				t.Error(err)
@@ -62,6 +38,29 @@ func TestGetIDs(t *testing.T) {
 			t.Errorf("got DeployID %q; want %q", did, test.ExpectedDID)
 		}
 	}
+
+	test(getIDsTestCase{
+		ExpectedErr: "update: You must select a cluster using the -cluster flag.",
+	})
+	test(getIDsTestCase{
+		Flags:       sous.ResolveFilter{Cluster: "blah"},
+		ExpectedErr: "update: You must provide the -tag flag.",
+	})
+	test(getIDsTestCase{
+		Flags:       sous.ResolveFilter{Cluster: "blah", Tag: "nope"},
+		ExpectedErr: `update: Version "nope" not valid: unexpected character 'n' at position 0`,
+	})
+	test(getIDsTestCase{
+		Flags:       sous.ResolveFilter{Cluster: "blah", Tag: "nope"},
+		ExpectedErr: `update: Version "nope" not valid: unexpected character 'n' at position 0`,
+	})
+	test(getIDsTestCase{
+		Flags:       sous.ResolveFilter{Cluster: "blah", Tag: "1.0.0"},
+		SL:          sous.MustParseManifestID("github.com/blah/blah"),
+		ExpectedSID: sous.MustParseSourceID("github.com/blah/blah,1.0.0"),
+		ExpectedDID: sous.DeployID{Cluster: "blah",
+			ManifestID: sous.ManifestID{Source: sous.SourceLocation{Repo: "github.com/blah/blah"}}},
+	})
 }
 
 var updateStateTests = []struct {
@@ -140,13 +139,15 @@ type DummyStateManager struct{}
 func (dsm *DummyStateManager) WriteState(s *sous.State) error  { return nil }
 func (dsm *DummyStateManager) ReadState() (*sous.State, error) { return nil, nil }
 
+//XXX should actually drive interesting behavior
 func TestSousUpdate_Execute(t *testing.T) {
 	dsm := &DummyStateManager{}
 	su := SousUpdate{
-		StateReader: graph.StateReader{StateReader: dsm},
-		StateWriter: graph.StateWriter{StateWriter: dsm},
-		GDM:         graph.CurrentGDM{Deployments: sous.MakeDeployments(0)},
-		Manifest:    graph.TargetManifest{Manifest: &sous.Manifest{}},
+		StateReader:   graph.StateReader{StateReader: dsm},
+		StateWriter:   graph.StateWriter{StateWriter: dsm},
+		GDM:           graph.CurrentGDM{Deployments: sous.MakeDeployments(0)},
+		Manifest:      graph.TargetManifest{Manifest: &sous.Manifest{}},
+		ResolveFilter: &sous.ResolveFilter{},
 	}
 	su.Execute(nil)
 }
--- a/ext/docker/image_mapping_test.go
+++ b/ext/docker/image_mapping_test.go
@@ -117,8 +117,8 @@ func TestReharvest(t *testing.T) {
 		AllNames:      []string{cn, in},
 	})
 	nc.dump(os.Stderr)
-	Log.Debug.SetOutput(os.Stderr)
-	Log.Vomit.SetOutput(os.Stderr)
+	//Log.Debug.SetOutput(os.Stderr)
+	//Log.Vomit.SetOutput(os.Stderr)
 	err = nc.GroomDatabase()
 	assert.NoError(err)
 	nc.dump(os.Stderr)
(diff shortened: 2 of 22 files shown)


nyarly:
This is, pending automatic review, ready to go.

server/handle_servers.go
```diff
@@ -12,7 +12,8 @@ type (
 	}
 
 	server struct {
-		URL string
+		ClusterName string
```
alrs: Same issue as noted previously with adding members to a struct: Is there any code that expects to s := &server{"http://example.com"}?

lib/status_poller.go
```diff
@@ -18,7 +18,8 @@ type (
 
 	// copied from server - avoiding coupling to server implemention
 	server struct {
-		URL string
+		ClusterName string
```
nyarly: I refer to http://labix.org/gopkg.in (the bottom discusses semver and structs). Since `server` itself isn't exported, it's not much of an issue. The purpose here is to serialize to and from JSON, and for this very reason the client has a corresponding struct rather than e.g. both drawing from a DTO package.

server/handle_servers.go
nyarly: None exists. Generally speaking I'd also worry about code being written, but it's considered bad form. `go vet` scolds you, for instance.

ext/docker/image_mapping_test.go
```diff
@@ -117,8 +117,8 @@ func TestReharvest(t *testing.T) {
 		AllNames:      []string{cn, in},
 	})
 	nc.dump(os.Stderr)
-	Log.Debug.SetOutput(os.Stderr)
-	Log.Vomit.SetOutput(os.Stderr)
+	//Log.Debug.SetOutput(os.Stderr)
```
nyarly: Honestly, I'd like to have a nicer way to do something like this. I find that I often want to enable debugging again anywhere that I've enabled it once, so I habitually comment out this kind of code rather than deleting it outright. In the actual production code, I'd be more loath to let it linger, but I'm less concerned in testing code.

lib/status_poller.go
samsalisbury: I concur, since it's not exported we don't need to worry here, and serialisation of structs to/from to JSON  doesn't care about order. In general, as @nyarly noted below, we should not use positional members to create values of types exported from other packages.

server/handle_servers.go
samsalisbury: True, go vet scolds. Furthermore, since the number of fields have changed from 1 to 2, compilation of `s := &server{"http://example.com"}` would also fail now, since when using positional field values you must specify every field, both exported and non-exported.

lib/httpclient.go
```diff
@@ -246,7 +251,13 @@ func (client *HTTPClient) httpRequest(req *http.Request) (*http.Response, error)
 
 	rz.Body = NewReadDebugger(rz.Body, func(b []byte, n int, err error) {
 		Log.Vomit.Printf("Client <- %s %q: %d", req.Method, req.URL, rz.StatusCode)
-		Log.Vomit.Print(string(b))
+		comp := &bytes.Buffer{}
+		if err := json.Compact(comp, b); err != nil {
+			Log.Vomit.Print(string(b))
```
samsalisbury: I feel like we should be printing the error here as well since we captured it?

samsalisbury: Can we also return early here to avoid the `else` case and resultant indentation? See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

lib/status_poller.go
```diff
@@ -188,15 +185,17 @@ func (sub *subPoller) start(rs chan statPair, done chan struct{}) {
...
 		select {
-		case <-time.Tick(time.Second / 2):
+		case <-ticker.C:
 			stat := sub.pollOnce()
 			rs <- statPair{url: sub.URL, stat: stat}
 			if stat >= ResolveComplete {
 				return
 			}
 		case <-done:
+			ticker.Stop()
```
samsalisbury: I wonder if this should be done using `defer` after `NewTicker` on line 188? Otherwise we leak the ticker on line 195.

lib/status_poller_test.go
```diff
@@ -5,14 +5,37 @@ import (
...
+	checkString(ResolveErred, "ResolveErred")
+	checkString(ResolveComplete, "ResolveComplete")
+	checkString(ResolveState(1e6), "unknown (oops)")
+
+	for rs := ResolveNotStarted; rs <= ResolveComplete; rs++ {
+		if rs.String() == "unknown (oops)" {
+			t.Errorf("ResolveState %d doesn't have a string", rs)
+		}
+	}
+}
```
samsalisbury: Heh, this test made me chuckle, but I guess these strings are part of the public interface, so making changing them hard is a good idea.

lib/status_poller_test.go
```diff
@@ -93,6 +120,7 @@ func TestStatusPoller(t *testing.T) {
 	otherSrv := httptest.NewServer(http.HandlerFunc(h))
 
 	repoName := "github.com/opentable/example"
+
 	serversJSON = []byte(`{
```
samsalisbury: Unless we are explicitly testing serialisation, I agree with your earlier suggestion that we may be better off using sous types to construct this object, and serialise it in the test. Obvs not blocking over this, just seeing the bug fix for SourceID here made it clearer that we might be best off going that way. I guess we should have some kind of integration test that checks bytes-in and bytes-out to make sure we don't accidentally change the public interface though.

cli/sous_update_test.go
```diff
@@ -62,6 +38,29 @@ func TestGetIDs(t *testing.T) {
...
+		Flags:       sous.ResolveFilter{Cluster: "blah", Tag: "nope"},
+		ExpectedErr: `update: Version "nope" not valid: unexpected character 'n' at position 0`,
+	})
+	test(getIDsTestCase{
+		Flags:       sous.ResolveFilter{Cluster: "blah", Tag: "1.0.0"},
+		SL:          sous.MustParseManifestID("github.com/blah/blah"),
+		ExpectedSID: sous.MustParseSourceID("github.com/blah/blah,1.0.0"),
+		ExpectedDID: sous.DeployID{Cluster: "blah",
+			ManifestID: sous.ManifestID{Source: sous.SourceLocation{Repo: "github.com/blah/blah"}}},
+	})
```
samsalisbury: Interesting pattern, do you find this easier to read than the old slice of structs [TableDrivenTests](https://github.com/golang/go/wiki/TableDrivenTests) approach? To me it's exactly equivalent. One issue with the old implementation that persists with this one afaik is that we did not print out the index of the test case, making it slightly harder to track down which one failed. I guess now we would get a better stack trace, but that's not printed by default in the test output, so each failure would be attributed to one of the lines of the `test` func.

I think this pattern would be better if the test func returned an error, and we wrapped each case in

    if err := test(...); err != nil {
        t.Error(err)
    }

As that would tell us exactly which test case failed, by line number.

Alternatively, the original slice of structs approach could be modified to print the index of the failing test case. Not blocking over any of this, but want to have the discussion. I would like our tests to be more uniform in general, as we have quite a few patterns in use at the moment, but I guess it's a hard problem to solve universally.

lib/httpclient.go
```diff
@@ -231,7 +231,12 @@ func (client *HTTPClient) httpRequest(req *http.Request) (*http.Response, error)
 	} else {
 		req.Body = NewReadDebugger(req.Body, func(b []byte, n int, err error) {
 			Log.Vomit.Printf("Client -> %s %q", req.Method, req.URL)
-			Log.Vomit.Print(string(b))
+			comp := &bytes.Buffer{}
+			if err := json.Compact(comp, b); err != nil {
+				Log.Vomit.Print(string(b))
```
samsalisbury: I wonder if we should print the error since we captured it?

ext/docker/image_mapping_test.go
samsalisbury: Yeah I prefer not to see commented code either, but we have limited debugging options at the moment, so I try to ignore these as they do sometimes come in handy. Ideally we should use something like [godebug](https://github.com/mailgun/godebug) or else have fine-grained control of debug output via flags or env vars. Until we have worked out a decent alternative though, I'm happy to leave this stuff in.

nyarly: I'm pretty much on board with all of this. I've actually started to contemplate writing a trashy package like "quickdebug" or something that's a thin wrapper on log so that it would be easy to test if it'd been left in (i.e. grep for "quickdebug" and fail CI if it's there.) That would also obviate the need to do SetOutput and SetFlags calls, so I wouldn't feel like there was value to leaving this stuff in.

Complete pie in the sky: Erlang has a library in their stdlib called 'dbg', where you can essentially write a little filter program to instrument your code after the fact. For instance, you can say "I'd like you to log every call to a method in package X, with it's arguments" or "When X:Y/3 is called, log its arguments, and also log when it returns, and its return value." Having used it a bit, I've found that it's *exactly* what I want in any println style debugging, doesn't have the drawback of littering code with extraneous logging lines, and encourages function modularity (because if there's a point in the code you'd like to log, it's easiest to extract a function at that point, and give it a name that will be meaningful in logging.) In the end, I prefer it  100% to an inline debugger for most circumstances. Given that there's mechanisms for instrumenting go code (i.e. the way gofmt and go cover work), I wonder if it would be possible to write such a tool for Go... in our copious free time :)

lib/httpclient.go
nyarly: The reason to capture an error here was just to ensure that we logged *something;* as I comment, though, if the JSON doesn't compact for logging, it'd be nice to have some explanation as to why.

lib/httpclient.go
nyarly: Not without breaking the body logging code out into its own function, which might itself be worth doing. It's quick enough, I might issue a PR for this in the next 5-10, though.

lib/status_poller.go
nyarly: That's absolutely how this should be done.

lib/status_poller_test.go
nyarly: To be honest, most of these were written to effect a respectable level of coverage. As I was writing them, I realized that there was the case (that I'd worried about in the first place) that the iota constants might not stay in sync with their String(), which is where this came from. The Strings are used only for user information, but that is a public interface...

lib/status_poller_test.go
nyarly: Yeah, I wouldn't write another test this way, but it was a good reminder that SourceLocations serialize as 'repo,offset' not { Repo: ,Offset: }

cli/sous_update_test.go
nyarly: The motivation for switching this over was actually that I was getting segfault panics. Because of the table driven approach, the source of the failure wasn't immediately apparent. Converting to a "function table" meant the panic backtrace led directly to the issue (the switch to a pointer meant that the table was passing nil instead of a zero struct...  once I realized what the issue is I changed the test code to pass an address of it's zero rather than nil.)

I actually really like the example given in [TableDrivenTests](https://github.com/golang/go/wiki/TableDrivenTests) (point of fact, I used table tests for years in Ruby when it was appropriate). But I think there are some salient details about that example that make it useful and appropriate. The crux of the matter is that the actual test loop has only one conditional, which is the test itself. Possibly, multiple test conditionals would be okay (i.e. if you were checking that an Equal() method is reflexive...), but the practice of examining the table case row to figure out what kind of test this is makes following the tests much harder. (There's a whole side conversation to be had about WET vs DRY here.) Related, the example there is able to use anonymous structs, because all the fields are always assigned, which makes the "table" much clearer. If nothing else "you can use anonymous structs" serves as a rule of thumb for "the test loop is clear enough."

What I've found with table tests in general is that I usually reach for them when I have many related test cases, but that I usually realize that the test cases differ *just enough* that the table driver gets too complicated.

Big picture, I've been trying to set aside the feeling that Go's stdlib testing package is too spare, but the variations in testing patterns springs from trying to cover its (IMO) deficiencies. Specifically, I think a testing tool needs direct support for [Four Phase Testing](https://robots.thoughtbot.com/four-phase-test) - table tests, for instance, are a way to reuse a pair of setup/verify phases across different exercises. If I were starting a large scale Go project from scratch, today, I'd probably insist on something like Testify or Goconvey. But I don't think we should try to backport that into our existing Sous tests, and I think there's valuable experience with Go testing still to be had by doing things from scratch.

PR #285, opened 2017-01-10
Multiple binary builds from Linux

alrs:
This Makefile is destined to run in travis-ci to deploy multi-arch binaries as a Github release.


nyarly:
Am I missing something?

As is, this'll cause a fail. The .travis.yml finishes test runs with "bin/publish", which calls out to bin/build, which is removed here.

*Probably* it's best to use Travis' deploy: key with multiple files. Worst case, bin/publish should swap out a call to make, right?

alrs:
@nyarly Right, this only partially covers the build process. In my testing, the current artifacts being created by these scripts: 1. Aren't tarballs, appear to just be .gzipped binaries 2. Don't run in Linux as they are supposed to 3. Version information injection via -ldflag doesn't look like it actually injects the version number.

samsalisbury:
Great stuff, I've just been testing this on my local machine, looks good, the version injection does appear to be working for the OSX build at least. I'm seeing gzipped tarballs containing an `artifacts` dir, containing `sous-VERSION`. See contents below.

So, the things I notice are that we have both binaries in the same tarball, and an extraneous `artifacts` directory. I'm going to take a look now to see if I can sort these out, as getting this working will make getting new people set up on Sous much easier so I'm eager to get it done asap :)

I'll also replace `bin/publish` with `make release` and look at the Travis config for publishing the binaries.

```shell
# Contents of sous-0.1.0-sam.tar.gz
.
├── artifacts
│   └── sous-0.1.0-sam
│       ├── LICENSE
│       ├── README.md
│       ├── alternatives.md
│       ├── architecture.md
│       ├── bridge-networking.md
│       ├── client-config.md
│       ├── cluster_and_container_iface.md
│       ├── code.md
│       ├── contributions.md
│       ├── deployment-workflow.md
│       ├── design.md
│       ├── disjoint-rectification.md
│       ├── features.md
│       ├── first-deployment-of-sous.md
│       ├── maxims_and_arrows.md
│       ├── ontology.md
│       ├── rationale.md
│       ├── release.md
│       ├── required-infrastructure.md
│       ├── setting-up-go.md
│       ├── sous-darwin-10.6-amd64
│       ├── sous-linux-amd64
│       ├── sous0.1.md
│       └── what-is-sous.md
```

PR #294, opened 2017-01-17
(wip) DI update

samsalisbury:
This needs a bit more tidy up, the idea was to remove the need for graph factories and replace them with a working `Psyringe.Clone` call (via newly-vendored psyringe library). To get this working, I added a method `BuildRequestGraph` which derives a request-scoped graph from the process-scoped main graph, this is used throughout. The aim of this change is that it's a refactoring, and should not affect functionality, but hopefully it will be a bit easier to reason about. Opening a PR early, more commits to come...

### DIFF
--- a/server/graph.go
+++ b/server/graph.go
@@ -13,8 +13,8 @@ type (
 )
 
 // AddsPerRequest registers items into a SousGraph that need to be fresh per request
-func AddsPerRequest(g Injector) {
-	g.Add(liveGDM)
+func AddsPerRequest(g Injector, sr sous.StateReader) {
+	g.Add(liveGDM, sr)
 }
 
 func liveGDM(sr graph.StateReader) (*LiveGDM, error) {
--- a/server/routemap.go
+++ b/server/routemap.go
@@ -27,7 +27,7 @@ type (
 	// RouteMap is a list of entries for routing
 	RouteMap []routeEntry
 
-	// A ResourceFamily bundles up the exchangers that deal with a kind of resources
+	// A Resource bundles up the exchangers that deal with a kind of resources
 	// (n.b. that properly, URL == resource, so a URL pattern == many resources
 	Resource interface{}
 
--- a/server/routemap.go
+++ b/server/routemap.go
@@ -60,10 +60,11 @@ type (
 	*/
 )
 
-func (rm *RouteMap) buildMetaHandler(r *httprouter.Router, grf func() Injector) *MetaHandler {
+func (rm *RouteMap) buildMetaHandler(r *httprouter.Router, processGraph, requestGraph Injector) *MetaHandler {
 	ph := &StatusMiddleware{}
 	mh := &MetaHandler{
-		graphFac:      grf,
+		processGraph:  processGraph,
+		requestGraph:  requestGraph,
 		router:        r,
 		statusHandler: ph,
 	}
@@ -72,10 +73,11 @@ func (rm *RouteMap) buildMetaHandler(r *httprouter.Router, grf func() Injector) 
 	return mh
 }
 
-// BuildRouter builds a returns an http.Handler based on some constant configuration
-func (rm *RouteMap) BuildRouter(grf func() Injector) http.Handler {
+// BuildRouter builds a returns an http.Handler based on some constant
+// configuration.
+func (rm *RouteMap) BuildRouter(processGraph, requestGraph Injector) http.Handler {
 	r := httprouter.New()
-	mh := rm.buildMetaHandler(r, grf)
+	mh := rm.buildMetaHandler(r, processGraph, requestGraph)
 
 	for _, e := range *rm {
 		get, canGet := e.Resource.(Getable)
--- a/server/router.go
+++ b/server/router.go
@@ -14,14 +14,15 @@ import (
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/opentable/sous/lib"
+	"github.com/samsalisbury/psyringe"
 )
 
 type (
 	// The MetaHandler collects common behavior for route handlers.
 	MetaHandler struct {
-		router        *httprouter.Router
-		graphFac      GraphFactory //XXX This is a workaround for a bug in psyringe.Clone()
-		statusHandler *StatusMiddleware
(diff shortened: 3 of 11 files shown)


server/routemap.go
```diff
@@ -27,7 +27,7 @@ type (
 	// RouteMap is a list of entries for routing
 	RouteMap []routeEntry
 
-	// A ResourceFamily bundles up the exchangers that deal with a kind of resources
+	// A Resource bundles up the exchangers that deal with a kind of resources
```
nyarly: This is a common misnomer in REST-related server code I wish we could avoid. A Resource is identified by a URL in RESTful HTTP, but a ResourceFamily(that was) manages URLs that match a pattern, including handling URLs that differ by `?&` parameters: i.e. a potentially infinite number of unique Resources that are related in an application-dependent way.

server/server.go
```diff
@@ -4,24 +4,29 @@ import (
...
+func New(laddr string, processGraph, requestGraph Injector) *http.Server {
 	return &http.Server{
 		Addr:    laddr,
-		Handler: SousRouteMap.BuildRouter(gf),
+		Handler: SousRouteMap.BuildRouter(processGraph, requestGraph),
 	}
 }
 
+// BuildRequestGraph derives a per-request scoped DI graph.
+func BuildRequestGraph(processGraph Injector) Injector {
```
nyarly: Part of the reason to isolate the code in server.go is that it'll be difficult to cover (there should be a comment to this effect...) - can we relocate BRG to another file since it should be easier to get a unit test around?

server/router.go
```diff
@@ -118,27 +119,33 @@ func (mh *MetaHandler) PutHandling(factory ExchangeFactory) httprouter.Handle {
...
+	requestGraph.Add(&ResponseWriter{ResponseWriter: w}, r, p, parseQueryValues)
+	return requestGraph
 }
 
 func (mh *MetaHandler) injectedHandler(factory ExchangeFactory, w http.ResponseWriter, r *http.Request, p httprouter.Params) Exchanger {
 	h := factory()
 
 	exGraph := mh.ExchangeGraph(w, r, p)
+	// Inject process-scoped stuff.
+	mh.processGraph.MustInject(h)
```
nyarly: Midway through the changes I was making last week, this double-DI solution occurred to me. At the time, the fact that it required as much per-scope setup put me off. I think, also, the idea of passing around more than one DI didn't occur to me. Also, I was averse to further expanding the `Injector` interface and further coupling the `server` and `psyringe` packages.

server/router.go
```diff
@@ -118,27 +119,33 @@ func (mh *MetaHandler) PutHandling(factory ExchangeFactory) httprouter.Handle {
...
 }
 
+// ExchangeGraph returns a per-request Injector.
 func (mh *MetaHandler) ExchangeGraph(w http.ResponseWriter, r *http.Request, p httprouter.Params) Injector {
-	g := mh.graphFac()
-	g.Add(&ResponseWriter{ResponseWriter: w}, r, p)
-	g.Add(parseQueryValues)
-	return g
+	// This Clone is very important so we get a fresh injector for each request.
+	// This type assertion is nasty, going away soon. - Sam
```
nyarly: *Actually* it seems to me that we get little value out of this Clone: all the values in the requestGraph are constructed, and none should be Injected between its New and this Clone. We could use BuildRequestGraph here instead of Clone and avoid the interface leak.

server/server.go
samsalisbury: Definitely, I had this in mind as a a tidy-up task .

server/graph.go
```diff
@@ -13,8 +13,8 @@ type (
 )
 
 // AddsPerRequest registers items into a SousGraph that need to be fresh per request
-func AddsPerRequest(g Injector) {
-	g.Add(liveGDM)
+func AddsPerRequest(g Injector, sr sous.StateReader) {
+	g.Add(liveGDM, sr)
```
samsalisbury: Yup, sorry, the (wip) was meant to signify that there is still work to do here, I appreciate the review regardless though, cheers.

server/server_injection_test.go
```diff
@@ -12,22 +12,23 @@ import (
...
+	processGraph.Add(&config.Verbosity{})
+	processGraph.Add(&config.DeployFilterFlags{Cluster: "test"})
+	processGraph.Add(graph.DryrunBoth)
 
-	gf := func() Injector {
-		return g.Clone()
-	}
+	requestGraph := psyringe.New(liveGDM, func() sous.StateReader {
+		return &sous.DummyStateManager{}
+	})
```
samsalisbury: Re "Scoped values" I actually started implementing this yesterday in psyringe, and I think there's quite a simple, neat solution to this. I didn't want to spend too much time on it before diving into the issues in Sous, but since you also recognise the need for this I might try to get that working today if I get time after the stories on the board.

server/server.go
```diff
@@ -4,24 +4,29 @@ import (
...
+	// Get a copy of StateReader to pass down to requestGraph.
+	// TODO: Make child scopes a feature of psyringe.
+	stateReaderGrabber := &struct{ graph.StateReader }{}
+	processGraph.Inject(stateReaderGrabber)
+	// Create requestGraph, which is to be cloned for each request.
+	return psyringe.New(liveGDM, stateReaderGrabber.StateReader)
```
samsalisbury: Yes, I agree, now Clone is working as intended we shouldn't need this extra type & constructor.

server/router.go
samsalisbury: We can't easily put `Clone` into that interface without generics, since `Psyringe.Clone`returns a `*Psyringe` not an `Injector`. The Clone here was needed to clone the `LiveGDM` constructor so that it gets called afresh each time.

server/router.go
samsalisbury: I agree we should avoid that coupling where possible, or at least limit it to a single file. I'll keep looking into this when I get a chance, hopefully later today.

alrs:
This PR is now three months old, so I've created a Jira ticket for the day this work is picked back up again.

https://opentable.atlassian.net/browse/TECHOPS-2254

PR #295, opened 2017-01-17
Makefile Improvements.

alrs:
Remove bin/install-dev

Remove bin/publish

Remove legendary from bin/cover


bin/cover
```diff
@@ -1,4 +1,2 @@
 #!/usr/bin/env bash
 engulf -s --coverdir=/tmp/sous-cover --exclude='/vendor/,/integration/?,/bin/?,/dev_support/?,/util/test_with_docker/?,/examples/?,/util/cmdr/cmdr-example/?' --exclude-files='_generated.go$,raw_client.go$' --merge-base=_merged.txt ./...
```
nyarly: I'd think that the rest of bin/cover should also become a make target, right?

alrs: Agreed, that was step #2.

alrs:
Trying to parallelize ci-setup was a bridge too far at this late hour, so I'm reverting back to parallel cover and test-three-tries on a new PR.

PR #296, opened 2017-01-17
Documenting the "proper" and "inverted" Sous workflows.

nyarly:


doc/sous_drives_builds.md
```diff
@@ -0,0 +1,132 @@
...
+in the project source repository.
+
+## More Pleasant Than True
+
+Unfortunately,
+the present implementation of Sous takes
+the resulting Docker container
+and deploys that.
+One consequence is that
+the deployment toolchain for each service
```
samsalisbury: Maybe "build toolchain" rather than "deployment toolchain" ?

doc/sous_drives_builds.md
```diff
@@ -0,0 +1,132 @@
...
+Push all your build chain into your Dockerfile,
+so that you copy your source code in,
+and use your build products as
+your Docker "CMD."
+Second, you can maintain a current build process,
+possibly changing whatever steps build an actual container
+with `sous build`.
+Since Sous is designed as the driver of the build process,
+the latter approach is refered to as the "inverted" workflow.
+
```
samsalisbury: Great to have this written down clearly, cheers.

doc/sous_drives_builds.md
```diff
@@ -0,0 +1,132 @@
...
+The advantage of the "inverted" approach is that
+you may well already be doing it,
+so you may be able to use Sous with very little change
+to your existing workflow.
+Many build chains can be converted by changing one line like
+`docker build` to
+`sous build`.
+The disadvantages have to do with work required in the future
+to retrofit to the buildpack solution,
+as well as missed opportunities to share a build chain.
```
samsalisbury: As well as this approach not working well when the host OS/Arch differs from that used in deployment. I think another benefit of Sous' approach will also be cross-platform builds as standard by performing all stages of a build inside Docker containers. So you'll get the same final Docker image whether building on Mac, Linux, Windows (in theory).

PR #299, opened 2017-01-18
sous init -use-otpl includes owners

samsalisbury:
UPDATE: No longer work in progress.

This addresses an issue where `sous init -use-otpl-deploy` ignores owners specificed in the otpl-deploy-scripts `singularity-request.json` file.

Unfortunately this feature was originally scoped to return `DeploySpecs` only, so the change is significant in size, changing this code to deal with `Manifest` as its basic unit.


ext/otpl/otpl.go
```diff
@@ -56,20 +55,20 @@ func (sr SingularityResources) SousResources() sous.Resources {
...
 	Spec *sous.DeploySpec
 }
 
-// GetDeploySpecs searches the working directory of wd to find otpl-deploy
+// ParseManifest searches the working directory of wd to find otpl-deploy
 // config files in their standard locations (config/{cluster-name}), and
 // converts them to sous.DeploySpecs.
-func (dsp *DeploySpecParser) GetDeploySpecs(wd shell.Shell) sous.DeploySpecs {
+func (dsp *ManifestParser) ParseManifest(wd shell.Shell) *sous.Manifest {
 	wd = wd.Clone()
```
nyarly: I think we'd be better prepared to extract this code into its own tool if we used https://golang.org/pkg/path/filepath/#Walk in preference to the internal `shell` package.

ext/otpl/otpl_test.go
```diff
@@ -47,5 +53,97 @@ func TestSingularityResources_SousResources(t *testing.T) {
...
+
+	if different, diffs := actual.Diff(expected); different {
+		t.Errorf("parsed manifest not as expected")
+		for _, diff := range diffs {
+			t.Error(diff)
+		}
+	}
+}
+
+type fileMap map[string]string
```
nyarly: I really like this pattern for testing; I wonder if we should promote it to an internal package?

ext/otpl/otpl_test.go
```diff
@@ -47,5 +53,97 @@ func TestSingularityResources_SousResources(t *testing.T) {
...
+
+func (f fileMap) Delete(dir string) error {
+	return os.RemoveAll(dir)
+}
+
+func (f fileMap) Write(dir string) error {
```
nyarly: From experience, Write should start with a Delete. Otherwise bombed or cancelled test runs leave garbage in the FS and the tests aren't actually starting from a clean slate. Granted, Go's defer much reduces that possibility (when everyone remembers to `defer Delete`) but the failure mode is really easy to miss and really hard to decipher when it happens.

Maaaaybe there should be a WriteMore method (and then Write is Delete; WriteMore), but my suspicion is YAGNI (and you wouldn't really want it anyway: better to have all your test's files in one fileMap.)

ext/otpl/otpl.go
```diff
@@ -56,20 +55,20 @@ func (sr SingularityResources) SousResources() sous.Resources {
...
 	Name string
 	Spec *sous.DeploySpec
 }
 
-// GetDeploySpecs searches the working directory of wd to find otpl-deploy
+// ParseManifest searches the working directory of wd to find otpl-deploy
 // config files in their standard locations (config/{cluster-name}), and
 // converts them to sous.DeploySpecs.
-func (dsp *DeploySpecParser) GetDeploySpecs(wd shell.Shell) sous.DeploySpecs {
+func (dsp *ManifestParser) ParseManifest(wd shell.Shell) *sous.Manifest {
```
nyarly: Receiver name should change.

(This is, btw, my biggest salient resistance to the Go idiom of "don't use `self`" - `self` *does* describe the variable as the method receiver, and the habit we get into is initializations of the type name, which are so meaningless that when we change the name we don't bother updating the variable names.)

ext/otpl/otpl.go
```diff
@@ -95,33 +95,39 @@ func (dsp *DeploySpecParser) GetDeploySpecs(wd shell.Shell) sous.DeploySpecs {
 				dsp.debug(err)
 				return
 			}
-			if otplConfig := dsp.GetSingleDeploySpec(wd); otplConfig != nil {
+			if otplConfig, owners := dsp.GetSingleDeploySpec(wd); otplConfig != nil {
 				name := path.Base(wd.Dir())
 				c <- namedDeploySpec{name, otplConfig}
+				for o := range owners {
+					manifestOwners.Add(o)
```
nyarly: This isn't safe, because manifestOwners is a struct, and Add writes to it.

ext/otpl/otpl.go
samsalisbury: Follow-up PR coming to rename receiver to 'mp'. I can see your point about 'self', when the receiver is a pointer as it is here, but not when it's a non-pointer value. I don't find it painful enough to jump to 'self' just yet.

What I would really like is for `gorename` to allow renaming all receivers on methods of a type at once. And also to rename godoc comments at the same time as the tokens they relate to.

ext/otpl/otpl.go
samsalisbury: Yeah possibly, not going down that path yet, as the concurrent thing we have is probably marginally faster, and working, but not married to it staying that way forever.

ext/otpl/otpl_test.go
samsalisbury: Done, see https://github.com/opentable/sous/pull/326

PR #300, opened 2017-01-19
Edge cases for status checks

nyarly:
Pollers aren't created for clusters that don't have a deployment assigned
to them.

Errors returned in the ResolveStatus are checked to see if there's a chance
they'll clear on their own; if not, we treat the failure as permanent.

ResolveStatus got some extra values, including a MAX (so that the tests can
confirm there's a string value for everything without knowing what the
"top" status number is) and TERMINALS (which allows for multiple end-state
cases.)

The StatusPoller comments expanded to describe that process a bit better.


lib/resolve_errors.go
```diff
@@ -60,6 +62,41 @@ func (re *ResolveErrors) Error() string {
...
+		// create a request or deploy. This might be the result of a conflicting
+		// Request name, in which case it's likely that the next attempt to resolve
+		// will be a Modify instead.
+		return true
+	case *DeleteError:
+		// XXX While "deletes" are no-ops, there's no chance that a DeleteError is going to "self correct"
+		//		return true
+		return false // XXX
+	}
+}
```
samsalisbury: Thanks for this documentation, it makes a huge difference to me when reviewing.

Interestingly the Go standard lib has a fairly common `Temporary() bool` interface which is often implemented on error types, e.g. `net.Error`. Not saying we should use it, but I always tend to mimic standard library language/names where possible, as it's more likely others will immediately understand it. However, this switch statement is extremely clear and readable, I'm not sure a `Temporary` method would add much here.

lib/status_poller.go
```diff
@@ -22,6 +22,11 @@ type (
 		URL         string
 	}
 
+	// copied from server
```
samsalisbury: I wonder if these structs copied from `server` would be better off defined primarily in `sous` (`./lib`) and consumed by `server`. Since these two packages are part of the same codebase, I can't see a strong reason not to couple them, other than that it would cause a circular dependency whilst they are defined in `server` rather than `sous`.

PR #303, opened 2017-01-20
Adds install-latest.sh

samsalisbury:
This is an extremely rough install script just to get people off the ground. It sniffs out the latest release from GitHub, downloads, extracts, symlinks into `/usr/local/bin`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,10 @@ and copy the `sous` executable into your `$PATH`
 You'll also find a copy of the complete
 Sous documentation in that archive.
 
+For convenience, we provide an install script, you can run it with:
+
+    sudo ./install-latest.sh
+
 ### Bleeding edge development
 
 Sous is written in Go.
--- a/install-latest.sh
+++ b/install-latest.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# This script is not intended to be bomb-proof. It is designed to optimistically
+# attempt to install the latest released version of Sous. If any of its assumptions
+# about how GitHub works or the structure of the tarball become false, it will fail.
+
+REPO_URL="https://github.com/opentable/sous"
+PREFIX=/usr/local/sous
+BIN=/usr/local/bin
+
+[ -d "$PREFIX" ] || mkdir -p "$PREFIX"
+
+if [ "$(uname)" == Darwin ]; then
+	PLATFORM=darwin-amd64
+elif [ "$(uname)" == Linux ]; then
+	PLATFORM=linux-amd64
+else
+	echo "Unix name $(uname) not recognised."
+	echo "Sous can be installed on Linux or Darwin (macOS)"
+fi
+
+# Get the git tag of the latest release by examining the HTTP redirect.
+TAG=$(curl -si $REPO_URL/releases/latest | grep '^Location: ' | cut -f 8 -d /)
+TAG="${TAG%?}"
+
+if [ "${TAG:0:1}" != "v" ]; then
+	echo "Expected tag to begin with 'v'."
+fi
+
+# VERSION is TAG without first character.
+VERSION="${TAG#?}"
+
+echo "Installing sous version $VERSION"
+
+# Calculate assumed filename and tarball URL.
+FILENAME="sous-${PLATFORM}_$VERSION"
+TARBALL="$FILENAME.tar.gz"
+TARBALL_URL="$REPO_URL/releases/download/$TAG/$TARBALL"
+
+echo "Downloading sous from $TARBALL_URL"
+curl -LO "$TARBALL_URL"
+mv "$TARBALL" "$PREFIX/"
+
+(
+	cd "$PREFIX"
+	echo "Extracting to $PREFIX..."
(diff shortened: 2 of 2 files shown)


alrs:
Once someone knows enough to `tar zxvf sous-whatever.tar.gz` one would hope they know enough to do something with the binary, either running it in place or moving it somewhere in the PATH. I'm somewhat allergic to supporting scripts like this, especially scripts that require root and start creating symlinks off in /usr.

nyarly:
Point of fact, there are devs for which this is a challenge, even the fetch and tar steps. They'll run stuff if it's in docs in front of them, but not everyone lives in the command line comfortably enough to know how to manipulate their PATH etc.

That said, I'd think it makes more sense to copy or move the sous executable rather than symlink it.

alrs:
If a script is needed to get the sous binary into the path, I suggest putting it in the release tarball instead of in the root of the project. I would prefer to avoid anything that has root, visits the Internet, and starts messing around in /usr.

Would the existing `brew` work need much updating to deal with this simple-installation use-case?

samsalisbury:
I agree that this script is pretty horrible, requiring sudo etc. It was intended to be the cheapest possible thing that stops each introduction of Sous being preceded by manual installation steps, which as Judson notes not everyone is immediately comfortable with. I think updating the homebrew formula is probably a better way to go about this given more time, but it's an additional maintenance cost, as we would need to synchronise the formula with each release of Sous (in order for homebrew to detect when an upgrade is available). Or, better, we would need to spend time automating the formula update on each Sous release. Either way, when I introduce Sous to a new person, or when I want someone to upgrade to latest, I want to avoid a repetitive procedure each time which detracts from the experience of using Sous. I find it interesting to see whether or not people balk at this script (so far no one has, although personally I would not just sudo some random script, I'd do the install manually).

Re copying the binary vs symlink, I have no strong opinion, was just going by what I thought was customary for this kind of thing.

Re where should such a script live, I wanted it to be in the main Sous repo because that's the first place people will likely look. For those who just want to trustingly install the latest release and start playing, something like this seems helpful. Until we have time to invest in the homebrew formula, I'll keep using this script from my branch, but adding a ticket to add homebrew formula updates to the deployment workflow.

samsalisbury:
This script as-is works on both Linux and Darwin. If we go down the Homebrew route we may still need a Linux installer/package of some kind if there is any demand for it.

alrs:
I don't think this sinks to the level of horrible. I'm sympathetic to what you're doing. My concerns, iterated and for the record, are:

1. Reinventing package management for each package on a system drives sysadmins nuts, especially when it's done in bash, especially when it involves unsigned artifacts. Homebrew can be potentially abstracted with https://github.com/jordansissel/fpm

2. Now that sous is prod-ready, we should accept that conservative users are likely to stay on the same version for ~18 months at a time. 

3. A half-measure that could keep us out of the DIY package management game would be to provide a "sous plumbing latest" command that would query Github and return the latest available release, maybe only the latest release with the same semver major/minor versions.

4. Should we be publishing an MD5 hash of the release tarball?

5. If perfect is the enemy of the good, and we wind up with a shell script: Does OS X come with a ~/bin directory in the $PATH by default?

samsalisbury:
OK, I think these concerns are totally valid. This was always intended as a stop-gap for quicker onboarding, until we decide to take the time to do it in a more proper way. Since I can see that way is now being worked on to some degree, I won't insist on merging this, but will continue to use this script from my fork until we have something better.

Re individual points...

> Reinventing package management for each package on a system drives sysadmins nuts, especially when it's done in bash, especially when it involves unsigned artifacts. Homebrew can be potentially abstracted with https://github.com/jordansissel/fpm

I hadn't come across FPM before, that should take a lot of the pain out of doingitrite :)

> Now that sous is prod-ready, we should accept that conservative users are likely to stay on the same version for ~18 months at a time.

I think we should force users to upgrade by having the server refuse to talk to old clients, even if they would sort of work. Sous is still not stable, and having to preserve backwards-compatibility in the server for old clients will hobble rapid development. IMO for now we should be strict any only allow exact major/minor/patch matches. Later we could loosen it if users complain, but a quick and easy upgrade route should reduce the likelihood of that.

We should definitely add a client option for disabling the server version check, and emit loud warnings if it's used.

> A half-measure that could keep us out of the DIY package management game would be to provide a "sous plumbing latest" command that would query Github and return the latest available release, maybe only the latest release with the same semver major/minor versions.

Maybe, although the more I think about it, the more I'm convinced the server should dictate client version, rather than simply the latest release. That would ensure we have time to roll out the new server before everyone upgrades to a client that's too new.

> Should we be publishing an MD5 hash of the release tarball?

Absolutely, this is required for Homebrew as well.

> If perfect is the enemy of the good, and we wind up with a shell script: Does OS X come with a ~/bin directory in the $PATH by default?

Doesn't look like it...  (see 'man path_helper' on OS X)

```shell
$ cat /etc/paths
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin
```

samsalisbury:
Closing this in favour of https://github.com/opentable/homebrew-osx-tools/pull/1

PR #305, opened 2017-01-24
Sous deploy -repo tag

nyarly:
I couldn't reproduce the issue described in a test - if you note the tests
added to cli_test.go, you'll see that testing with and without -repo produces
the expected behavior.

Along the way I noted that `sous update` (which should perhaps properly be
moved to `plumbing`) will automatically perform a manifest creation, in a way
that is similar to, but separate from `sous init`. Given that manifests cannot
be deleted (at the moment, it's a git operation that requires rebooting all the
running sous servers), creating them automatically seems like a problem to me.
I've removed that option, with a warning to `sous init.`

Finally, I started to see how the StatusPoller doesn't quite align with other
tools that read from the git context. This bears more discussion, but part of
the upshot is that the poller (and therefore `deploy`) does respect the
workspace git context, then the interface to e.g. wait for a completely stable
cluster becomes harder.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -64,10 +64,20 @@ func SuccessYAML(v interface{}) cmdr.Result {
 // Plumbing injects a command with the current psyringe,
 // then it Executes it, returning the result.
 func (cli *CLI) Plumbing(cmd cmdr.Executor, args []string) cmdr.Result {
-	if err := cli.SousGraph.Inject(cmd); err != nil {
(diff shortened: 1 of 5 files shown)


samsalisbury:
Thanks for adding the test, apologies for the erroneous bug report, it seems some TeamCity config didn't get updated when I thought it had been.

samsalisbury:
Re 'sous update' being a plumbing command: sounds reasonable, and also makes sense to remove that 'sous init' mimic from there.

PR #306, opened 2017-01-24
Version in deployID

alrs:
Adding version information to the Singularity deployID.

### DIFF
--- a/ext/singularity/deployer.go
+++ b/ext/singularity/deployer.go
@@ -100,7 +100,7 @@ func (r *deployer) RectifySingleCreate(d *sous.Deployable) (err error) {
 		return err
 	}
 	return r.Client.Deploy(
-		d.Cluster.BaseURL, newDepID(), reqID, name, d.Resources,
+		d.Cluster.BaseURL, computeDeployID(d), reqID, name, d.Resources,
 		d.Env, d.DeployConfig.Volumes)
 }
 
@@ -172,7 +172,7 @@ func (r *deployer) RectifySingleModification(pair *sous.DeployablePair) (err err
 
 		if err := r.Client.Deploy(
 			pair.Post.Cluster.BaseURL,
-			newDepID(),
+			computeDeployID(pair.Post),
 			computeRequestID(pair.Prior),
 			name,
 			pair.Post.Resources,
@@ -250,8 +250,4 @@ func ParseRequestID(id string) (sous.DeployID, error) {
 		},
 		Cluster: parts[2],
 	}, nil
-}
-
-func newDepID() string {
-	return MakeDeployID(uuid.NewV4().String())
 }
--- a/ext/singularity/deployer.go
+++ b/ext/singularity/deployer.go
@@ -203,6 +203,18 @@ func computeRequestID(d *sous.Deployable) string {
 	return MakeRequestID(d.ID())
 }
 
+func computeDeployID(d *sous.Deployable) string {
+	flavor := illegalDeployIDChars.ReplaceAllString(d.Deployment.Flavor, "")
+	if len(flavor) == 0 {
+		flavor = "SINSABOR"
+	}
+	return strings.Join([]string{
+		illegalDeployIDChars.ReplaceAllString(d.Deployment.SourceID.Version.String(), ""),
+		flavor,
+		illegalDeployIDChars.ReplaceAllString(uuid.NewV4().String(), ""),
+	}, "-")
+}
+
 // MakeRequestID creats a Singularity request ID from a sous.DeployID.
 func MakeRequestID(mid sous.DeployID) string {
 	sl := strings.Replace(mid.ManifestID.Source.String(), "/", ">", -1)
--- a/ext/singularity/deployer.go
+++ b/ext/singularity/deployer.go
@@ -206,7 +206,7 @@ func computeRequestID(d *sous.Deployable) string {
 func computeDeployID(d *sous.Deployable) string {
 	flavor := illegalDeployIDChars.ReplaceAllString(d.Deployment.Flavor, "")
 	if len(flavor) == 0 {
-		flavor = "SINSABOR"
+		flavor = "NOFLAVOR"
 	}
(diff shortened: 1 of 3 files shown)


ext/singularity/deployer.go
```diff
@@ -203,6 +203,18 @@ func computeRequestID(d *sous.Deployable) string {
 	return MakeRequestID(d.ID())
 }
 
+func computeDeployID(d *sous.Deployable) string {
+	flavor := illegalDeployIDChars.ReplaceAllString(d.Deployment.Flavor, "")
+	if len(flavor) == 0 {
+		flavor = "NOFLAVOR"
```
samsalisbury: I couldn't help myself and implemented this, adding a test case for empty flavor. See https://github.com/samsalisbury/sous/commit/3f25c05bf6dfe0a96872e6a3cdd55d4d551be8b9 not opening a PR, but feel free to cherry-pick that if you want.

ext/singularity/deployer.go
```diff
@@ -203,6 +203,18 @@ func computeRequestID(d *sous.Deployable) string {
 	return MakeRequestID(d.ID())
 }
 
+func computeDeployID(d *sous.Deployable) string {
+	flavor := illegalDeployIDChars.ReplaceAllString(d.Deployment.Flavor, "")
```
nyarly: I have bad news: I just tried this on a live singularity, and discovered this: 

```
Failed to redeploy df1d7456ac654e4097ae451b0dc6c0fa. The server responded with a HTTP 400 and said:
Id must not be null and can not contain any of the following characters: [@, -, \, /, *, ?, %, , [, ], #, $]
```

illegalDeployIDChars is only `[-/:]`

ext/singularity/deployer.go
nyarly: The UUID already (nearly) guarantees that the DeployID will be unique, so dropping the flavor component is likely safe, and I tend to agree that since the goal here is to provide human-readable information, the less noise the better.

ext/singularity/deployer.go
```diff
@@ -203,6 +203,18 @@ func computeRequestID(d *sous.Deployable) string {
...
+	if len(flavor) == 0 {
+		flavor = "NOFLAVOR"
+	}
+	return strings.Join([]string{
+		illegalDeployIDChars.ReplaceAllString(d.Deployment.SourceID.Version.String(), ""),
+		flavor,
+		illegalDeployIDChars.ReplaceAllString(uuid.NewV4().String(), ""),
+	}, "-")
```
nyarly: Yeah, sadly, "-" is in the characters that Singularity theoretically will reject in a deployID.

Note that we should probably also include whatever delimiter we do decide on in the ReplaceAll, but that the UUID makes it safe not to.

ext/singularity/deployer.go
nyarly: AND (hurray :|) `Deploy id must be less than 50 characters`, which leaves us with 18 characters after the UUID. ~8 consumed by a no-meta semver (xx.yy.zz), leaves 8 for any flavor at all (assuming to delimiter characters.)

nyarly: Sorry: 7. Singularity really does mean "less than 50"

nyarly: These characters are allowed: `=()_+{}`

Note that I found that other symbols not in the explicit exclusion list would still fail - the Executor issued and error when it tries to start the task. I didn't do an exhaustive search to determine which characters in particular it was. I would guess something in `!&'"` caused problems on the shell? Or in an exec?

nyarly: If anyone cares:

```
Failed to launch container: Failed to create container: Failed to chown: Failed to execute 'chown -R 0:0 '/mnt/mesos-slave/slaves/f38ff857-4831-46dc-b143-d8aaa0c44d06-S0/frameworks/Singularity/executors/github.com>opentable>sous-demo::ci-sf-<>|=()'_+{}"^&df1d7456ac654e4097ae451b0dc6c0fa-1485370286019-1-mesos_slave12_qa_sf.qasql.opentable.com-FIXME/runs/9967afcc-0feb-4421-baa7-094d6d8896fb'' (exit status: 512): Success; Abnormal executor termination
```

samsalisbury: Thanks for digging into this. I think it's reasonable to truncate version strings at, say, 18 characters, and warn the user that this has happened. For the vast majority, 18 characters will be plenty to show the whole version, and for the others they can just live with it, since uniqueness will come from the uid any way.

ext/singularity/deployer.go
samsalisbury: I concur.

nyarly:
Just to bring the conversation about Deploy ids home: based on investigation with Singularity itself, discussion with the Singularity team, and an extended private muttering session, let me suggest:

Stringify the M.m.p version (i.e. no meta part), generate a UUID. Put them in a slice.

If the flavor isn't empty, truncate it to, say, 12 characters, and prefix it on the slice.

Join the 2 or 3 element slice with `_`.

Truncate the whole thing to 49 characters.

There's the one pathological case where the version is so long that it runs out meaningful bytes of the UUID, but that seems unlikely, not tragic, and easy enough to fix in the breach.

samsalisbury:
I think leaving out the [prerelease field](http://semver.org/#spec-item-9) of the version is too much, since people who use that field will likely really care about its contents. So, we should use 'M.m.p-prerelease' but I concur with leaving out the [metadata field](http://semver.org/#spec-item-10) proper.

Having thought about `flavor` more, I now realise that we probably shouldn't have this as part of the version string at all, since it's already part of the request name, which is visible whenever you can see the deploy ID. Dropping that duplicate info also gives us more space for everything else.

So, I would advocate truncating the entire version string to (49-17=32) characters, and appending as many of the 32 char UUID after that as possible (guaranteed minimum 16 chars), separated by `_` as you suggest. People with version strings longer than 32 characters would not see them in their entirety, but I don't think this is a very likely occurrence.

PR #307, opened 2017-01-25
Does this work?

alrs:
Tests are passing locally, which is surprising.


alrs:
Dunno if this works, I forked from a broken master.

PR #308, opened 2017-01-25
Undoing nil Manifest from newManifest

nyarly:
Consequent to changes in the way otpl flags are processed,
newTargetManifest could (effectively) return a nil Manifest. None of its
consumers expect this behavior, so I'm reverting it, ensuring that an
empty Manifest (with default DeploySpecs) is always returned.

### DIFF
--- a/cli/cli_test.go
+++ b/cli/cli_test.go
@@ -100,6 +100,8 @@ func TestInvokeDeploy_RepoFlag(t *testing.T) {
 	sd.CLI.Plumb(su, sps)
 
 	assert.Equal(su.ResolveFilter.Repo, "")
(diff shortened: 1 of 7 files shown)


samsalisbury:
This seems reasonable, just curious what led to this change/what does it solve outwardly in Sous?

nyarly:
There was a test that had been added (ironically, the couldn't-reproduce case that I added for the "-repo is ignored" bug) that was failing. The alternative was to add guards for the nil case in update (which has the opposite guard (I.e. for an empty, missing manifest) already), and to greatly expand or remove the test case from ci_test.

samsalisbury:
Ah, I see :)

PR #310, opened 2017-01-26
Remove sudo from test scripts

alrs:
This is an ongoing experiment to remove sudo from the test scripts.


dev_support/sous_qa_setup/certificates.go
```diff
@@ -19,6 +19,15 @@ import (
 //go:generate inlinefiles --vfs=Templates --package main templates vfs_template.go
 
 func registryCerts(testAgent test_with_docker.Agent, composeDir string, desc desc.EnvDesc) error {
+	// Setting up registryCerts has been shown not to matter when running against a Docker
```
alrs: I reckon this to be the most controversial part.

PR #313, opened 2017-01-31
De-Sudo Tests

alrs:
The integration tests previously used the sudo command and hacked up the local Docker config, which made them unsuitable for running locally. I've removed the use of sudo, put the test running into the Makefile, and put the development container key generation into the Makefile as well.

I disabled key generation in sous_qa_setup by bailing out when encountering LocalDaemon structs. This doesn't feel as pure as the wind-driven snow. If we're lucky this approach will work with Docker machine as well, and we can ditch all of the keygen/docker certs.d stuff.


integration/test-registry/ssl.conf
```diff
@@ -0,0 +1,12 @@
+[ req ]
+prompt = no
+distinguished_name=req_distinguished_name
+x509_extensions = va_c3
+encrypt_key = no
+default_keyfile=testing.key
+default_md = sha256
+
+[ va_c3 ]
+basicConstraints=critical,CA:true,pathlen:1
```
nyarly: IIRC, this file was templated before? One of the key details was that there was a SAN for the IP of the registry, so having it be a static file seems like it's completely superfluous?

alrs: The templated version is still there. This version is for making certs from the Makefile for a Docker LocalDaemon and not needing to compile any Go or template anything. I want to generate certs before I've run sous_qa_setup, so I need either a conf file or to figure out all of the command line switches to do this entirely from the Makefile.

PR #316, opened 2017-02-01
Add User config

samsalisbury:
UPATE: Ready to merge.

This PR adds a new section in the config.yaml file:

```yaml
User:
  Email: some@email.com
  Name: Some Name
```

And sends that user to the server with each request. It also sets the git commit author information to whichever user issued the instruction to update the GDM, via the git commit flag `--author "Some Name <some@email.com>`


samsalisbury:
Please see the draft release notes in this PR for full details of user-facing changes.

lib/httpclient.go
```diff
@@ -72,18 +72,18 @@ func NewClient(serverURL string) (*LiveHTTPClient, error) {
 // style query params. It deserializes the returned JSON into rzBody. Errors
 // are returned if anything goes wrong, including a non-Success HTTP result
 // (but note that there may be a response anyway.
-func (client *LiveHTTPClient) Retrieve(urlPath string, qParms map[string]string, rzBody interface{}) error {
+func (client *LiveHTTPClient) Retrieve(urlPath string, qParms map[string]string, rzBody interface{}, user User) error {
 	return errors.Wrapf(func() error {
 		url, err := client.buildURL(urlPath, qParms)
-		rq, err := client.buildRequest("GET", url, nil, nil, err)
+		rq, err := client.buildRequest("GET", url, map[string]string{"Sous-User": user.String()}, nil, err)
```
nyarly: Two things:

First, let's make the user an arg to buildRequest, and let buildRequest set the headers.

Second, let's make two headers: "Sous-Username" and "Sous-Email" rather than interpolate/parse.

Eventually, we plan to use this data for authentication, at which point it becomes really important to be positive we have the correct information.

server/user.go
```diff
@@ -0,0 +1,26 @@
...
+type ClientUser sous.User
+
+// parseHeaders parses a Headers from the headers of a HTTP request.
+func getUser(req *http.Request) ClientUser {
+	var user ClientUser
+	userString := req.Header.Get("Sous-User")
+	if userString == "" {
+		user.Name = "Anonymous"
+	}
+	parts := strings.SplitN(userString, "<", 2)
```
nyarly: The code here becomes much easier, and much more reliable if these are passed as two headers.

lib/user.go
```diff
@@ -0,0 +1,15 @@
+package sous
+
+import "fmt"
+
+// User represents a user of the Sous client.
+type User struct {
+	// Name is the full name of this user.
+	Name,
```
nyarly: While I'm at it, an `env` tag for Name and Email would be all that a build agent would need to e.g. use the author of the triggering commit to provide a user.

lib/httpclient.go
nyarly: Two things:

First, let's make the user an arg to buildRequest, and let buildRequest set the headers. Among other things, the user should be transmitted with every request, not just Retrieves.

Second, let's make two headers: "Sous-Username" and "Sous-Email" rather than interpolate/parse.

Eventually, we plan to use this data for authentication, at which point it becomes really important to be positive we have the correct information.

server/user.go
samsalisbury: Yeah, I agree, changes on the way...

lib/user.go
samsalisbury: 👍

lib/httpclient.go
samsalisbury: 👍

PR #317, opened 2017-02-01
Made -v depend on VERBOSE in make test

nyarly:
You can `make test VERBOSE=1` if you want verbose output. Or you can
export VERBOSE=1. travis config updated to do the latter.


alrs:
Tests don't touch the Makefile, and this lgtm.

PR #320, opened 2017-02-02
Version DeployID

alrs:
Back to https://github.com/opentable/sous/pull/306, minus the de-sudo work that was moved into its own PR.


ext/singularity/deployer.go
```diff
@@ -203,6 +210,35 @@ func computeRequestID(d *sous.Deployable) string {
...
+		versionTrunc = versionEntire
+	}
+
+	// naiveLen is the length of the truncated Version plus
+	// the length of an entire UUID plus the length of a separator
+	// character.
+	naiveLen := len(versionTrunc) + len(uuidEntire) + 1
+
+	if naiveLen > maxDeployIDLen {
+		uuidTrunc = uuidEntire[0:(maxDeployIDLen - naiveLen - 1)]
```
nyarly: Actually, I think this is wrong. If `naiveLen > maxDeployIDLen`, then `maxDeployIDLen - naiveLen <= 0`.

ext/singularity/deployer.go
```diff
@@ -203,6 +210,35 @@ func computeRequestID(d *sous.Deployable) string {
 	return MakeRequestID(d.ID())
 }
 
+func computeDeployID(d *sous.Deployable) string {
+	var uuidTrunc, versionTrunc string
+	uuidEntire := StripDeployID(uuid.NewV4().String())
+	versionSansMeta := stripMetadata(d.Deployment.SourceID.Version.String())
```
nyarly: If this PR needs update, semv.Version has a bunch of facility for producing strings e.g. without metadata in the first place.

ext/singularity/deployer.go
nyarly: IOW, is there a test for computeDeployID where the version is long enough to trip over into this behavior?

PR #322, opened 2017-02-03
(WIP) Consider stable only deployments whose tasks have started.

samsalisbury:
Work in progress.


lib/status_poller.go
```diff
@@ -38,7 +38,7 @@ type (
 
 	// copied from server - avoiding coupling to server implemention
 	statusData struct {
-		Deployments           []*Deployment
+		Deployments           []*DeployState
```
nyarly: n.b. that Deployments from the server are "the GDM when I started resolution" - from the poller's perspective, it's used to compute whether the server is working on our deployment in the first place (i.e. "is the current state ResolveNotVersion").

I think the most straightforward approach would be to put the active/pending distinction into the ResolveStatus logs?

PR #323, opened 2017-02-04
Wait for scheduler stabilization

nyarly:
This is a WIP.

The logic here is that the DiffChans have been changed from a mix of Deployment
chans and DeploymentPair chans (the latter only used for modified Deployments)
to all being `chan *DeploymentPair`.  This should ultimately allow for a more
consistent interface across the board. In the meantime, it allows us to add a
Status field to DeploymentPair, which we can then inspect as we prepare the
DiffResolution log entries in `lib/resolver.go`.

I've added constants for the DiffResolution "Desc" field, (the distinction
between e.g. "created" and "not created" being elided into the presence or
absence of an error) and added a fifth value for this field: "coming" - the
idea is to set that in `reportStable` when the DeploymentPair has the
appropriate Status (i.e. because there's a Pending Deploy on the Request in
Singularity.)

Ultimately, the StatusPoller should treat the "coming" status as
ResolveTasksPending (or w/e).

The throughline is pretty clear here, I feel. The next step would be write some
tests for the existing changes, and then TDD out the rest of the threading here.


nyarly:
@samsalisbury if you want to pick up from here, cool.

Also pleased because this approach gives a better insight into handling Deploy failures, I think - where the Retained pairs currently will have two Statuses (stable, coming), failures will imply three Statuses (I think): fresh (a clean update to a new version or w/e), blocked (because there's an existing Pending deploy (Singularity replies 419 until the Request stabilizes), or "retry" (because the top of the Deploy History is a DeployFailed (ideally matched against the Deployment we're attempting, but I think that'll require setting Deploy Metadata to be able to correctly track that.)) Then a "retry" modified DiffResolution implies a failure (I think? Maybe the matching up is required to know that for sure.)

nyarly:
I'm at something of a loss how to pull coverage up much further without sinking a ton of effort into it. I think the salient parts of the change are covered: computing the DeployStatus based on the Singularity data, and then computing the overall ResolveStatus based on the DeployStatus.

PR #324, opened 2017-02-06
Tiny change for Makefile compatibility

nyarly:
Removed -r from xargs

'cuz OS X is kinda BSD

The - at the start of a Makefile recipe line means "no worried if this fails"


samsalisbury:
Tx for explanation of the `-`

PR #327, opened 2017-02-14
Epic CLI integration test work

nyarly:
Epic only in the size and number of commits.

@samsalisbury I'm afraid this is going to chew up your day. I'll try to do some
guidance on the PR about what's going on.

The key thing here is that this a shell-only integration test of Sous, starting
from *almost* nothing, installing sous, setting up a sous-server, doing a
workstation deploy to bootstrap the clusters, creating a demo service, init
build deploy that, and confirming that it gets deployed.


integration/test-homedir/bin/ssh_wrapper
```diff
@@ -0,0 +1,3 @@
+#! /bin/sh
+
+{{.SSHExec}} -o PasswordAuthentication=no -F "{{.Homedir}}/dot-ssh/config" "${@}"
```
nyarly: We need this for git because SSH *always* uses /etc/passwd (or the equivalent) rather than $HOME. Since our home directory is a tempdir, we need to use a wrapper to get the appropriate ssh configs.

integration/test-registry/clean-singularity.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
nyarly: We use this to make sure all the requests from previous test runs have been cleaned out.

integration/test-registry/docker-compose.yml
```diff
@@ -43,22 +49,23 @@ slave:
     - /sys:/sys
 
 scheduler:
-  image: hubspot/singularityservice:0.4.11
+  #image: hubspot/singularityservice:0.13.0
+  build: singularityservice
```
nyarly: This turns out to be important, and merits documentation. Sous uses the bridge networking configuration and *relies* on an option added by Hubspot for us.

We may want to consider configuration our requests so that this isn't required.

integration/test-registry/singularityservice/singularity.yaml
```diff
@@ -0,0 +1,43 @@
...
+  baseUrl: http://localhost:7099/singularity
+
+logging:
+  loggers:
+    "com.hubspot.singularity" : INFO
+
+enableCorsFilter: true
+
+network:
+  defaultPortMapping: true
```
nyarly: This is the crucial configuration, and why we overlay the default configuration setup.

lib/user.go
```diff
@@ -16,5 +16,18 @@ type User struct {
 //
 // This is used directly by git commit --author, and is reasonable in logs etc.
 func (u User) String() string {
-	return fmt.Sprintf("%s <%s>", u.Name, u.Email)
+	parts := []string{}
```
nyarly: This just makes sure that if parts of the User are unset we don't wind up Stringing to ` <>`

server/graph.go
```diff
@@ -13,8 +14,9 @@ type (
 )
 
 // AddsPerRequest registers items into a SousGraph that need to be fresh per request
-func AddsPerRequest(g Injector) {
+func AddsPerRequest(g restful.Injector) {
```
nyarly: Debugging the update of servers, I refactored out what was properly the server library code into `restful`, leaving the Sous server in place. Hopefully this makes the code easier to follow.

server/handle_servers.go
```diff
@@ -22,13 +36,33 @@ type (
...
 func (slh *ServerListHandler) Exchange() (interface{}, int) {
 	data := serverListData{Servers: []server{}}
 	for name, url := range slh.Config.SiblingURLs {
 		data.Servers = append(data.Servers, server{ClusterName: name, URL: url})
 	}
 	return data, 200
 }
+
+// Exchange implements restful.Exchanger on ServerListUpdater
+func (slh *ServerListUpdater) Exchange() (interface{}, int) {
```
nyarly: Because of this, it's now possible to update the list of sibling servers. There's not really any client support for this yet, or a gossip or whatever, but at least running servers can be given the most up to date list of clusters without being rebooted.

nyarly: Needed this in testing, since couldn't be sure what the IP:port pair that Singularity would give us, so had to come back after the servers were running to update the list.

util/configloader/configloader.go
```diff
@@ -86,14 +86,34 @@ func (cl *configLoader) overrideWithEnv(target interface{}) error {
...
-	if v.Kind() != reflect.Ptr && v.Elem().Kind() != reflect.Struct {
+	if v.Kind() != reflect.Ptr || v.Elem().Kind() != reflect.Struct {
 		return fmt.Errorf("target was %T; need a pointer to struct", target)
 	}
 	v = v.Elem()
 	t := v.Type()
 	for i := 0; i < v.NumField(); i++ {
-		if err := f(t.Field(i), v.Field(i)); err != nil {
-			return err
+		field := t.Field(i)
```
nyarly: Needed nested config structs to get populated from ENV as well, which is what this is about.

util/docker_registry/client.go
```diff
@@ -98,19 +100,36 @@ func (rs *Registries) DeleteRegistry(n string) error {
 
 // NewClient builds a new client
 func NewClient() Client {
+	xport := &http.Transport{}
+	if extraCA := os.Getenv("SOUS_EXTRA_DOCKER_CA"); extraCA != "" {
```
nyarly: Again, this was needed for testing: the running servers need a way to be told to trust our test docker registry. Giant pain, but here it is solved once and for all. We can probably do away with `BecomeFoolishlyTrusting` because of this as well.

util/shelltest/captiveshell.go
```diff
@@ -0,0 +1,241 @@
...
+		pipe: from,
+		bufs: []*bytes.Buffer{&bytes.Buffer{}},
+	}
+
+	go ls.reader(events)
+	return ls
+}
+
+// NewShell creates a new CaptiveShell with an environment dictacted by env.
+func NewShell(env map[string]string) (sh *CaptiveShell, err error) {
```
nyarly: Here's the "get a single bash, feed it commands, see what it does, then feed it more commands" thing. Important so that e.g. environment and filesystem remain between script blocks.

util/shelltest/shelltest.go
```diff
@@ -0,0 +1,133 @@
...
+		tmplContext    interface{}
+	}
+
+	// A CheckFn receives the Result of running a script, and should inspect it,
+	// calling methods on the testing.T as appropriate
+	CheckFn func(string, Result, *testing.T)
+)
+
+// New creates a new ShellTest
+func New(t *testing.T, name string, ctx interface{}, env map[string]string) *ShellTest {
```
nyarly: The ShellTest then, drives a captive shell, inspects output, writes files etc.

ext/docker/names.go
```diff
@@ -56,6 +55,8 @@ func Labels(sid sous.SourceID) map[string]string {
 	return labels
 }
 
+var stripRE = regexp.MustCompile("^([[:alpha:]]+://)?(github.com(/opentable)?)?")
```
samsalisbury: We should probably not be stripping out anything opentable-specific here, nor github-specific. Ideally, the name of the docker repository should be a 1-1 mapping from SourceLocation, and impose no more opinions than that. Sous needs to be rejecting up-front any SourceLocations that contain unusual characters like colons, semicolons, plus signs, tildes, etc for now.

ext/storage/git_state_manager.go
```diff
@@ -99,8 +103,13 @@ func (gsm *GitStateManager) WriteState(s *sous.State, u sous.User) error {
 	if !gsm.needCommit() {
 		return nil
 	}
-	author := u.String()
-	if err := gsm.git("commit", "-m", "sous commit: Update State", "--author", author); err != nil {
+	commitCommand := []string{"commit", "-m", "sous commit: Update State"}
+	if u.Complete() {
```
samsalisbury: I wonder if we are better off changing the implementation of User.String so that it always returns a valid string, e.g. "No Author Set In Config <nobody@nowhere.com>" that way we don't silently add commits with no author info, which also means we don't need to rely on global git configuration for this info (which this code implies we are using otherwise). Alternatively, I would be happy for Sous to refuse to operate without user name + email being set.

graph/graph.go
```diff
@@ -514,6 +514,12 @@ func NewCurrentState(sr StateReader) (*sous.State, error) {
 
 // NewCurrentGDM returns the current GDM.
 func NewCurrentGDM(state *sous.State) (CurrentGDM, error) {
+	if state == nil {
+		// XXX Sometimes, regardless of an error returned by NewCurrentState, this
+		// function is still called with a nil State, resulting in a panic. Race
+		// condition in psyringe?
```
samsalisbury: Hmm, I guess we'll need to look into this more. The Psyringe tests are always run with `-race` and fail on any detected race condition, although the race detector is not perfect. The other possibility is that somehow we are generating a nil state as well as a nil error.

doc/shellexamples/001_Preconditions for CLI integration tests.blended
```diff
@@ -0,0 +1,12 @@
+
+bash-3.2$ go get github.com/nyarly/cygnus # cygnus lets us inspect Singularit</nyarly/cygnus # cygnus lets us inspect Singularity                          for ports
```
samsalisbury: These `^M` (aka `\r`) carriage returns look odd on my machine, and on github by the looks of it. I wonder if the output could be slightly sanitised to use `\n` or `\r\n` or something? Unless that would break assertions?

doc/shellexamples/001_Preconditions for CLI integration tests.blended
```diff
@@ -0,0 +1,12 @@
...
+bash-3.2$ if echo "http://192.168.99.100:7099/singularity" | egrep -q '192.16<92.168.99.100:7099/singularity" | egrep -q '192.168                         |127.0.0'; then
+> /Users/jlester/golang/src/github.com/opentable/sous/integration/test-regist</github.com/opentable/sous/integration/test-registr                         y/clean-singularity.sh ht</integration/test-registry/clean-singularity.sh htt                         p://192.168.99.100:7099/s</clean-singularity.sh http://192.168.99.100:7099/si                         ngularity
```
samsalisbury: I wonder if we should sanitise test output so that usernames don't appear anywhere, just feels messy/noisy & is giving away some information that we don't need to.

graph/local_config.go
```diff
@@ -115,7 +118,12 @@ func userInput(prompt, vDefault, eg string, v *string) {
 
 func userInitConfig(c *config.Config) {
 	if !terminal.IsTerminal(int(os.Stdout.Fd())) {
-		sous.Log.Warn.Println("Unable to run config wizard; no terminal attached.")
+		sous.Log.Warn.Println("Unable to run interactive configuration; no terminal attached.")
+		return
+	}
+	if os.Getenv("TASK_HOST") != "" { // XXX This is terrible, but the terminal check fails (which breaks the Mesos servers)
```
samsalisbury: Yeah, it gets us going for now though. Let's check both stdout and stdin are terminals, any maybe try to think of some less Singularity/OpenTable specific heuristics. Not blocking over this though.

doc/shellexamples/001_Preconditions for CLI integration tests.blended
nyarly: There aren't any assertions that it would break, but it'd be an extra task - I can see if I can get to it today.

Just to clarify, since this is the output of the shell, what we're seeing is the \n that Bash emits at the end of a line so that your input will wrap while you're entering it. It should be possible to do a gsub on the output before it's saved, though.

ext/docker/names.go
nyarly: Outright rejection isn't an option for these tests to work. Sanitization is an option, though, and the code already exists.

I do think we're even more likely to see a revolt if we start adding docker images named "github.com/opentable/project", but I also see the need to establish some kind of namespace with respect to avoiding clobbering existing docker repos.

Sanitization has its own name collision problems, though. The simplest solution would be to require configuring an explicit docker repo name in the manifest. The convention of an org wide set of git remote -> docker repo name mappings would ameliorate the worst of the configuration burden there.

ext/storage/git_state_manager.go
nyarly: Automatic agents (e.g. TeamCity) are the biggest problem with this. Of course, they'd need to do something to pull that data from e.g. a git commit and then provide it to Sous, so maybe that's the right approach.

Still, if we're going to require it be configured, the error should be raised much earlier. (which adds the wrinkle of not requiring it for commands that don't require it...)

graph/graph.go
nyarly: I learned recently that the limit of the race detector is that it only detects *global* races - local contentions that still allow the main program to proceed for instance aren't caught.

To clarify, I confirmed that there was an error generated in a precursor state, but that this provider was still called with the nil state.

doc/shellexamples/001_Preconditions for CLI integration tests.blended
nyarly: Captured this concern in a story.

nyarly:
Re isolating the tests - I generally agree with the concern. Setting up a Vagrantfile, though, loses us running these tests on a Mac, which I'm optimistic as a means to address user issues... I wonder if we could do something like a chroot to contain the tests? That might even be an appropriate extension of the shelltest tool - setup a tempdir with templated files, and then chroot into it?

graph/graph.go
samsalisbury: Hmm, I did not know that about the race detector, where did you read that? Typically programs can still continue in spite of a data race. My understanding was that with [100% test coverage](https://codecov.io/gh/samsalisbury/psyringe) and running [tests with the -race flag](https://circleci.com/gh/samsalisbury/psyringe), we should be able to be reasonably confident that there aren't any data races...

samsalisbury: Maybe it's time to turn on the race detector in our QA deployments, since it should be better at picking up this kind of thing in real operation than solely in tests... Added a story to the private JIRA board: 

### Enable race detector; enumerate data races as JIRA tickets.

> Some of our tests seem to indicate intermittent problems that smell like data races. This isn't good for the confidence of Sous' customers.
>
> First thing to do is to start running tests in CI with the -race flag, and generate a bug report for each race condition. Once they are fixed, start failing the build over any new race conditions that are introduced. Then let's turn on race detection in our QA deployments and add further bug tickets for any that are detected there.

ext/storage/git_state_manager.go
samsalisbury: The trouble with CI contexts is that they may not have a global author set. In that case, Git will refuse the commit saying there's no author information. So, this code currently assumes that the CI agent has git --global author information set up already, otherwise it will fail to commit. If we just fudge a made up author for now, when there is no author info set up in the config.yaml, we avoid failed commits, and can revisit it later for a better solution I think...

samsalisbury:
Yeah, running on Mac is valuable as well... Chroot sounds like an interesting idea.

This is still failing on my machine, see log tail below. I'm going to look into this and see if I can get it working this afternoon. The bit that worries me most is that it's doing stuff in my `$HOME` dir, where I've got all sorts of valuable files that I would rather not have the tests touch. (E.g. I already have a multitude of files in `~/bin` some of which I may not actually want to be made executable with `chmod +x`.) I wonder if providing an alternative dir for `$HOME` would be enough to write these files in `/tmp` ... ?

```
2017/02/15 10:12:12 util.go:39: Templating "/Users/ssalisbury/go/src/github.com/opentable/sous/integration/test-homedir" -> "/tmp/sous-work/home".
2017/02/15 10:12:12 util.go:105: Linked 1 files, Templated 10 files.
--- FAIL: TestShellLevelIntegration (37.31s)
    --- FAIL: TestShellLevelIntegration/Test_environment_setup (0.32s)
        sh_test.go:142: Error in Test environment setup: 
                        1   26  chmod go-rwx -R ~/dot-ssh
                1   27  chmod +x -R ~/bin/*
                255   28  ssh -o ConnectTimeout=1 -o PasswordAuthentication=no -F "${HOME}/dot-ssh/config" root@192.168.99.100 -p 2222 /reset-repos < /dev/null
        shelltest.go:116: Shell script, output and errors written to "/Users/ssalisbury/go/src/github.com/opentable/sous/doc/shellexamples".
FAIL
FAIL    github.com/opentable/sous/integration   62.453s
make: *** [test-integration] Error 1
```

ext/docker/names.go
samsalisbury: > Outright rejection isn't an option for these tests to work.

I'm not sure I follow? The rejection would be at a very high level in Sous, namely when parsing a SourceLocation, it would be an error if that string contains anywhere a colon, tilde, etc, so we would never end up with SourceLocations that caused issues when converting to Docker repo names. Of course, this would exclude from using Sous any project that uses characters outside of `[a-z0-9_\-\.]` (for example, not saying exactly that pattern). But I don't think that will affect anyone at OpenTable, and will let us kick that can down the road a bit further...

> I do think we're even more likely to see a revolt if we start adding docker images named "github.com/opentable/project"

I can't imagine why anyone would have a problem with a docker repo named `github.com/opentable/project` or similar.. It's precise and makes obvious what should be inside it without adding any maintenance overhead for users of Sous (i.e. maintaining a unique docker repo name).

graph/graph.go
nyarly: https://blog.acolyer.org/2017/02/02/fencing-off-go-liveness-and-safety-for-channel-based-programming/

Also, note that we're a ways off from 100% test coverage, and it'll be a challenge to get there.

samsalisbury: Ah, that paper is talking about the deadlock detector, which is distinct from the race detector.

ext/docker/names.go
nyarly: Of necessity, these tests set up a git repo at e.g. 192.168.99.100:2222, so leading digit and a colon in the git remote name. If Sous were to simply fail to `sous init` because of a badly chosen remote, these tests right here wouldn't be able to proceed. (Whatever) code that can handle that remote will handle other (maybe not 100%, but more) legitimate git remotes being used as Docker repositories.

nyarly: I guess the alternative would be to change these tests so that they somehow present the git server on a name that can be used as a docker registry... but that'd be more code to write without any benefit apart from getting the tests passing.

samsalisbury:
Some more detail on the above error...

After running `make test` the file `shellexamples/002_Test environment setup.blended` ends with this:
```
bash-3.2$ cd /tmp/sous-work
bash-3.2$ chmod go-rwx -R ~/dot-ssh
chmod: -R: No such file or directory
bash-3.2$ chmod +x -R ~/bin/*
chmod: -R: No such file or directory
bash-3.2$ ssh -o ConnectTimeout=1 -o PasswordAuthentication=no -F "${HOME}/do
<eout=1 -o PasswordAuthentication=no -F "${HOME}/dot                         -ssh/config" root@192.168
<cation=no -F "${HOME}/dot-ssh/config" root@192.168.                         99.100 -p 2222 /reset-rep
<ssh/config" root@192.168.99.100 -p 2222 /reset-repo                         s < /dev/null
Warning: Permanently added '[192.168.99.100]:2222' (ECDSA) to the list of known hosts.
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0644 for '/tmp/sous-work/home/dot-ssh/git_pubkey_rsa' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/tmp/sous-work/home/dot-ssh/git_pubkey_rsa": bad permissions
Permission denied (publickey,password,keyboard-interactive).
bash-3.2$ 
bash-3.2$ 
```

ext/storage/git_state_manager.go
nyarly: Doesn't this code assume, actually, that Sous Server has a global author set (which is does) - CI agents would send an empty SOUS_USER, which would then lead to the server omitting --author and defaulting to the Sous Server git config?

Regardless, we need a better story here.

PR #328, opened 2017-02-16
Treat Offset "" and Flavor "" as real values.

nyarly:
This is in response to a StatusPoller bug I observed: empty offset or flavors
match all offsets or flavors, even though empty offsets and flavors are
(relatively) common. Maybe the rest of the filterable fields should be treated
the same way (with the defaults being "*" instead of "")? But in the meantime
this was a quick fix for an immediate problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ with respect to its command line interface and HTTP interface.
 
 ## [Unreleased]
 
+
 ### Added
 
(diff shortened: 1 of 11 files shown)


samsalisbury:
Hmm, this isn't compiling right now. I had a quick look and I can get it to compile by replacing `f.Offset` with `f.Offset.Match` in a few places , however I'm worried there may be an important commit missing from this PR, so I'll hold off trying to fix this until you're sure you've checked everything in.

PR #330, opened 2017-02-20
singularity: collect failed deploys from deploy history

samsalisbury:
This is a horribly large PR with the result that DeployStates collected from Singularity can now include DeployStatusSucceeded and DeployStatusFailed.

Best place to start is probably `ext/singularity/runningdeployments_test.go`

TODO in follow-up PRs...

- [ ] Ensure the exact same deployment is not attempted again when it failed last time.
- [ ] Surface deploy status in `sous query ads`
- [ ] Quieten log output from coaxer unless debugging.
- [ ] Handle missing image labels (ads.go:29)
- [ ] Add more context to logs.
- [ ] Ensure `sous plumbing status` eventually times out.
- [ ] Ensure `sous plumbing status` correctly exits non-zero when deployment fails for any reason.

Related, not required for this story:
- [ ] Allow configuring healthcheck options in GDM (i.e. timeouts, URLs)


nyarly:
Re followup stuff: seems to me that while several of those items are important, only a few actually block our consumers. Notably:

- [ ] Ensure `sous plumbing status` eventually times out.
- [ ] Ensure `sous plumbing status` correctly exits non-zero when deployment fails for any reason.

The remainder can either be split into separate work or get picked up in by-blows of future work. What do you think?

nyarly:
CLI tests indicate that the deploy hangs forever because the left cluster's /status endpoint doesn't include sous-demo in its response.

nyarly:
Closed as completed.

PR #331, opened 2017-02-21
Quick stab at mechanizing WIP

nyarly:
Concept here is simple: there's an extra make target called "wip" that creates
and commits a specially named file to the repo. There's a Travis build script
step that fails if that file exists. Since we don't merge failing PRs, there's
no chance we'll accidentally merge an outstanding PR with the WIP tag file.

On the other hand, this means that using a WIP PR to build on Travis becomes
slightly more annoying, since you'll need to inspect the logs to check for
issues. The ideal case would be a separate check that simply looked for the
file and issued a failing check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ install:
 - "./bin/vendor-check"
 script:
 - make -j2 test && make coverage
-- cat doc/shellexamples/*.blended
+- make reject_wip
+  #- cat doc/shellexamples/*.blended
 after_success:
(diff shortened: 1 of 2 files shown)


samsalisbury:
I'm a bit uneasy about commits indicating personal workflow rather than interesting changes to Sous, but let's try it out. I think I'll be tempted to squash these commits out of existence in the final PR sent for review; I wonder how hard it would be to automate doing this?

PR #333, opened 2017-03-01
Corrects search for failed deploys

nyarly:
Turns out it was a very good idea to write integration test for the failed
deploy case. The assumption was that a failed deploy would remain as either the
active or pending deploy for the request, but this isn't the case.

Instead, for requests with neither active nor pending deploys, we need to do one
extra HTTP request to get their recent deploy history.

On the one hand, this GET is unavoidable: we don't know the current deploy
ID without it. On the other, the fraction of all Requests that will require
this request should generally be low - since it's only Requests that are in
a failed state.


ext/singularity/deployment_builder.go
```diff
@@ -137,13 +138,49 @@ func (db *deploymentBuilder) determineDeployStatus() error {
...
 		db.Target.Status = sous.DeployStatusActive
 		db.depMarker = rds.ActiveDeploy
-	} else {
-		return malformedResponse{"Singularity deploy state included no dep markers. ReqID: " + reqID(rp)}
 	}
 	return nil
 }
 
 func (db *deploymentBuilder) retrieveDeploy() error {
+	if db.depMarker == nil {
```
samsalisbury: I think this check will only work where the first deployment to a request fails. In the case that there are successful deployments first, we'll still have an active deployment (the last successful one) so we'll miss the failure. I'll get an integration test together now to check if that's the case.

samsalisbury: Attempting to write a test showing a failed deploy following a successful one, I realised there was no test for successful deploy status. It looks from this test that all deployments are showing up as failed... https://github.com/samsalisbury/sous/commit/36409d57ca3dfddff3f80714a097c166034c4013#diff-3032bdd5c889553a7a1a73af5a64173fR228

nyarly: This is addressed now, but simply ignoring the presence of an active deploy. If there is one (with no Pending), then it as I understand it, the active deploy will be in the history. While we could avoid the GetDeploys for that marker, we'd still need to compare sequencing with the top of the deploy history and the active deploy...

nyarly:
I need to call it a day. I think it'd be ideal if we could do a successful deploy followed by a failed in tests (ideally in the singularity integration tests, since those are still fasters and more compatible than the CLI smoke tests.)

I think there's a lot of worthwhile stuff in the other PR, which would be worth bringing in here. Even if this isn't 100%, it's now, I think, very close.

samsalisbury:
(wip) test for succeed then fail... https://github.com/samsalisbury/sous/commit/5729d29716350a65f8193dead260e077333d5346

PR #336, opened 2017-03-06
Bugfix init offset

samsalisbury:
This is a little noisy and corrects a number of issues. See commits for details.

Most important bit is that a bug is fixed where 'sous init' (and every other command that relies on graph.TargetManifest) was not correctly picking up the source location offset. See the self review which follows for details.

### DIFF
--- a/graph/targetmanifest.go
+++ b/graph/targetmanifest.go
@@ -5,6 +5,7 @@ import (
 	"github.com/pkg/errors"
 )
 
+// RefinedResolveFilter is a sous.ResolveFilter refined by user-requested flags.
 type RefinedResolveFilter sous.ResolveFilter
 
 func newRefinedResolveFilter(f *sous.ResolveFilter, discovered *SourceContextDiscovery) (*RefinedResolveFilter, error) {
--- a/graph/targetmanifest.go
+++ b/graph/targetmanifest.go
@@ -10,7 +10,7 @@ type RefinedResolveFilter sous.ResolveFilter
 
 func newRefinedResolveFilter(f *sous.ResolveFilter, discovered *SourceContextDiscovery) (*RefinedResolveFilter, error) {
 	c := discovered.GetContext()
-	if f == nil { // XXX I think this needs to be supplied anyway by consumers...
+	if f == nil { // XXX I think this needs to be supplied anyway by consumers..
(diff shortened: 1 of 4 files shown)


graph/targetmanifest.go
```diff
@@ -22,7 +23,7 @@ func newRefinedResolveFilter(f *sous.ResolveFilter, discovered *SourceContextDis
 	if repo == "" {
 		return nil, errors.Errorf("no repo specified, please use -repo or run sous inside a git repo with a configured remote")
 	}
-	if !f.Offset.All {
+	if !f.Offset.All && offset.Match == "" {
```
samsalisbury: NOTE: This is the only functional change in this PR, and fixes the bug described.

graph/targetmanifest.go
```diff
@@ -34,20 +35,20 @@ func newRefinedResolveFilter(f *sous.ResolveFilter, discovered *SourceContextDis
 	return &rrf, nil
 }
 
-func newTargetManifestID(rrf *RefinedResolveFilter) (tmid TargetManifestID, err error) {
+func newTargetManifestID(rrf *RefinedResolveFilter) (TargetManifestID, error) {
```
samsalisbury: Stylistic refactor for readability only.

lib/source_context_test.go
```diff
@@ -19,3 +21,24 @@ func TestVersion(t *testing.T) {
 	assert.Equal("sub", string(id.Location.Dir))
 	assert.Equal("1.2.3", id.Version.String())
 }
+
+func TestNormalisedOffset_nosymlinks(t *testing.T) {
```
samsalisbury: My first suspicion was this which I couldn't find a test for, turned out it was working fine, but now has an explicit test.

test/projects/fail-immediately/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM busybox
+CMD echo "This image fails immediately."; exit 1
```
samsalisbury: I want to 'sous init' in this directory, but 'sous init' was not picking up the right offset, which issue is fixed in this PR.

samsalisbury:
Bump @nyarly @alrs any chance of a review?

PR #339, opened 2017-03-08
Minor integration improvements

alrs:
I was debugging integration tests today, so I came up with improvements that helped me.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -230,7 +230,7 @@ func (suite *integrationSuite) TestFailedService() {
 func (suite *integrationSuite) TestSuccessfulService() {
 	if os.Getenv("TRAVIS") == "true" {
 		// Note: I would normally use t.Skipf() for this, but it isn't part of suite.
-		log.Println("SKIPPING TestSuccessfulService in Travis")
+		log.Println("SKIPPING TestSuccessfulService() in Travis")
 		return
 	}
 	clusters := []string{"test-cluster"}
@@ -252,7 +252,7 @@ func (suite *integrationSuite) TestSuccessfulService() {
 func (suite *integrationSuite) TestFailedDeployFollowingSuccessfulDeploy() {
 	if os.Getenv("TRAVIS") == "true" {
 		// Note: I would normally use t.Skipf() for this, but it isn't part of suite.
-		log.Println("SKIPPING TestSuccessfulService in Travis")
+		log.Println("SKIPPING TestFailedDeployFollowingSuccessfulDeploy() in Travis")
 		return
 	}
 	clusters := []string{"test-cluster"}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -117,7 +117,7 @@ test-unit:
 
 test-integration: test-setup
 	go test -c -tags integration ./integration
-	SOUS_QA_DESC=$(QA_DESC) go test $(TEST_VERBOSE) ./integration --tags=integration
+	SOUS_QA_DESC=$(QA_DESC) go test -timeout 30m $(TEST_VERBOSE) ./integration --tags=integration
```
samsalisbury:
```
$ make test
go test -c -tags integration ./integration
SOUS_QA_DESC=`pwd`/qa_desc.json go test  ./integration --tags=integration
ok      github.com/opentable/sous/integration   90.604s

real    2m13.243s
user    1m17.782s
sys     0m19.466s
```

nyarly: They occasionally will (especially e.g. on Travis), and then the go test timeout clobbers other information we might get.

Maybe the right approach here is to expend TEST_VERBOSE into TEST_OPTIONS and put a -timeout flag based on environment.  I'd also like to see SOUS_TEST_EXTRA or something so that we could toss flags at the test make targets.

integration/integration_test.go
```diff
@@ -214,45 +214,51 @@ func (suite *integrationSuite) TestFailedService() {
...
+		time.Sleep(time.Millisecond * time.Duration(sleepMillis))
 	}
 	suite.statusIs(succeeds, sous.DeployStatusActive)
 }
 
 func (suite *integrationSuite) TestFailedDeployFollowingSuccessfulDeploy() {
 	if os.Getenv("TRAVIS") == "true" {
 		// Note: I would normally use t.Skipf() for this, but it isn't part of suite.
-		log.Println("SKIPPING TestSuccessfulService in Travis")
+		log.Println("SKIPPING TestFailedDeployFollowingSuccessfulDeploy() in Travis")
```
alrs: I tried to add a docker-compose pull in Makefile test-setup, but it turned out that was redundant. I've pulled out the longer timeout entirely, we can visit it if we have problems again.

integration/integration_test.go
```diff
@@ -214,45 +214,51 @@ func (suite *integrationSuite) TestFailedService() {
...
-		time.Sleep(time.Millisecond * 500)
+		log.Printf("sleeping for %s", sleepTime)
+		time.Sleep(sleepTime)
 	}
 	suite.statusIs(fails, sous.DeployStatusFailed)
 }
 
 func (suite *integrationSuite) TestSuccessfulService() {
 	if os.Getenv("TRAVIS") == "true" {
 		// Note: I would normally use t.Skipf() for this, but it isn't part of suite.
```
nyarly: @samsalisbury quick note, you can suite.T() to get access to the testing.T the suite is wrapping

PR #340, opened 2017-03-08
Deploy timeout

samsalisbury:
'sous deploy' times out 30 minutes after successfully writing to the GDM (by way of nested 'sous plumbing status' call).

### DIFF
--- a/cli/sous_plumbing_status.go
+++ b/cli/sous_plumbing_status.go
@@ -1,8 +1,10 @@
 package cli
 
 import (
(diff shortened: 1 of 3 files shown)


samsalisbury:
Ideally the context should be passed through to each http request through the sub pollers etc, however that would be significant extra work so I'm not including it in this PR. The core sous.HTTPClient interface methods should probably be changed to accept a context, and at that point we can pass the context down.

PR #347, opened 2017-03-20
Log environment variable configuration overrides.

alrs:
I want to be aware when Sous is overriding default configuration with values taken from environment variables.

This works, but feels wrong. Alternate approaches welcome.

### DIFF
--- a/util/configloader/configloader.go
+++ b/util/configloader/configloader.go
@@ -171,6 +171,9 @@ func (cl *configLoader) overrideField(sf reflect.StructField, originalVal reflec
 	if !present {
 		return nil
 	}
+	// I'd rather be logging here, but I'm tied up in knots figuring this out. I see a cl.Info()
+	// commented-out in this file, so I suspect I'm in good company.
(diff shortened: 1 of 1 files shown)


util/configloader/configloader.go
```diff
@@ -171,6 +171,9 @@ func (cl *configLoader) overrideField(sf reflect.StructField, originalVal reflec
 	if !present {
 		return nil
 	}
+	// I'd rather be logging here, but I'm tied up in knots figuring this out. I see a cl.Info()
+	// commented-out in this file, so I suspect I'm in good company.
+	fmt.Printf("Environment configuration OVERRIDE: %s=%s\n", envName, envVal)
```
nyarly: This was one of a couple of places we experimented with a logging interface so that we could do away with a global logger.

`Cl.Debug` should work. It's not clear if you tried it or not. In general, I think logging unconditionally is too verbose.

alrs:
From hipchat:

Judson Lester·12:18 PM
Lgtm.

PR #348, opened 2017-03-22
Treat the configured docker registry as canonical

nyarly:
Updates the NameCache to record the canonical name of images as being on the
registry host it was configured with as opposed to the registry name returned
in the metadata.

Consequence of this change was that integration tests needed to be updated to
configure the registry.

The returned registry name, if it differs for some reason, is recorded as well,
to minimize unneeded redeploys.


alrs:
My logging concerns were discussed out-of-band and addressed.

PR #354, opened 2017-03-30
Adding a env var to be able to pass extra build flags to go

nyarly:


alrs:
`make test-unit` and `make test-integration` lgtm when not setting extra environment variables.

PR #355, opened 2017-03-31
Binding the "intended deploytments" to the phase is statuses.

nyarly:
Here's the problem: the /status endpoint reports two things: the list of
deployments when the current resolution cycle started, and the resolutions that
have been completed so far. From those two things, the client can figure out
what's up with the deployment it cares about, effectively by answering two
questions: first, is the server working on my deployment? Second, has it
resolved yet?

Because this is polling an ongoing process, we in fact keep two status lists:
the most recent "complete" status, and the current live one. The in progress
list is fresher, but the stable one has a resolution for everything.

Trouble was, we were sharing a single list of deployments between both, which
meant for updates (i.e. not the first successful deployment) the list would
include the new update, and the stable list would include "it's settled", so it
looked like a stable status.

This is fixed by attaching a copy of the intended deployments to each of the
two status reports, so that resolutions tie to their intentions. Obvious when
you think about it, I suppose.

The server continues to serve the buggy common depenedencies so that old
clients won't freak out, and the client will use the common list if the server
doesn't provide the new ones - otherwise it would never see the intention to
deploy the service it was waiting on. Those concessions to backwards
compatibility will need to be torn out at some point, maybe even soon.

### DIFF
--- a/ext/docker/image_mapping.go
+++ b/ext/docker/image_mapping.go
@@ -239,12 +239,12 @@ func (nc *NameCache) getImageName(sid sous.SourceID) (string, strpairs, error) {
 	if _, ok := errors.Cause(err).(NoImageNameFound); !ok {
 		// We got a probable database error, give up.
 		Log.Info.Printf("Cache error: %s", err)
-		return "", nil, errors.Wrap(err, "getting name from cache")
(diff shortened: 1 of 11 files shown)


alrs:
Discussed API versioning on Hipchat.

PR #358, opened 2017-04-05
Split containers

nyarly:
projects can specify Dockerfiles which are the builder for a intentionally very
limited deploy container.


nyarly:
@samsalisbury Could I get you to review this? @alrs and I had a long talk about it - he has some reservations about the approach, which he'll discuss with you at standup. In the meantime, it does _work_ and I've started setting up some of our basic projects against it to try it out.

Ultimately, my feeling is this: probably a standard build system is going to have detractors, objections, and legitimate flaws. I don't see concrete catastrophic problems with this approach, and I'd like to put it in front of people, and actually use it, to get a sense of what it's like and how it could be improved.

ext/docker/splitcontainer_buildpack_test.go
```diff
@@ -0,0 +1,197 @@
...
+		map[string]docker_registry.Metadata{
+			".*blub-builder.*": {
+				Env: map[string]string{"SOUS_BUILD_MANIFEST": "/sous-manifest.json"}},
+		})
+	assertAccepted(t, dr, err)
+	assertArgs(t, dr, true, true)
+
+	// n.b. Docker does not record ARG lines in containers, so there's no way for
+	// the build container to expose APP_VERSION or APP_REVISION
+	// Perhaps we should consider ENVs for those?
```
samsalisbury: If the final build of the runnable docker image used 'sous build' as well (with explicit flags to maintain context) then the templated Dockerfile could just contain these ARGs in the case that the build Dockerfile contained them.

ext/docker/splitcontainer_buildpack_test.go
```diff
@@ -0,0 +1,197 @@
...
+	sb := &splitBuilder{
+		Manifest: &SplitBuildManifest{
+			Container: sbmContainer{From: "scratch"},
+			Files: []sbmInstall{
+				{Source: sbmFile{Dir: "src"}, Destination: sbmFile{Dir: "dest"}},
+			},
+			Exec: []string{"cat", "/etc/shadow"},
+		},
+		VersionConfig:  "APP_VERSION=1.2.3",
+		RevisionConfig: "APP_REVISION=cabbagedeadbeef",
```
samsalisbury: `cabba9edeadbeef`? :P

ext/docker/splitcontainer_buildpack_test.go
```diff
@@ -0,0 +1,197 @@
...
+	dockerfile := buf.String()
+	hasString := func(needle string) {
+		if strings.Index(dockerfile, needle) == -1 {
+			t.Errorf("No %q in dockerfile.", needle)
+		}
+	}
+	hasString("FROM")
+	hasString("APP_VERSION")
+	hasString("COPY dest dest")
+}
```
samsalisbury: I feel like the above assertions could be stronger? Or just one assertion on total Dockerfile contents?

ext/docker/splitcontainer_buildpack_test.go
```diff
@@ -0,0 +1,197 @@
...
+	sb.Manifest = &SplitBuildManifest{}
+	dec := json.NewDecoder(mBuf)
+	dec.Decode(sb.Manifest)
+
+	if sb.Manifest.Container.From != "scratch" {
+		t.Error("Manifest didn't load Container.From")
+	}
+
+	if len(sb.Manifest.Files) != 1 {
+		t.Error("No files loaded")
```
samsalisbury: `t.Fatal` and skip the `else` ?

nyarly:
@samsalisbury I have a morning obligation, so I'll be missing stand-up.

In general, while the simplest implementations will have a user COPY the manifest file into their build Dockerfile, the general intention is that the build Dockerfile be FROM a parent container that would be the source of that file. I take your concern, and it's already handled. Some build chains might not have a static file (e.g. a Maven build producing snapshot jars), so the build container would produce that file during it's build process.

As to naming: Absolutely no objections. I'll make those changes today.

ext/docker/splitcontainer_buildpack_test.go
nyarly: To clarify: `sous build` performs the whole build, including the final packaging. I've opted to include a templated ENV instruction in the run image Dockerfile for that purpose. The args for the build image exist in the event that the build process wants to use those values in the build process e.g. to edit a package.json.

PR #361, opened 2017-04-12
Gdm git conflict

samsalisbury:
Addresses TECHOPS-2164

### DIFF
--- a/ext/storage/git_state_manager.go
+++ b/ext/storage/git_state_manager.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


ext/storage/git_state_manager.go
```diff
@@ -97,25 +98,58 @@ func (gsm *GitStateManager) WriteState(s *sous.State, u sous.User) error {
...
+			return err
+		}
+		if err := gsm.git("cherry-pick", newTag); err != nil {
+			// If cherry-pick fails, then there's a real conflict.
+			sous.Log.Warn.Printf("attempt to rectify conflicts with git cherry-pick failed: %s", err)
+			if err := gsm.git("cherry-pick", "--abort"); err != nil {
+				sous.Log.Warn.Printf("cherry-pick --abort failed: %s", err)
+				return err
+			}
+			sous.Log.Debug.Printf("Successfully cherry-picked new changes, re-attempting push.")
```
samsalisbury: I noticed today that this block looks totally bogus. It solves the issue as far as tests are concerned, but the logic is clearly wrong here. I'll send a folllow-up PR with a test that fails because of this and a fix for the code. cc @nyarly @alrs FYI

samsalisbury: Also, let's panic if cherry-pick fails as discussed in standup.

PR #362, opened 2017-04-12
Adds a test for deleted manifests

nyarly:
The first time I saw issues with the git strategy, it was with deletes, so I
wanted to be sure there was a test for that.

### DIFF
--- a/ext/storage/git_state_manager_test.go
+++ b/ext/storage/git_state_manager_test.go
@@ -194,6 +194,55 @@ func TestGitConflicts(t *testing.T) {
 
 	sameYAML(t, actual, expected)
 }
+func TestGitRemoteDelete(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	gsm, remote := setupManagers(t)
+
+	actual, err := gsm.ReadState()
+	require.NoError(err)
+
+	expected := exampleState()
+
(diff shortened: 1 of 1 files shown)


ext/storage/git_state_manager_test.go
```diff
@@ -194,6 +194,55 @@ func TestGitConflicts(t *testing.T) {
 
 	sameYAML(t, actual, expected)
 }
+func TestGitRemoteDelete(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	gsm, remote := setupManagers(t)
+
+	actual, err := gsm.ReadState()
```
samsalisbury: Staticcheck is complaining 'actual' not being used. I'll sort and merge.

ext/storage/git_state_manager_test.go
```diff
@@ -194,6 +194,55 @@ func TestGitConflicts(t *testing.T) {
...
+	require.NoError(err)
+
+	expected := exampleState()
+
+	expected.Manifests.Add(&sous.Manifest{Source: sous.SourceLocation{Repo: "github.com/opentable/brandnew"}})
+	remote.WriteState(expected)
+
+	expected, err = remote.ReadState()
```
samsalisbury: Staticcheck complaining about this value of 'expected' not being used, I'll sort and merge.

PR #367, opened 2017-04-19
Adds an opaque "ExecutorData" field

nyarly:
Singularity stashes its requestID there, and picks it back up to do modify actions.

There's also sundry improved logging, and I've started the idea od
ExecutorMessage for failures - the idea, eventually, is to thread that back
through /status so that `sous deploy` can report details from Singularity - so
that when your task dies on launch, you can at least follow a link to determine
the cause.

### DIFF
--- a/integration/hello-server-labels/nginx.conf
+++ b/integration/hello-server-labels/nginx.conf
@@ -22,7 +22,7 @@ http {
         }
 
         location ~ \.php$ {
-            fastcgi_pass   127.0.0.1:9000;
+            fastcgi_pass   127.0.0.1:9100;
             fastcgi_index  index.php;
             include         fastcgi_params;
             fastcgi_param   SCRIPT_FILENAME    $document_root$fastcgi_script_name;
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -110,7 +110,7 @@ func (suite *integrationSuite) waitUntilNotPending(clusters []string, sourceRepo
 
 func (suite *integrationSuite) statusIs(ds *sous.DeployState, expected sous.DeployStatus) {
 	actual := ds.Status
-	suite.Equal(actual, expected, "deploy status is %q; want %q\nIn: %#v", actual, expected, ds)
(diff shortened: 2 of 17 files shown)


integration/hello-server-labels/nginx.conf
```diff
@@ -22,7 +22,7 @@ http {
         }
 
         location ~ \.php$ {
-            fastcgi_pass   127.0.0.1:9000;
+            fastcgi_pass   127.0.0.1:9100;
```
nyarly: I was having weird local issues with testing. Still not 100% sure this was the fix

integration/integration_test.go
```diff
@@ -94,6 +93,26 @@ func (suite *integrationSuite) newNameCache(name string) *docker.NameCache {
 	return docker.NewNameCache(registryName, suite.registry, db)
 }
 
+func (suite *integrationSuite) waitUntilNotPending(clusters []string, sourceRepo string) *sous.DeployState {
```
nyarly: Extracting repeated exercise from the tests.

integration/integration_test.go
```diff
@@ -386,7 +337,7 @@ func (suite *integrationSuite) TestResolve() {
 	stateTwoThree := sous.State{
 		Defs: clusterDefs,
 		Manifests: sous.NewManifests(
-			suite.manifest(suite.nameCache, "opentable/two", "test-two", repoTwo, "1.1.1"),
+			suite.manifest(suite.nameCache, "opentable/two", "test-two-updated", repoTwo, "1.1.2"),
```
nyarly: The resolve tests weren't testing for *updates* to deploys, which is how the issue this PR fixes got through in the first place. This change makes sure that we actually issue new deploys to the same request.

integration/integration_test.go
```diff
@@ -424,18 +375,18 @@ func (suite *integrationSuite) TestResolve() {
 
 	// XXX Let's hope this is a temporary solution to a testing issue
 	// The problem is laid out in DCOPS-7625
-	for tries := 0; tries < 3; tries++ {
+	for tries := 50; tries > 0; tries-- {
```
nyarly: Unfortunately, that meant that the previous deploy had to actually complete, which can take a while. 50 was 2x what I needed for reliable test passing.

PR #370, opened 2017-04-24
Shorten SinguarityRequest ID

alrs:


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ with respect to its command line interface and HTTP interface.
 - Singularity RequestIDs are generated with a suffix of the MD5 sum of
   pre-slug data instead of a random UUID.
 
+### Major
+- Singularity RequestIDs are shortened to no longer include FQDN or
+  organization of Git repo URL.
+- DeploymentID receives the Digest() function.
+- SourceLocation receives the ShortName() function.
```
samsalisbury: Hi, I know this PR is WIP, but just noticed this... I'm not sure adding new functions constitutes a major version bump here. Minor bump sure. Also, "receives" was confusing for me, I thought that meant DeploymentID and SourceLocation constructors were receiving digests and short names as arguments. Maybe "new func `DeployID.Digest` would be simpler?

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ with respect to its command line interface and HTTP interface.
 - Singularity RequestIDs are generated with a suffix of the MD5 sum of
   pre-slug data instead of a random UUID.
 
+### Major
```
nyarly: Per http://keepachangelog.com/en/0.3.0/, the subheads of a version should be "Added", "Changed", "Fixed", "Deprecated", or "Security" - I think this is a "Changed" entry.

lib/deployment_test.go
```diff
@@ -47,11 +48,7 @@ func TestDeploymentEqual(t *testing.T) {
 	dep := Deployment{}
 	assert.True(dep.Equal(&Deployment{}))
 
-	other := Deployment{
-		Annotation: Annotation{
-			RequestID: "somewhere around here",
-		},
-	}
+	other := Deployment{}
```
nyarly: This part of the test no longer makes sense, since the point was that Annotations didn't participate in equality. With annotations gone, the second test isn't doing anything.

lib/source_location_test.go
```diff
@@ -57,3 +57,20 @@ func TestSourceLocation_UnmarshalText(t *testing.T) {
...
 	}
 }
+
+func TestSourceLocationShortName(t *testing.T) {
+	expected := "project"
+	sl := SourceLocation{
+		Repo: "fake.tld/org/" + expected,
+		Dir:  "down/here",
+	}
+	shortName, err := sl.ShortName()
```
nyarly: Test also for the error case?

CHANGELOG.md
pennello: Minor, for sure.  The API has changed, in that new functionality has been added, but backwards compatibility has been maintained, so no major bump.

CHANGELOG.md
```diff
@@ -7,9 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 with respect to its command line interface and HTTP interface.
 
 ## [Unreleased]
-### Fixed
+### Changed 
 - Singularity RequestIDs are generated with a suffix of the MD5 sum of
   pre-slug data instead of a random UUID.
+- Singularity RequestIDs are shortened to no longer include FQDN or
+  organization of Git repo URL.
```
pennello: Ah, interesting—so we just chose to unconditionally remove this prefix instead of having a setting.  Reasonable, simple place to start.

ext/singularity/deployer.go
```diff
@@ -124,7 +122,10 @@ func (r *deployer) RectifySingleCreate(d *sous.DeployablePair) (err error) {
 	if err != nil {
 		return err
 	}
-	reqID := computeRequestID(d.Post)
+	reqID, err := computeRequestID(d.Post)
+	if err != nil {
```
pennello: Do you guys avoid the style `if reqId, err := computeRequestID(d.Post); err != nill {`?  I guess it would be a little long in this case.

ext/singularity/deployer.go
```diff
@@ -229,33 +230,27 @@ func changesDep(pair *sous.DeployablePair) bool {
...
+func MakeRequestID(depID sous.DeploymentID) (string, error) {
+	sn, err := depID.ManifestID.Source.ShortName()
+	if err != nil {
+		return "", err
+	}
+	sn = illegalDeployIDChars.ReplaceAllString(sn, "_")
+	dd := illegalDeployIDChars.ReplaceAllString(depID.ManifestID.Source.Dir, "_")
 	fl := illegalDeployIDChars.ReplaceAllString(depID.ManifestID.Flavor, "_")
 	cl := illegalDeployIDChars.ReplaceAllString(depID.Cluster, "_")
+	digest := depID.Digest()
```
pennello: Forgive my lack of understanding of the project, but why do we unconditionally add this digest to the end?  Seeing a lot of hex in the UI makes these IDs more difficult to read.  Could we add it only if uniqueness was required (i.e., there would be a collision)?  Or, perhaps, could we add only a sub-portion of the digest, kind of like how Git often shows only the first 7 hex characters of the SHA digest?

ext/singularity/deployer_test.go
```diff
@@ -145,7 +155,7 @@ func TestShortComputeDeployID(t *testing.T) {
 		Deployment: &sous.Deployment{
 			SourceID: sous.SourceID{
 				Location: sous.SourceLocation{
-					Repo: "reqid",
+					Repo: "fake.tld/org/project",
```
pennello: You might consider using `example.com` in these examples.  It's explicitly reserved for such use.

http://example.com/
https://www.iana.org/domains/reserved

lib/deployment.go
```diff
@@ -55,6 +45,20 @@ type (
...
+	sep := []byte{0}
+	io.WriteString(h, depID.ManifestID.Source.String())
+	h.Write(sep)
+	io.WriteString(h, depID.ManifestID.Flavor)
+	h.Write(sep)
+	io.WriteString(h, depID.Cluster)
+
+	return h.Sum(nil)
+}
+
```
pennello: Is a digest really the best approach here?  What if we just made a sufficiently-compact assemblage of the involved information?  It would still be just as unique as the digest, and perhaps much more readable.

ext/singularity/deployer.go
alrs: @nyarly and @samsalisbury are more likely than I to use that pattern. I'm not bothered by the repetition.

lib/deployment.go
nyarly: What do you have in mind? Not long ago we were concatenating the full git repo remote, the build subdirectory, a "flavor" argument and the logical cluster. There was feedback that that was verbose ;)

Quite seriously, though, _some_ uniquifier is needed to ensure that the squashing that happens in naming the request doesn't lead to collisions. If the unique suffix (and we preferred a suffix as more readible) is simply random, there's a mode where Sous would create many empty requests. Once you get to "a compact, stable unique identifier", a digest is what falls out.

ext/singularity/deployer_test.go
nyarly: example.com would definitely be my personal first pick, but the specifics aren't important at the moment - the behavior being tested doesn't even care if the first part is a domain.

ext/singularity/deployer.go
nyarly: (see below) The unconditional digest is in fact to prevent collisions between request IDs. By using hashes we mostly avoid dealing with a check-then-set race condition.

CHANGELOG.md
nyarly: More sophisticated approaches are in discussion, but we wanted to get something reasonable out ASAP.

ext/singularity/deployer.go
samsalisbury: The inline form you noted @pennello is always preferable IMO, except where there are multiple return values, and one or more of them are needed outside the scope of the if block. In this case, reqID is required outside of the if block, so that inline form cannot be used here.

@alrs the point of the inline form is not just to save a line of code, it also limits the scope of those variables to the if block. Very short-scoped variables make code easier to read, as you can totally forget about them once they go out of scope. So where that short scope is possible, I would always rather use it to aid future readers of the code.

lib/deployment.go
samsalisbury: Hmm, I overlooked that we were eliding "flavor" here. We should certainly add flavor back in, as it is very significant for deployments that use it.

Likewise, we should make sure that "cluster" is part of the RequestID, because that is also important for humans reading the request name. (Sous clusters do not map to Singularity clusters, e.g. pp-sf and ci-sf Sous clusters both live on the same qa-sf Singularity cluster).

ext/singularity/deployer.go
pennello: Oh, my bad @samsalisbury; you're right—`reqId` is used below.  👍

CHANGELOG.md
pennello: Right; makes total sense.

lib/deployment.go
pennello: > Not long ago we were concatenating the full git repo remote, the build subdirectory, a "flavor" argument and the logical cluster. There was feedback that that was verbose ;)

Heh.  : ]

> What do you have in mind?

Well, I was trying to suggest an idea in the OP—
> What if we just made a sufficiently-compact assemblage of the involved information?

More specifically, I was imagining something like

    env-repo[-flavor]

But I guess this is a _deployment_ ID, not a _request_ ID (which is what I was sloppily imagining).  A unique hash for those makes more sense.

Right now, it looks like we have the version and the hash, and that's it.  What was difficult when I was trying this out recently was that there was no username indicating who initiated the deployment.  In the pre-existing deploy scripts, we also included a timestamp.  Not strictly necessary, as Singularity keeps track of when deployments are made, but it's largely unique-ifying, and is also often convenient to be able to easily read off the timestamp.

PR #371, opened 2017-04-26
doc/install: add homebrew install instructions

stevenschlansker:


wmelani:
Just leaving a comment here homebrew-osx-tools has internal URLs to our local artifactory, not sure if this project should be public / referenced in this public repo since it won't work outside of OT

doc/install.md
```diff
@@ -4,6 +4,15 @@
 
 Sous is distributed as a single binary, in a tarball also containing documentation.
 
+# Homebrew
+[Install via Homebrew](https://github.com/opentable/homebrew-osx-tools)!
```
nyarly: I only hesitate to add to the main docs references to internal Opentable resources.

Two possiblities: 

* either this is a non-issue and I'm being too fussy, or
* we could easily extract a `homebrew-sous` repo and make that public.

In the latter case I will definitely want to look into setting up a CD pipeline triggered by Sous releases, though.

stevenschlansker: Part of the problem is when I have `sous` as the root of my knowledge-tree, it is hard to find auxiliary projects.  For example I had no idea to look for `homebrew-osx-tools` until someone pointed me at it.

If you want to keep the main Sous documentation "pure" then I think we will need a separate set of documentation "Using Sous at OT" because that's the docs that our main users actually care about (nobody at the company is going to do a side project open source deployment of their own Sous server)

Another option would be to open source the homebrew stuff as well, and make that public too.

nyarly: I think for the interim, opening up the homebrew stuff (or having an open homebrew and an internal) would be the best solution here.

@samsalisbury ?

samsalisbury: The homebrew-osx-tools repo has always been open source, since Homebrew does not play well with HTTP authentication, which some people at OpenTable use. I went the open source way for ease of installation. Having that artifactory link there is wrong. I have a mind to deprecate that formula and remove the link until discover is updated to publish somewhere public.

I think @nyarly's idea of having a separate tap for sous which we can link to is probably the cleanest idea, but will mean breaking the upgrade path for existing homebrew users. However, changing it at a later date would be even harder to do, and I've never liked the name of this repo. 'osx-tools' adds no information especially now there's also http://linuxbrew.sh/ which uses the same formula with alternative binaries.

Whilst this PR is definitely an improvement to having no mention of the homebrew formula, I would rather not add a link that we intend to deprecate soon.

I will see if I can create a generic formula allowing `brew install opentable/sous` (minus osx-tools. If that works, let's link to that and communicate somehow that the old formula is deprecated.

samsalisbury: OK, I've created a new Homebrew tap at [opentable/public](https://github.com/opentable/homebrew-public) and deprecated the osx-tools repo, including adding an explicit migration for the sous formula.

I have also [greatly expanded the README](https://github.com/opentable/homebrew-public/blob/master/README.md) in the new opentable/public repo so there should be enough detail for both users and maintainers. Please read it through to check for any errors, I pushed it straight to master so I could verify it works, which it does on my machine.

Now when you try to install the old version you get redirected to the new one with a warning...

```shell
$ brew install opentable/osx-tools/sous
Warning: Use opentable/public/sous instead of deprecated opentable/osx-tools/sous
==> Installing sous from opentable/public
==> Downloading https://github.com/opentable/sous/releases/download/0.5.1/sous-darwin-amd64_
Already downloaded: /Users/ssalisbury/Library/Caches/Homebrew/sous-0.5.1.tar.gz
🍺  /usr/local/Cellar/sous/0.5.1: 5 files, 15.8MB, built in 0 seconds
```

I will update this PR and merge it.

cc @nyarly @stevenschlansker @alrs

PR #380, opened 2017-05-11
Fixing output for missing config. #trivial

nyarly:
The output looked like

"warn: Missing config file, using defaults%!(EXTRA map[string]interface {}=map[path:/home/teamcity/.config/sous/config.yaml])"

because of a mismatch between Printf and no template. Simple update to output more nicely.
Better still would be to give hints about what to do if that's the wrong path
(e.g. the SOUS_CONFIG_FILE env) but this is a general config loader function.

### DIFF
--- a/lib/http_state_manager.go
+++ b/lib/http_state_manager.go
@@ -20,12 +20,12 @@ type (
 	}
 )
 
-func (g *gdmWrapper) manifests(defs Defs, base Manifests) (Manifests, error) {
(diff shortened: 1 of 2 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

alrs:
I agree with Danger.

PR #381, opened 2017-05-13
Env defaults handled correctly.

nyarly:
Redundant Env defaults retained, while unmentioned defaults come through.
We can re-add sub-flavored deployment env vars now.

Interestingly, this makes use of a Harmony-style "lens":
Manifest/Deployment mapping is performed using a Get/Putback operational pair.
c.f. http://www.seas.upenn.edu/~harmony/

One consequence of this is that the existing behavior
of the HTTPStateManager needed to change: it doesn't have the context
to update the manifests correctly,
so it needs to PUT the new state of the GDM.
This winds up being much simpler in terms of code.

It may be that the result is that updates conflict more often.
We'll see how this goes in practice.
If so, there's a note elsewhere about
a refactor from mode-switching
between client and workstation modes
by means of local vs. network Sous server,
instead of HTTPStateManager vs GitStateManager.
This change would be desirable for other reasons,
but it would address this case in particular.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 1478 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

lib/map_state_to_deployments.go
```diff
@@ -41,8 +46,82 @@ func (s *State) Deployments() (Deployments, error) {
 	return ds, nil
 }
 
-// Manifests creates manifests from deployments.
-func (ds Deployments) Manifests(defs Defs) (Manifests, error) {
+// PutbackManifests creates manifests from deployments.
+func (ds Deployments) PutbackManifests(defs Defs, olds Manifests) (Manifests, error) {
```
alrs: What does PutbackManifests mean? It returns "Manifests," so I gather it isn't putting anything anywhere.

PR #389, opened 2017-05-26
Added an artifact names document

nyarly:
Trying to sketch out the problem cases with names and images,
and how (I believe) advisories can help solve that issue.

### DIFF
--- a/doc/artifact_names.md
+++ b/doc/artifact_names.md
@@ -1,5 +1,11 @@
 # Artifact Naming
 
+**nB: as of June 13, 2017,
+this document should be considered
+a design proposal.
+It's mostly true,
+with the exception of the "rebuilt" advisory tag.**
+
 Sous takes the opinionated stance that deployed artifacts should be named
 based on the primary repository of their source code,
 and a semantic version.
--- a/doc/artifact_names.md
+++ b/doc/artifact_names.md
@@ -1,0 +1,105 @@
+# Artifact Naming
+
+Sous takes the opinionated stance that deployed artifacts should be named
+based on the primary repository of their source code,
+and a semantic version.
+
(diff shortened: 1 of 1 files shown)


sous-machine-account:
LGTM

</p>

samsalisbury:
Cool, I think this idea would solve some of the problems with rebuilds and build artifact pinning. The only problem I can see with doing this is knowing if the Docker tag is being built for the first time or not. There will be difficult to avoid race conditions here, which make me think a simpler if imperfect solution may be preferable (e.g. timestamping the images, and adding that timestamp to the manifest). If there's a way to quickly overcome the issue of knowing whether this is the first build or not, I would be happy to pursue this as the results of it would be neat.

doc/artifact_names.md
```diff
@@ -0,0 +1,105 @@
...
+a QA system within Sous builds,
+which includes in the labeling on each image
+a list of advisories
+about how the build was in one way or another
+"unhygienic."
+Generally speaking, then,
+for a particular tag,
+the image with the fewest advisories
+should be considered "most canonical."
+The "rebuilt" advisory means that the absolute case
```
samsalisbury: "Rebuilt" as a build advisory is difficult to do as it assumes the builder has access to a canonical list of all build artifacts. Ideally, 'sous build' should work without recourse to a central authority.

Maybe Sous should just refuse to update already-recorded images for a version, and fail the build with a clearly worded error message? Along with that, I would rather 'sous build' did not automatically push or record images, maybe we should have a 'sous publish' command that builds and pushes/records, whilst 'sous build' just builds. 'sous publish' would just push an already built local image if it happened to reflect the present SourceContext.

PR #395, opened 2017-06-01
travis: Bump Go version to 1.8.3

samsalisbury:


sous-machine-account:
LGTM

</p>

samsalisbury:
I think this is enough to show we can move to 1.8.3 for all Sous development... It took a few re-runs of the tests before it passed, so not sure if 1.8.3 is making them more faily or if that was just the usual coin toss happening. My preference is to merge this so we don't lag behind on Go versions too much, and spend some time next sprint addressing any issues with extra failyness if it transpires to be an issue.

nyarly:
While we're worried about quick fixes for consumers, let's hold off on this?

PR #397, opened 2017-06-01
Fixing tests for deploy and request ids #trivial

nyarly:


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

ext/singularity/deployer_test.go
```diff
@@ -58,6 +58,25 @@ func TestMakeRequestID(t *testing.T) {
...
+				Dir:  "and/also/i/bury/my/services/super/deep/in/the/build/tree/for/no/good/reason/blame/maven",
+			},
+			Flavor: "wellwehavetohaveaflavorforthisservicebecausethereseighteeninstancesofitwithweirdquirkstotheirconfig",
+		},
+		Cluster: "foo",
+	})
+	if err != nil {
+		t.Fatalf("Error: %v", err)
+	}
+	if len(actual) >= 100 {
```
samsalisbury: `const maxLen = 99; if len(actual) > maxLen` would be clearer here, no?

ext/singularity/deployer_test.go
```diff
@@ -216,7 +235,7 @@ func TestLongComputeDeployID(t *testing.T) {
 
 	idLen := len(deployID)
 	logLenTmpl := "Got length:%d Max length:%d"
-	if len(deployID) > maxDeployIDLen {
+	if len(deployID) >= 50 { // 50 is how our Singularity is configured
```
samsalisbury: This test would read better when as `if len(deployID) > 49`. Singularity may have chosen to make their configuration confusing, but I don't think we should perpetuate that by making our code more complex.

samsalisbury:
This change introduces a regression, in the case that `len(depBase)` == `50` we will end up with a deployID of 50 characters, which will result in a failed deploy See https://github.com/opentable/sous/blob/f90349e782bd38b7d311bec36423f6ea976dcf4f/ext/singularity/deployer.go#L273 PR to cover this with a test coming... cc @nyarly

samsalisbury:
Test here fails with DeployID == 50 chars https://github.com/opentable/sous/pull/400/commits/215520bd9ce6c953f9e3f6036322d2edd7341f7e fix coming.

PR #400, opened 2017-06-02
singularity: Fix regression with Singularity DeployIDs len > 49

samsalisbury:
Failing test for ComputeDeployID
- When the version string is exactly 17 chars long, we get a DeployID
  50 chars long, which is too long.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/)
 with respect to its command line interface and HTTP interface.
 
+## Unreleased
+### Fixed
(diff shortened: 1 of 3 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

nyarly:
As irritating as it is, the Singularity configuration is "50 characters", so my feeling is that it would be more comprehensible to use the Singularity config as the value. I do really like "exactly 49 characters" as the test though. Eventually, a successful Sous would need this configurable, since other Singularity clusters will have different values configurated for the lengths of deploy ids, and requiring our users to solve our off-by-one problem in configuration seems inhumane.

That said, it's not worth blocking this merge over.

PR #404, opened 2017-06-02
Startup options recognised by Singularity

samsalisbury:
Includes #403 

Our singularity code includes a filter for changes to be applied to requests or deploys. This is to allow some changes to the GDM (e.g. sous-only metadata) to be ignored for the purpose of Singularity.   This change adds Startup as a field that needs to be reflected in Singularity.


samsalisbury:
NOTE: This is not yet covered by any tests, but has worked with a manual rectify. I think we should merge it even without the tests, since that unblocks our customers, and add tests as a priority ticket asap.

sous-machine-account:
LGTM

</p>

nyarly:
Closed in favor of #405

PR #411, opened 2017-06-12
Regression test for env-vars -> new deploy #trivial

nyarly:
Test passes though. If this is a real problem, it exists elsewhere. Still deserves to be in the test though.


sous-machine-account:
LGTM

</p>

PR #414, opened 2017-06-13
Adds additional metadata to client/server/state interactions.

samsalisbury:
This PR adds `sous.StateContext` which can contain some additional information about an interaction with sous state (i.e. the GDM). The StateContext is included in every read or write of State, but at present none of the fields are required. Log lines are emitted when errors parsing values occur.

This is the groundwork for adding much richer messages to GDM commits, which will help us diagnose and fix issues quicker in future. Also gives us a hook to reject certain clients or anonymous requests etc in future if we need to.

```go
// StateContext contains additional data about what is being read or written
// by a StateManager.
type StateContext struct {
	// User is the user a read or write is attributed to.
	User User
	// TargetManifestID is the manifest this write is expected to affect.
	// Implementations of StateWriter.WriteState should check that the
	// change being written corresponds with this manifest ID.
	TargetManifestID ManifestID
	// ClientVersion is the version of the sous client reading or writing state.
	ClientVersion semv.Version
	// Command is the command entered by a user.
	Command string
}
```


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

lib/httpclient.go
```diff
@@ -23,10 +23,10 @@ type (
 	// HTTPClient interacts with a HTTPServer
 	//   It's designed to handle basic CRUD operations in a safe and restful way.
 	HTTPClient interface {
-		Create(urlPath string, qParms map[string]string, rqBody interface{}, user User) error
-		Retrieve(urlPath string, qParms map[string]string, rzBody interface{}, user User) error
-		Update(urlPath string, qParms map[string]string, from, qBody Comparable, user User) error
-		Delete(urlPath string, qParms map[string]string, from Comparable, user User) error
+		Create(urlPath string, qParms map[string]string, rqBody interface{}, ctx StateContext) error
```
nyarly: I'm leery of this change - while it made sense that every HTTP request have a User associated with it, I'm less clear that they should all have a StateContext involved. Specifically, requests that relate to a manifest ID seem like they should all be on /manifests; the client version makes more sense - but then it seems like UserAgent is the right field for that.

lib/state_manager.go
```diff
@@ -3,35 +3,16 @@ package sous
 type (
 	// StateReader knows how to read state.
 	StateReader interface {
-		ReadState() (*State, error)
+		ReadState(StateContext) (*State, error)
```
nyarly: No use of ReadState makes use of the State context. Do you see I place where it might get used? In general I'd prefer not to change the interface here.

lib/statecontext.go
```diff
@@ -0,0 +1,66 @@
...
+
+// StateContext contains additional data about what is being read or written
+// by a StateManager.
+type StateContext struct {
+	// User is the user a read or write is attributed to.
+	User User
+	// TargetManifestID is the manifest this write is expected to affect.
+	// Implementations of StateWriter.WriteState should check that the
+	// change being written corresponds with this manifest ID.
+	TargetManifestID ManifestID
```
nyarly: As mentioned elsewhere: if the request refers to a single manifest, it should be made on a /manifests URI.

Based on discussion, we were planning to check that a git commit only included changes to a single manifest. Transmitting the ManifestID seems like scope creep from that perspective.

I agree that additional context for git commit messages and/or log entries would be extremely useful, but neither requires anything beyond the strings passed by the client in the headers.

If the ManifestID is intended as debugging context, maybe we should send (at least) the SourceID as directed (and perhaps as it would be captured from the SourceContext.)

lib/statecontext.go
```diff
@@ -0,0 +1,66 @@
...
+
+// WriteHeaders writes this StateContext to a set of HTTP headers.
+func (sc *StateContext) WriteHeaders(header http.Header) {
+	header.Add("Sous-User-Name", sc.User.Name)
+	header.Add("Sous-User-Email", sc.User.Email)
+	header.Add("Sous-Target-Manifest-ID", sc.TargetManifestID.String())
+}
+
+// NewStateContextFromHTTPHeader parses a StateContext from an http.Header.
+func NewStateContextFromHTTPHeader(req *http.Request) (StateContext, error) {
```
nyarly: It seems like we could simply assemble strings into a struct here and log the result. I don't understand why this function does more than that.

lib/statecontext.go
```diff
@@ -0,0 +1,66 @@
...
+
+	clientVersionString := req.Header.Get(ClientVersionHeaderKey)
+	clientVersion, err := semv.Parse(clientVersionString)
+	if err != nil {
+		Log.Warn.Printf("Unable to parse client version %q from header %s: %s",
+			clientVersionString, ClientVersionHeaderKey, err)
+	}
+
+	midString := req.Header.Get(TargetManifestIDHeaderKey)
+	mid, err := ParseManifestID(midString)
```
nyarly: Specifically, this is the only non-test code to make use of ParseManifestID.

PR #425, opened 2017-06-20
Non-destructive updates.

nyarly:
A reasonably complete description can be found in docs/api_change_guide.md

### DIFF
--- a/doc/api_change_guide.md
+++ b/doc/api_change_guide.md
@@ -1,0 +1,73 @@
+# API Change Guide
+
+Sous transimits several of its data structures
+via JSON
+between client and server.
+For the most part,
+it should be possible to extend those structs
+without causing problems,
+so long as we follow a few simple rules.
+
+Existing fields should not be changed.
+Their types should remain the same,
+and they certainly shouldn't be removed -
+even if their values are no longer inspected.
+
+New fields can be added reasonably freely,
+with one caveat:
+If the struct is used
+as an item in a slice,
+it's fields should be left as-is.
+As a result,
+it is probably preferable to add
+maps of structs
+rather than
+lists of structs.
+
+The reason for this
+is that in order to
+handle values that the client doesn't understand
+(because, for instance,
+the server has been updated since the client was installed)
+they need to be compared,
+and unchanged values returned unchanged.
+A list can be compared,
+but if any item in the list is changed,
+the whole list has to be considered changed.
+Consider: how would we determine which items were changed,
+and which should be returned untouched.
+So if there's a change anywhere in a list,
+the whole list is replaced in the update.
+Which means that if fields on the items were added,
+they'll be destroyed on update.
+(With a map, it's clear which items changed,
+so we can isolate the change and retain fields we don't understand.)
+
+## Canary Field
+
+To protect against clients that do not yet know how to do these "safe" updates,
+the server adds a field to its JSON responses.
+The name of this field is unique to the request
+(it's simply the Etag already generated).
+
+The server expects to see this field
+in PUTs with "If-Match" set,
+(and extracts the name from the If-Match header)
+and rejects updates without the field.
+
+Properly behaved clients retain the field
+the same way they retain any unrecognized field.
+In fact, its purpose is effectively to be a field
+that **no** client recognizes.
+
+A client that discards the field
+(because it doesn't match a field of its struct definitions)
+signals that it might perform
+inadvertantly destructive updates,
+and so its updates cannot be trusted.
+
+## Compatibility
+
+This change was made after
+version 0.5.14
+so versions of Sous more recent than that should all be fine.
--- a/lib/httpclient.go
+++ b/lib/httpclient.go
@@ -233,57 +233,6 @@ func (client *LiveHTTPClient) getBodyEtag(url string, user User, body Comparable
 	return etag, nil
 }
 
-type jsonMap map[string]interface{}
-
-func putbackJSON(originalBuf, baseBuf, changedBuf io.Reader) *bytes.Buffer {
-	var original, base, changed jsonMap
-	mapDecode(originalBuf, &original)
-	mapDecode(changedBuf, &base)
-	mapDecode(changedBuf, &changed)
-	original = applyChanges(base, changed, original)
-	return encodeJSON(original)
-}
-
-// mutates base
-func applyChanges(base, changed, target map[string]interface{}) map[string]interface{} {
-	for k, v := range changed {
-		switch v := v.(type) {
-		default:
-			if b, old := base[k]; !old {
-				target[k] = v //created
-			} else {
-				delete(base, k)
-				if b != v { // changed
-					target[k] = v
-				}
-			}
-		case map[string]interface{}:
-			// Unchecked cast: if base[k] isn't also a map, we have bigger problems.
-			// If target[k] isn't a map, then the server has changed the type under us, and we should crash
-			target[k] = applyChanges(base[k].(map[string]interface{}), v, target[k].(map[string]interface{}))
-		}
-	}
-
-	// the remaining fields were deleted
-	for k := range base {
-		delete(target, k)
-	}
-
-	return target
-}
-
-func mapDecode(buf io.Reader, into *jsonMap) error {
-	dec := json.NewDecoder(buf)
-	return dec.Decode(into)
-}
-
-func encodeJSON(from interface{}) *bytes.Buffer {
-	buf := &bytes.Buffer{}
-	enc := json.NewEncoder(buf)
-	enc.Encode(from)
-	return buf
-}
-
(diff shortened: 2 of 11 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 645 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

doc/api_change_guide.md
```diff
@@ -0,0 +1,73 @@
...
+(because it doesn't match a field of its struct definitions)
+signals that it might perform
+inadvertantly destructive updates,
+and so its updates cannot be trusted.
+
+## Compatibility
+
+This change was made after
+version 0.5.14
+so versions of Sous more recent than that should all be fine.
```
samsalisbury: Thanks for this thorough description, all makes sense, I like it. I have not yet reviewed the code, but I wonder if this unpredictably-named field should be wrapped inside a descriptively named field, so that users editing their manifests can see clearly what it is and that it is necessary. E.g.

```yaml
ETag:
  <ETAG_VALUE>: "Do not delete."
```

lib/httpclient.go
```diff
@@ -20,13 +21,25 @@ type (
 		http.Client
 	}
 
+	resourceState struct {
+		etag         string
+		body         io.Reader
+		resourceJSON io.Reader
+	}
+
+	// An Updater represents a point in time from which an update can be attempted.
```
samsalisbury: Not sure I understand how this is a "point in time"? Maybe a bit more explanation could be helpful here.

lib/httpclient.go
```diff
@@ -20,13 +21,25 @@ type (
...
+	Updater interface {
+		Update(HTTPClient, string, map[string]string, Comparable, User) error
```
samsalisbury: Better to name these fields since we have bare `string` and `map[string]string` and the method name does not give any indication of the relationships between these parameters and each other, or the implementor of `Update`.

As a public interface, the `Update` method should also have a comment to explain the expected behaviour of implementations.

lib/putbackjson.go
```diff
@@ -0,0 +1,130 @@
...
+	"encoding/json"
+	"io"
+)
+
+type jsonMap map[string]interface{}
+
+func putbackJSON(originalBuf, baseBuf, changedBuf io.Reader) *bytes.Buffer {
+	var original, base, changed jsonMap
+	err := mapDecode(originalBuf, &original)
+	if err != nil {
```
samsalisbury: Any reason to keep a long-lived `err` variable around here? To me, the short scope of inline err checking would aid readability. I had to scan down to the bottom of this func to see if this value was being re-used, it wasn't, but the code could have told me that if it had read:

```go
if err := mapDecode(originalBuf, &original); err != nil {
```

The same goes for each of the lines beginning `err =` below.

lib/putbackjson.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// the remaining fields were deleted
+	for k := range base {
+		delete(target, k)
+	}
+
+	return target
+}
+
+func same(left, right interface{}) bool {
```
samsalisbury: This function is very dense, I think a comment explaining what "same" means would be really helpful here. Is it merely a deep equality?

lib/putbackjson.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+		}
+
+		return true
+	}
+}
+
+func mapDecode(buf io.Reader, into *jsonMap) error {
+	dec := json.NewDecoder(buf)
+	return dec.Decode(into)
```
samsalisbury: Could be a one-liner?

util/restful/canaryattr.go
```diff
@@ -0,0 +1,69 @@
...
+		mh.writeHeaders(http.StatusInternalServerError, w, r, fmt.Sprintf("Error getting body from request: %v", err))
+		return false
+	}
+
+	rbody := bytes.NewReader(rbytes)
+	dump := map[string]interface{}{}
+	dec := json.NewDecoder(rbody)
+	err = dec.Decode(&dump)
+
+	if err != nil {
```
samsalisbury: Again, I'm scanning all over this function to find out where this `err` is re-used outside of the cope of this if block. I can't spot it anywhere, so would have appreciated the ease of reading provided by keeping that `err` tightly-scoped by using inline initialisation, i.e.:

```go
if err := dec.Decode(&dump); err != nil {
```

util/restful/server_test.go
```diff
@@ -78,6 +81,36 @@ func justBytes(b []byte, e error) io.ReadCloser {
...
+
+	mh.renderData(200, rr, rq, data)
+
+	rz := rr.Result()
+	bodyB, err := ioutil.ReadAll(rz.Body)
+	assert.NoError(t, err)
+
+	dump := map[string]interface{}{}
+	err = json.Unmarshal(bodyB, &dump)
+	assert.NoError(t, err)
```
samsalisbury: Again, long-scoped `errs` adding adding to reading time, would suggest:

```go
assert.NoError(t, json.Unmarshal(bodyB, &dump))
```

doc/api_change_guide.md
nyarly: As discussed, this field is meant to be something that the client obliviously echoes - not something to present to users or make any other use of.

If we were going to add an etag field to `sous manifest get`'s output, IMO that'd be handled between getting the HTTP response and rendering the YAML - effectively a parallel usage to what we're doing here.

util/restful/logging.go
```diff
@@ -0,0 +1,33 @@
...
+
+	logSetWrapper struct {
+		logSet
+	}
+)
+
+// PlaceholderLogger returns a log set that fulfills the restful logging
+// interface - a convenience for when you don't want or need to wrap a logger
+// in appropriate interface fulfillment.
+func PlaceholderLogger() logSet {
```
nyarly: PlaceholderLogger is to be used by outside packages in order to supply a restful.logSet back to restful - consumers shouldn't ever *need* to use the return value.

nyarly:
I believe I've made the corrections as requested.

Cleaning up the HTTPClient interface led me to realize that, first, it could be further simplified and second that HTTPClient belongs in the `restful` package. The new changes are dominated by the result of that migration, including some interface-based logging approaches, so the resulting PR is larger still than it was in the first place.

nyarly:
Regarding the long-scoped errs, I'm not unsympathetic. The dual of the problem, though, is errs returned by two-value returns (e.g. `x, err := F()`) can't be used in the `if` conditional - or rather, the scope of the error is the same as the scope of the normal value. In these cases, the errs have to be long-scoped unless the value can be used in the conditional.
 
As a result, I've tried to always use `err` as the lval for `error` type values, and try to be consistent in my assignment - which has led to the long-scoped err situation. I don't know if there's a solution to this; I think it's just a wart of Go.

I'll try to use the `if err :=` construction when it makes sense. It does mean that it contexts where there are calls with both a single error result and dual return, there'll be shadowed `err`s around, which I try to avoid in general.

util/restful/canaryattr.go
nyarly: One thing that would help me: why are you scanning for uses of this err variable? My assumption when I see a variable named `err` is that it'll be checked against nil and returned (or something like), and that it's otherwise not significant. Am I missing some important possibility?

doc/api_change_guide.md
samsalisbury: Right, I get that that's the design, but my understanding is that if we include the etag in the YAML the user gets to edit and send back (as a key, thus unpredictably named) we would be guarding against 2 people racing to edit the same manifest at the same time. It would result in a potentially greater number of ETag mismatches than at present, but I think that's a good thing for consistency? Tools we write to allow small, specific edits to a manifest should reduce the time between `get` and `set` so should have a smaller chance of conflicts.

Either that, or maybe `manifest set` should explicitly mean "set the manifest to exactly this, and don't worry that that it may have been edited by someone else since you retrieved it". Obvs `manifest get|set` should be plumbing tools not for general use eventually in this case.

I actually think I lean towards the second (no checking of etags for get/set) as that's what our users assume it will do, and they seem happy with that idea from anecdotal evidence. This is prob outside the scope of this change, but seems relevant now that ETags are our canary fields.

samsalisbury:
Yeah in the case that there are multiple return values that need to be used outside the `if` block, we are left with long-scoped `err`s, and that is unavoidable. Keeping the others short-scoped still aids readability a lot for me. There is shadowing, but typically over only a single line of error-handling code, so easy to reason about.

samsalisbury:
LGTM, reviewed putbackJSON tests, makes sense. Happy for someone to merge, that coverage report seems broken so I'm inclined to ignore it. @nyarly @alrs

alrs:
Travis is happy, only Codecov is sad. I'll merge this.

PR #433, opened 2017-06-30
Implements locking Git state via Etags

nyarly:
The state gets an etag from the GitStateManager
(the revision of the current head).

The restul.Server will reject PUTs unless they
If-Match the current git-rev Etag.

The Etag gets set on the written state,
which the GSM checks against the head revision.

The upshot should be that it is impossible now to make an update
to the GDM unless
you are basing that update on the current revision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 with respect to its command line interface and HTTP interface.
 
 ## [Unreleased](//github.com/opentable/sous/compare/0.5.16...HEAD)
+
 ### Changed
+- Etags on /gdm are generated using the current head revision of the GDM git
+  repo. Updates are checked against this revision, which should mean we have a
+  closed loop against conflicting updates.
 - Makefile directives to build a Sous .deb and upload it to an Artifactory server.
 
 ### Fixed
 - Increased buffer size in shell to handle super-long tokens.
+
 
 ## [0.5.16](//github.com/opentable/sous/compare/0.5.15...0.5.16)
 - Fixed prefixed versions for update and deploy. At this point, git "package" tags should work everywhere they're used.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,15 @@ with respect to its command line interface and HTTP interface.
 ## [Unreleased](//github.com/opentable/sous/compare/0.5.16...HEAD)
 
 ### Changed
-- Etags on /gdm are generated using the current head revision of the GDM git
-  repo. Updates are checked against this revision, which should mean we have a
-  closed loop against conflicting updates.
-- Makefile directives to build a Sous .deb and upload it to an Artifactory server.
+- Developer: Makefile directives to build a Sous .deb and upload it to an Artifactory server.
+- Server: /gdm now sets Etag header based on current revision of the GDM. Clients attempting
+  to update with a stale Etag will have update rejected appropriately.
 
 ### Fixed
-- Increased buffer size in shell to handle super-long tokens.
-
+- Client: Increased buffer size in shell to handle super-long tokens.
+- Client: Conflicting updates to the same manifest now fail appropriately (e.g. during 'sous deploy').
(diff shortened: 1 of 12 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

ext/storage/git_state_manager_test.go
```diff
@@ -199,6 +154,33 @@ func setupManagers(t *testing.T) (clone *GitStateManager, remote *DiskStateManag
 	return gsm, dsm
 }
 
+func TestStateEtags(t *testing.T) {
+	gsm, _ := setupManagers(t)
+
+	s, err := gsm.ReadState()
+
+	if err != nil {
+		t.Errorf("Unexpected error when reading state")
```
samsalisbury: Always helpful to include the text of captured errors for greater debuggability.

lib/state.go
```diff
@@ -95,6 +96,32 @@ func NewState() *State {
 	}
 }
 
+// etag is a private field to ensure that we don't interfere with YAML storage,
+// hence the getter/setter
```
samsalisbury: Maybe this comment should be above `etag` field?

server/graph.go
```diff
@@ -28,5 +29,6 @@ func liveGDM(sr graph.StateReader) (*LiveGDM, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &LiveGDM{gdm.Deployments}, nil
+	etag, _ := state.GetEtag()
```
samsalisbury: Should prob not swallow this error, could be confusing.

server/handle_gdm.go
```diff
@@ -85,6 +87,10 @@ func (h *PUTGDMHandler) Exchange() (interface{}, int) {
 		return "Invalid GDM", http.StatusBadRequest
 	}
 
+	if _, got := h.Header["Etag"]; got {
+		state.SetEtag(h.Header.Get("Etag"))
```
samsalisbury: Could read the header once in the initialisation?

util/restful/router.go
```diff
@@ -170,23 +170,46 @@ func (mh *MetaHandler) writeHeaders(status int, w http.ResponseWriter, r *http.R
 	mh.statusHandler.HandleResponse(status, r, w, data)
 }
 
+var etagHeader = http.CanonicalHeaderKey("Etag")
+var contentTypeHeader = http.CanonicalHeaderKey("Content-Type")
+var contentLengthHeader = http.CanonicalHeaderKey("Content-Length")
```
samsalisbury: These are already canonical! :P could be constants, but doesn't make much difference I guess.

CHANGELOG.md
```diff
@@ -7,12 +7,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 with respect to its command line interface and HTTP interface.
 
 ## [Unreleased](//github.com/opentable/sous/compare/0.5.16...HEAD)
+
 ### Changed
+- Etags on /gdm are generated using the current head revision of the GDM git
+  repo. Updates are checked against this revision, which should mean we have a
+  closed loop against conflicting updates.
```
samsalisbury: I feel like the changelog should be written from a users' perspective, e.g. "Fixed: performance issues caused by conflicting concurrent writes to the GDM taking multiple passes to resolve." or something...

server/graph.go
samsalisbury: Or if that's what we want, then a comment helps.

server/handle_gdm.go
samsalisbury: Oh, multiple fields with the same name, ignore that.

samsalisbury:
I worry that this will generate a lot of conflicts, many of which are not relevant as they impact other manifests. It's definitely an improvement from where we were, but I think we'll need more re-tries for 'sous update' for now, until we narrow these Etags to a specific manifest...

PR #437, opened 2017-07-07
Default values for Startup fields # TECHOPS-2491

nyarly:


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

nyarly:
Here's the problem right now: we'd like to have defaults, which implies that omitted values should be replaced by the cluster default - that much works as is. Switching from pointers-to-strings to strings-with-defaults led to simpler code overall, I find.

However, the integration tests - specifically "succeed after failing" assumed that they could set a timeout and omit a healthcheck URI and that they'd boot successfully. The "succeed" service doesn't actually provide a healthcheck. Singularity sees the "" healthcheckURI and tries to ping it, but never gets a response.

I'm of the opinion that we should probably set up the Sous defaults to query a well-known startup URI - unless specifically opted out of; my plan then is to add a "no HTTP" field to the Startup config, and manage that behavior.

samsalisbury:
I think defaulting to a well known URI makes sense. I would be happy with "/health". Do you mean setting this default in Sous code rather than defs? If so, I'm curious why, but not against doing that.

Re "no HTTP" Startup config field, is that necessary in addition to having a different service type (Kind) and handling that? Or are you suggesting this as an alternative to Kinds? I quite like the idea of making Kinds less fine grained (e.g. just long-running or scheduled) and having specific fields in the manifest determine specific features. That feels more direct. In fact, maybe Kind is not even necessary. Merely having an eventual schedule field would be enough to know it's a scheduled task.

I haven't reviewed this code, but I like where you're going with this.

PR #441, opened 2017-07-21
Adding 'plumbing normalizegdm'

nyarly:
This just reads the SOUS_STATE_LOCATION into memory and then writes it back.

Because storing is a stable operation, but the format is flexible enough to
represent the same configuration many ways, it's easy to produce an intended
configuration in a way that will be stored in a different textual form. The
result is that spurious differences cause a conflict in the GitStateManager and
prevent deploys.

It's possible that running `sous plumbing normalizegdm` after any manual
changes to the Git backed GDM will alleviate this problem entirely.

### DIFF
--- a/cli/sous_plumbing_normalizegdm.go
+++ b/cli/sous_plumbing_normalizegdm.go
@@ -1,0 +1,44 @@
+package cli
+
+import (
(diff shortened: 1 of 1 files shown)


sous-machine-account:
LGTM

</p>

PR #443, opened 2017-07-29
Begins the process of transition from storage to server

nyarly:
In master, the command line app switches modes (mostly) by choosing a
StateManager. The client mode uses a HTTPStateManager to send whole-GDM
updates. The workstation mode uses the GitStateManager to update a local git
repo.

This mostly works, but there are several scenarios where it's insufficient. In
those cases, the app needs to determine separately whether it is a client or
workstation and do different things. There are also commands that only work in
certain modes as a result.

This PR begins a different approach: the app always has an HTTP interface to
the server - in workstation mode that server is run locally, and the interface
communicates with it in memory via a channel.

The result should be that we get uniformity across our commands,
much simpler code in many cases, and smaller network requests.

There's still work to do here, which might make for a good place for new
contributors to pick up. The important thing is that the pattern is
established, including testing.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 1579 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

PR #450, opened 2017-08-22
Fixes a DI context issue between CLI and server injection.

nyarly:
There was a problem with how DI contexts were being passed around: the command object responsible for `sous server` registered providers with its DI, but the actual server component didn't get that DI - it got one from further up the chain of contexts, which meant that important machinery wasn't available - so the server would blow up when certain requests were made of it.

Fixing this involved pulling a fair amount of the plumbing out and packing it back in.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime/debug"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/opentable/sous/graph"
 	"github.com/opentable/sous/util/cmdr"
 	"github.com/opentable/sous/util/logging"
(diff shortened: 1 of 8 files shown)


sous-machine-account:
LGTM

</p>

cli/cli.go
```diff
@@ -91,11 +93,26 @@ func BuildCLIGraph(cli *CLI, root *Sous, in io.Reader, out, err io.Writer) *grap
...
 
 	return cli.SousGraph
 }
 
+func (cli *CLI) scopedGraph(cmd, under cmdr.Command) *graph.SousGraph {
+	if g, ok := cli.scopedGraphs[cmd]; ok {
+		return g
+	}
+
+	parent := cli.scopedGraphs[under]
```
SegFaultAX: Does this have the potential to cause a panic?

cli/cli.go
```diff
@@ -91,11 +93,26 @@ func BuildCLIGraph(cli *CLI, root *Sous, in io.Reader, out, err io.Writer) *grap
...
+
+	g := &graph.SousGraph{Psyringe: parent.Clone()}
+	if r, ok := cmd.(Registrant); ok {
```
SegFaultAX: I haven't seen this form yet `cmd.(Registrant)`, what's the significance of it?

cli/cli.go
nyarly: It does, but under normal operations it shouldn't - only if you pass in the non-parent of a missing command. (At one point a test caused panics.) If that happens, there's not much to do *but* panic.

cli/cli.go
nyarly: It's a type assertion: "cmd fulfills the Registrant interface" in this case. The two-arg result works like map access: ok will be true if the assertion holds. (There's a single-arg form that panics if the assertion fails.)

SegFaultAX: I see! Interesting syntax for that. Thanks!

cli/cli.go
SegFaultAX: Cool, thanks.

PR #454, opened 2017-08-28
Fixes `manifest set` JSON putback.

nyarly:
Null fields were being put into the proposed document, then removed because
they weren't being marked as found (by deletion from the base map.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ with respect to its command line interface and HTTP interface.
 
 ### Fixed
 - Client: certain commands were missing DI for a particular value. These are fixed, and tests added for the omission.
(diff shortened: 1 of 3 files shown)


sous-machine-account:
LGTM

</p>

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ with respect to its command line interface and HTTP interface.
 
 ### Fixed
 - Client: certain commands were missing DI for a particular value. These are fixed, and tests added for the omission.
+- Client: new values for optional fields no longer elided on `manifest set`
```
samsalisbury: Should this read 'null values' rather than 'new values'? (null not nil because yaml) otherwise lgtm

PR #457, opened 2017-09-01
Logging and tests regard polling and flavors.

nyarly:


config/deploy_filter.go
```diff
@@ -54,8 +48,8 @@ func (f *DeployFilterFlags) BuildFilter(parseSL func(string) (sous.SourceLocatio
 		if err != nil {
 			return nil, errors.Wrap(err, "error parsing -source flag")
 		}
-		rf.Repo = sl.Repo
-		rf.Offset.Match = &sl.Dir
+		rf.Repo = sous.NewResolveFieldMatcher(sl.Repo)
+		rf.Offset = sous.NewResolveFieldMatcher(sl.Dir)
```
richardhopton: You are losing the implicit "match" and overriding the value you set before https://github.com/opentable/sous/pull/457/files#diff-c4b01d2d5807d8fd7954db4a2f55afafR30

sous-machine-account:
LGTM

</p>

config/deploy_filter.go
nyarly: This is true. I'm considering unexporting Match, actually.

PR #464, opened 2017-09-19
Flavor wasn't being recovered into the ADS.

nyarly:
Although flavor data was being recorded in the deploy by Sous, when
rectifaction collected the actual deployment set, the flavor wasn't collected
out of the metadata, meaning that the ADS appeared to only include unflavored
deploys. Rectification therefore got stuck in a permanent create/delete pair,
rather than a single matching unchanged deployment.

One consequence is that `sous deploy` would never return successful -
because the Sous server was in fact reporting that deploy was ongoing.


sous-machine-account:
LGTM

</p>

PR #468, opened 2017-09-22
Commands were using logging.LogSet directly TECHOPS-3149 TECHOPS-3152

nyarly:
Everything should be using the LogSink interface, not LogSet

n.b. that there are many uses of the global logging.Log LogSet - which
means that e.g. the logwrappers can't be removed yet - the correct
behavior, where possible, should be to replace those uses with injected
LogSinks.

TECHOPS-3149 TECHOPS-3152


sous-machine-account:
LGTM

</p>

PR #469, opened 2017-09-25
Improving Log output

nyarly:
Several places where logging was badly formatted or incomplete have been cleaned up.

There's certainly more work to do, but of note: the EchoSelector has been moved to the docker package,
where it belongs, and uses a specialized message to report the selection of build strategy to the user.

Kafka and graphite report on their configuration at boot time, so that we can be sure of what's going on.

Graphite errors are no longer silent (there was a bug there, with a regression test.)


nyarly:
I'm shaky on how the tests should work here. Is it worth wiring up spies for each of these messages?

sous-machine-account:
LGTM

</p>

PR #470, opened 2017-09-25
Init flavor fixes

richardhopton:


sous-machine-account:
LGTM

</p>

cli/sous_init.go
```diff
@@ -74,7 +77,16 @@ func (si *SousInit) Execute(args []string) cmdr.Result {
 	flavor := si.DeployFilterFlags.Flavor
 
 	if flavor != "" {
-		for _, d := range m.Deployments {
+		for c, d := range m.Deployments {
+			if d.Env == nil {
+				d = sous.DeploySpec{
```
nyarly: Why not 
```
if d.Env == nil {
  d.Env = map[string]string{ OT_ENV_FLAVOR: flavor }
} else {
  d.Env[OT_ENV_FLAVOR] = flavor
}
```
?

richardhopton: I tried that first but it didn't write Env to Yaml. I also tried setting Env on DeployConfig property too. This was the only way I could get it to work :-(

PR #472, opened 2017-09-27
Fixes DB DDL related race condition

nyarly:
* Wraps DDL modifications in a Once
* Reports on errors from the DDL change
* "scoops" out NameCaches instead of reusing the constructor

### DIFF
--- a/ext/docker/image_mapping.go
+++ b/ext/docker/image_mapping.go
@@ -13,6 +13,7 @@ import (
 	"text/tabwriter"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/reference"
 	"github.com/pkg/errors"
@@ -33,6 +34,7 @@ type (
 		DB                 *sql.DB
 		DockerRegistryHost string
 		Log                logging.LogSink
+		groomOnce          sync.Once
 	}
 
 	imageName string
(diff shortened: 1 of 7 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

ext/docker/image_mapping.go
```diff
@@ -458,37 +459,41 @@ func semverEqual(a, b string) (bool, error) {
...
-		err = nil
-		repos := nc.captureRepos(db)
-
-		nc.clobber(db)
-
-		for _, cmd := range schema {
-			if err := sqlExec(db, cmd); err != nil {
-				return errors.Wrapf(err, "groom DB/create: %v", db)
+	nc.groomOnce.Do(func() {
+		queryErr := db.QueryRow("select value from _database_metadata_ where name = 'fingerprint';").Scan(&tgp)
```
richardhopton: I've found this check rarely fixes race conditions.  A better option is to try the insert and if it fails then look to see if the document is there and respond success if so.

richardhopton:
Did you mean to squash your commit?

ext/docker/image_mapping.go
nyarly: That's not what this check is for. It confirms that the database-as-is has the right schema; if it does we don't have to fix it. The fingerprint is recorded as the last step, and if it's generated as a digest of the DDL code.

Also: insert-then-select narrows the window but still doesn't close the race condition problem. What you want is upsert, which is a fairly late addition in SQL (e.g. Postgres 9.4, I think, allows for it?) We fake it here with some unique indexes and some insert...or update lines.

nyarly:
I'm okay without squashing.

PR #484, opened 2017-10-09
Autodetect offset for deploy, manifest get & set 

samsalisbury:
Ideally this needs an integration test but not got that far yet. All existing tests pass, and I have manually verified it so opening PR for discussion.

Most of the commits here just add or tidy up unit tests. The crux of this change is https://github.com/opentable/sous/commit/21bf57ec2edf5ce73474a228a24db79ccef53b2c

### DIFF
--- a/graph/targetmanifest.go
+++ b/graph/targetmanifest.go
@@ -22,12 +22,18 @@ func newRefinedResolveFilter(f *sous.ResolveFilter, discovered *SourceContextDis
 		if c.PrimaryRemoteURL != "" {
 			rrf.Repo = sous.NewResolveFieldMatcher(c.PrimaryRemoteURL)
 		}
-		if rrf.Offset.All() && c.OffsetDir != "" {
+		if rrf.Offset.All() || *rrf.Offset.Match == "" {
 			rrf.Offset = sous.NewResolveFieldMatcher(c.OffsetDir)
 		}
 		if f.Tag.All() && discovered.SourceContext != nil && discovered.TagVersion() != "" {
 			rrf.SetTag(discovered.TagVersion())
 		}
+	} else {
+		var offset string
+		if rrf.Offset.Match != nil {
+			offset = *rrf.Offset.Match
+		}
+		rrf.Offset = sous.NewResolveFieldMatcher(offset)
 	}
 
 	if rrf.Repo.All() {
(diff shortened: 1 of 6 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

samsalisbury:
Happy for this to be merged prior to adding integration test, as it fixes a bug and improves tests, but ticket won't be closed til integration test done @nyarly

graph/targetmanifest.go
```diff
@@ -22,12 +22,18 @@ func newRefinedResolveFilter(f *sous.ResolveFilter, discovered *SourceContextDis
 		if c.PrimaryRemoteURL != "" {
 			rrf.Repo = sous.NewResolveFieldMatcher(c.PrimaryRemoteURL)
 		}
-		if rrf.Offset.All() && c.OffsetDir != "" {
+		if rrf.Offset.All() || *rrf.Offset.Match == "" {
```
nyarly: If Offset is nil, won't this segfault?

nyarly: (except .All() tests for nil) - the ResolveFieldMatchers have a match() method that does exactly this test, though.

Maybe there's a larger conversation to have about their interface, since as designed they're obviously not clear about their usage.

graph/targetmanifest.go
```diff
@@ -22,12 +22,18 @@ func newRefinedResolveFilter(f *sous.ResolveFilter, discovered *SourceContextDis
...
 		}
 		if f.Tag.All() && discovered.SourceContext != nil && discovered.TagVersion() != "" {
 			rrf.SetTag(discovered.TagVersion())
 		}
+	} else {
+		var offset string
+		if rrf.Offset.Match != nil {
+			offset = *rrf.Offset.Match
+		}
+		rrf.Offset = sous.NewResolveFieldMatcher(offset)
```
nyarly: (not about this line)

I don't see any tests that touch this behavior directly.

I also don't see how this change updates `sous manifest get` - it should impact `sous deploy` and `sous init`, but `sous manifest get` uses a simple ResolveFilter.

samsalisbury: Hi, `sous manifest get` uses `TargetManifestID` which relies on `RefinedResolveFilter`. Although it is additionally composed of simple `ResolveFilter` which is used for reporting failure to find manifest. This is probably wrong, `sous manifest get` should use the `RefinedResolveFilter` instead.

graph/targetmanifest.go
samsalisbury: Ah yes, if that method was exported we could use it here.. Any reason not to export it?

PR #486, opened 2017-10-11
Log entries for deployment diffs

nyarly:
Includes a shift in architecture around how DeployableChans work:

The differs produce DeployableChans, on which Pipeline() gets called to add
processors. ResolveNames is now a specialization of Pipeline().

Logging is handled by a pipeline processor.

Also included: the logging.AssertMessageFormat helper emits propsed OTLs to $TMPDIR.
They're not ready for immediate inclusion, but they sure save a lot of work.

### DIFF
--- a/lib/nameresolver.go
+++ b/lib/nameresolver.go
@@ -98,6 +98,12 @@ func resolveName(r Registry, d *Deployable) (*Deployable, *DiffResolution) {
 }
 
 func resolvePair(r Registry, depPair *DeployablePair) (*DeployablePair, *DiffResolution) {
+	// In every case where we resolve a pair of names, the "prior" name is not a target -
+	// in other words, it's nice to know, but not knowing it doesn't prevent progress of the resolution process.
+	// We therefore can discard errors here in a similar way that we discard them in maybeResolveSingle.
+	// However, the larger view is that we might want to undo any change we make, and not having the artifact name
+	// of the *current* deploy would mean that we're performing an update we don't know how (in this moment) to reverse.
+	// [ this issue is captured in SOUS-727 for tracking purposes ]
 	prior, _ := resolveName(r, depPair.Prior)
 	post, err := resolveName(r, depPair.Post)
 
@@ -114,25 +120,23 @@ func guardImage(r Registry, d *Deployment) (*BuildArtifact, error) {
 		return nil, &MissingImageNameError{err}
 	}
 	for _, q := range art.Qualities {
-		if q.Kind == "advisory" {
-			if q.Name == "" {
-				continue
+		if q.Kind != "advisory" || q.Name == "" {
+			continue
+		}
+		advisoryIsValid := false
+		var allowedAdvisories []string
+		if d.Cluster == nil {
+			return nil, fmt.Errorf("nil cluster on deployment %q", d)
+		}
+		allowedAdvisories = d.Cluster.AllowedAdvisories
+		for _, aa := range allowedAdvisories {
+			if aa == q.Name {
+				advisoryIsValid = true
+				break
 			}
-			advisoryIsValid := false
-			var allowedAdvisories []string
-			if d.Cluster == nil {
-				return nil, fmt.Errorf("nil cluster on deployment %q", d)
-			}
-			allowedAdvisories = d.Cluster.AllowedAdvisories
-			for _, aa := range allowedAdvisories {
-				if aa == q.Name {
-					advisoryIsValid = true
-					break
-				}
-			}
-			if !advisoryIsValid {
-				return nil, &UnacceptableAdvisory{q, &d.SourceID}
-			}
+		}
+		if !advisoryIsValid {
+			return nil, &UnacceptableAdvisory{q, &d.SourceID}
 		}
 	}
 	return art, err
--- a/lib/nameresolver.go
+++ b/lib/nameresolver.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/opentable/sous/util/logging"
 	"github.com/pkg/errors"
 )
@@ -25,7 +24,6 @@ func (names *nameResolver) Start(dp *DeployablePair) (*DeployablePair, *DiffReso
 	logging.Log.Vomit.Printf("Deployment processed, needs artifact: %#v", dep)
 
 	da, err := resolveName(names.registry, dep)
-	spew.Printf("Start resolve name %v %v", dep.Deployment.SourceID, err)
 	if err != nil {
 		logging.Log.Info.Printf("Unable to create new deployment %q: %s", dep.ID(), err)
 		logging.Log.Debug.Printf("Failed create deployment %q: % #v", dep.ID(), dep)
@@ -112,7 +110,6 @@ func guardImage(r Registry, d *Deployment) (*BuildArtifact, error) {
 		return nil, nil
 	}
 	art, err := r.GetArtifact(d.SourceID)
-	spew.Dump("resolving name", d.SourceID, err)
 	if err != nil {
 		return nil, &MissingImageNameError{err}
 	}
--- a/lib/nameresolver.go
+++ b/lib/nameresolver.go
@@ -14,10 +14,10 @@ type nameResolver struct {
 }
 
 // ResolveNames resolves diffs.
-func (d *DeployableChans) ResolveNames(r Registry) *DeployableChans {
+func (d *DeployableChans) ResolveNames(ctx context.Context, r Registry) *DeployableChans {
 	names := &nameResolver{registry: r}
 
-	return d.Pipeline(context.Background(), names)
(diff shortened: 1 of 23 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 1052 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

lib/loggingprocessor_test.go
```diff
@@ -0,0 +1,149 @@
...
+		"sous-prior-clustername":                "test-cluster",
+		"sous-prior-env":                        "{}",
+		"sous-prior-flavor":                     "",
+		"sous-prior-kind":                       "http-service",
+		"sous-prior-metadata":                   "{}",
+		"sous-prior-numinstances":               1,
+		"sous-prior-offset":                     "",
+		"sous-prior-owners":                     "",
+		"sous-prior-repo":                       "github.com/opentable/example",
+		"sous-prior-resources":                  "{\"cpus\":\"0.100\",\"memory\":\"356\",\"ports\":\"2\"}",
```
samsalisbury: I wonder if there would be value in expanding resources, could be a useful filter? At the moment, resources are dynamic-ish, but in reality it's always these 3 fields for now so should be safe for an OTL. Relatedly, I really wish `memory` was called `memoryMB` since I would usually assume bytes if no unit given.

lib/loggingprocessor_test.go
```diff
@@ -0,0 +1,149 @@
...
+		"sous-post-startup-skipcheck":          true,
+		"sous-post-startup-timeout":            0,
+		"sous-post-status":                     "DeployStatusActive",
+		"sous-post-tag":                        "0.0.1",
+		"sous-post-volumes":                    "[]",
+	}
+
+	msg.pair.name = msg.pair.Prior.ID()
+
+	logging.AssertMessageFields(t, msg, fixedFields, fields)
```
samsalisbury: Feels like `logging.AssertMessageFields` should be defined nearer the tests to make it quicker to see exactly what's being tested.

samsalisbury: AssertMessageFields has `variableFields` param first then `fixedFields` slightly confusing but I can see the issue, `fields` should be called `fixedFields` and vice versa. https://github.com/opentable/sous/pull/486/files#diff-e55bece6cce54370d6fbf305664a77a4R149

lib/loggingprocessor_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	fixedFields := []string{
+		"@timestamp",
+		"call-stack-file",
+		"call-stack-line-number",
+		"call-stack-function",
+		"thread-name",
+	}
+
+	fields := map[string]interface{}{
```
samsalisbury: Rename to `fixedFields`? See comment below.

lib/loggingprocessor_test.go
```diff
@@ -0,0 +1,149 @@
...
+	msg := &deployableMessage{
+		pair: &DeployablePair{
+			Diffs: Differences{},
+			Prior: deployableFixture(""),
+			Post:  deployableFixture(""),
+		},
+		callerInfo: logging.GetCallerInfo(),
+	}
+
+	fixedFields := []string{
```
samsalisbury: Rename `fields` or `variableFields`? See comment below.

samsalisbury:
I'm still getting intermittent integration test failures, roughly 75% of the time, errors like this:

```
--- FAIL: TestMissingImage (6.18s)                                       
        Error Trace:    integration_test.go:381
        Error:          An error is expected but got nil.
        Messages:       should report 'missing image'
        integration_test.go:44: sous.DeployStates{mu:(*sync.RWMutex)(0xc4
```

lib/testfixtures.go
```diff
@@ -0,0 +1,115 @@
...
+			BuildArtifact: &BuildArtifact{
+				Type:      "docker",
+				Name:      "dockerhub.io/example:0.0.1",
+				Qualities: []Quality{},
+			},
+		}
+	}
+}
+
+func manifestFixture(name string) *Manifest {
```
samsalisbury: This func appears unused according to the coverage report: https://codecov.io/gh/opentable/sous/compare/20990939a72daa962a2baa7d8b45ae2128951a5f...f743769a6ce4bd90995bffe09b0578f67d879635/diff#D2-61 is this intentional or should we delete it?

lib/nameresolver.go
```diff
@@ -0,0 +1,139 @@
...
+			DeploymentID: d.ID(),
+			Error:        &ErrorWrapper{error: err},
+		}
+	}
+	d.BuildArtifact = art
+	return d, nil
+}
+
+func resolvePair(r Registry, depPair *DeployablePair) (*DeployablePair, *DiffResolution) {
+	prior, _ := resolveName(r, depPair.Prior)
```
samsalisbury: I always get nervous when throwing errors away but I can see that the error here is operationally irrelevant. However, if an error is returned, I would rather at least modify `prior` to indicate that fetching it resulted in an error, to make any logs printing it out more explicit, e.g. `fmt.Sprintf("failed to fetch image name because %s", err)` I don't think it's worth logging by itself though. If modifying `prior` would have a negative impact, then at least a comment explaining why we are ignoring the error would be better than nothing.

lib/nameresolver.go
```diff
@@ -0,0 +1,139 @@
...
+	if d.NumInstances == 0 {
+		logging.Log.Info.Printf("Deployment %q has 0 instances, skipping artifact check.", d.ID())
+		return nil, nil
+	}
+	art, err := r.GetArtifact(d.SourceID)
+	if err != nil {
+		return nil, &MissingImageNameError{err}
+	}
+	for _, q := range art.Qualities {
+		if q.Kind == "advisory" {
```
samsalisbury: Maybe avoid indentation with `if q.Kind != "advisory" { continue }` ?

lib/pipeline.go
```diff
@@ -0,0 +1,99 @@
...
+
+	process := func(from, to chan *DeployablePair, doProc func(dp *DeployablePair) (*DeployablePair, *DiffResolution)) {
+		defer close(to)
+		defer wg.Done()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case dp, ok := <-from:
+				if ok {
```
samsalisbury: Maybe avoid indentation with `if !ok { return }` ?

lib/testfixtures.go
```diff
@@ -0,0 +1,115 @@
+package sous
+
+import "github.com/samsalisbury/semv"
+
+func deployConfigFixture(name string) DeployConfig {
+	switch name {
```
samsalisbury: Guess this lost its meaning during refactor? Can lose the switch now...

lib/testfixtures.go
```diff
@@ -0,0 +1,115 @@
...
+			Volumes:      Volumes{},
+			Startup: Startup{
+				SkipCheck: true,
+			},
+		}
+	}
+}
+
+func deploymentFixture(name string) *Deployment {
+	switch name {
```
samsalisbury: Hmm, maybe I'm missing something this could just be `return ...` ?

lib/testfixtures.go
```diff
@@ -0,0 +1,115 @@
...
+			Flavor: "",
+			Owners: OwnerSet{},
+			Kind:   ManifestKindService,
+			//Cluster *Cluster,
+		}
+	}
+}
+
+func deployableFixture(name string) *Deployable {
+	switch name {
```
samsalisbury: Same here, guess there was originally a reason for this switch?

util/logging/testsupport.go
```diff
@@ -157,4 +160,55 @@ func AssertMessageFields(t *testing.T, msg eachFielder, variableFields []string,
 	}
 
 	assert.Equal(t, fixedFields, actualFields)
+
+	if _, hasOTL := actualFields["@loglov3-otl"]; !t.Failed() && hasOTL {
```
samsalisbury: I'm not comfortable with an innocuous looking `Assert` method writing local files, although it's cool to have these OTLS. Maybe just split this into its own method or external tool for calling manually?

lib/testfixtures.go
nyarly: The intention is that future fixtures be added here with names. There are many places where we're hand rolling a deployConfiguration, the contents of which aren't really significant. I left the otherwise purposeless switch to communicate that idea.

samsalisbury: OK cool, maybe a comment to that effect would help for future readers? I would just remove the cruft otherwise next time I come to it by which time I will have forgotten this conversation.

lib/testfixtures.go
nyarly: I'd rather say: this code is never used in production - again, it exists to suggest its usage: it'll already exist never time we need to gin up a dummy manifest.

PR #488, opened 2017-10-19
Poll full resolution

samsalisbury:
The crucial part of this PR is the rather verbose `computeStatus` function. Because we have different rules to handle `ResolveComplete` vs `ResolveFailed` simple comparison with `ResolveTERMINALS` isn't very helpful for this method. The test `TestStatusPoller_updateState` reveals the new spec for computing this state.

Other changes include making `ResolveFailed` greater than `ResolveComplete` since when computing `maxStatus` we want failed states to take precedence over successful ones (although this may not now be important with the latest implementation of `computeStatus`.

I was using the race detector to check for anything fishy that might be affecting these tests, as a result this PR includes a couple of fixes for race conditions which were probably not relevant to the eventual solution but nonetheless were worth addressing.

`ResolveState` is now in its own file.

### DIFF
--- a/lib/status_poller.go
+++ b/lib/status_poller.go
@@ -321,9 +321,6 @@ func (sub *subPoller) start(rs chan pollResult, done chan struct{}) {
 	ticker := time.NewTicker(PollTimeout)
 	defer ticker.Stop()
 	for {
-		//if sub.lastCycle && pollResult.stat >= ResolveTERMINALS {
-		//	return
-		//}
 		select {
 		case <-ticker.C:
 			latest := sub.pollOnce()
--- a/lib/status_poller.go
+++ b/lib/status_poller.go
@@ -314,34 +314,39 @@ func (sp *StatusPoller) finished() bool {
 // c.f. pollOnce.
 func (sub *subPoller) start(rs chan pollResult, done chan struct{}) {
 	rs <- pollResult{url: sub.URL, stat: ResolveNotPolled}
-	stat, err := sub.pollOnce()
-	rs <- pollResult{url: sub.URL, stat: stat, err: err}
+	pollResult := sub.pollOnce()
+	rs <- pollResult
 	ticker := time.NewTicker(time.Second / 2)
 	defer ticker.Stop()
 	for {
-		if stat >= ResolveTERMINALS {
+		if pollResult.stat >= ResolveTERMINALS {
 			return
 		}
 		select {
 		case <-ticker.C:
-			stat, err = sub.pollOnce()
-			rs <- pollResult{url: sub.URL, stat: stat, err: err}
+			rs <- sub.pollOnce()
 		case <-done:
 			return
 		}
 	}
 }
 
-func (sub *subPoller) pollOnce() (ResolveState, error) {
+func (sub *subPoller) result(rs ResolveState, data *statusData, err error) pollResult {
+	resolveID := data.InProgress.Started.String()
+	return pollResult{url: sub.URL, stat: rs, resolveID: resolveID, err: err}
+}
+
+func (sub *subPoller) pollOnce() pollResult {
 	data := &statusData{}
 	if _, err := sub.Retrieve("./status", nil, data, sub.User.HTTPHeaders()); err != nil {
 		logging.Log.Debugf("%s: error on GET /status: %s", sub.ClusterName, errors.Cause(err))
 		logging.Log.Vomitf("%s: %T %+v", sub.ClusterName, errors.Cause(err), err)
 		sub.httpErrorCount++
 		if sub.httpErrorCount > 10 {
-			return ResolveFailed, fmt.Errorf("more than 10 HTTP errors, giving up; latest error: %s", err)
+			return sub.result(ResolveFailed, data,
+				fmt.Errorf("more than 10 HTTP errors, giving up; latest error: %s", err))
 		}
-		return ResolveErredHTTP, err
+		return sub.result(ResolveErredHTTP, data, err)
 	}
 	sub.httpErrorCount = 0
 
@@ -359,10 +364,11 @@ func (sub *subPoller) pollOnce() (ResolveState, error) {
 	if currentState == ResolveNotStarted ||
 		currentState == ResolveNotVersion ||
 		currentState == ResolvePendingRequest {
-		return sub.computeState(sub.stateFeatures("completed", data.Completed))
+		state, err := sub.computeState(sub.stateFeatures("completed", data.Completed))
+		return sub.result(state, data, err)
 	}
 
-	return currentState, err
+	return sub.result(currentState, data, err)
 }
 
 func (sub *subPoller) stateFeatures(kind string, rezState *ResolveStatus) (*Deployment, *DiffResolution) {
--- a/lib/status_poller.go
+++ b/lib/status_poller.go
@@ -426,18 +426,18 @@ func (sub *subPoller) computeState(srvIntent *Deployment, current *DiffResolutio
 		// error for operator action, and consider this subpoller done as failed.
 		logging.Log.Vomitf("%#v", current)
 		logging.Log.Vomitf("%#v", current.Error)
-		//subject := ""
-		//if sub.locationFilter == nil {
-		//	subject = "<no filter defined>"
-		//} else {
-		//	sourceLocation, ok := sub.locationFilter.SourceLocation()
-		//	if ok {
-		//		subject = sourceLocation.String()
-		//	} else {
-		//		subject = sub.locationFilter.String()
-		//	}
-		//}
-		//logging.Log.Warn.Printf("Deployment of %s to %s failed: %s", subject, sub.ClusterName, current.Error.String)
+		subject := ""
+		if sub.locationFilter == nil {
+			subject = "<no filter defined>"
+		} else {
+			sourceLocation, ok := sub.locationFilter.SourceLocation()
+			if ok {
+				subject = sourceLocation.String()
+			} else {
+				subject = sub.locationFilter.String()
+			}
+		}
+		logging.Log.Warn.Printf("Deployment of %s to %s failed: %s", subject, sub.ClusterName, current.Error.String)
 		return ResolveFailed, current.Error
 	}
 
--- a/lib/status_poller.go
+++ b/lib/status_poller.go
@@ -283,34 +283,61 @@ func maxStatus(a, b ResolveState) ResolveState {
 }
 
 func (sp *StatusPoller) updateStatus() {
-	max := ResolveNotPolled
-	totalMax := ResolveFailed
-	totalMin := ResolveNotPolled
-	allLastCycle := true
+	sp.status = sp.computeStatus()
+}
+
+func (sp *StatusPoller) computeStatus() ResolveState {
+	firstCycleMax := ResolveState(0)
+	firstCycleMin := ResolveMAX
+	lastCycleMax := ResolveState(0)
+	lastCycleMin := ResolveMAX
 	for u, s := range sp.statePerCluster {
 		logging.Log.Vomitf("Current state from %s: %s", u, s)
-		//max = maxStatus(max, s.LastResult.stat)
-		if !s.LastCycle {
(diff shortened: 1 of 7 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 601 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

lib/status_poller.go
```diff
@@ -309,36 +314,42 @@ func (sp *StatusPoller) finished() bool {
 
 // start issues a new /status request every half second, reporting the state as computed.
 // c.f. pollOnce.
-func (sub *subPoller) start(rs chan statPair, done chan struct{}) {
-	rs <- statPair{url: sub.URL, stat: ResolveNotPolled}
-	stat := sub.pollOnce()
-	rs <- statPair{url: sub.URL, stat: stat}
-	ticker := time.NewTicker(time.Second / 2)
+func (sub *subPoller) start(rs chan pollResult, done chan struct{}) {
```
nyarly: done channel should probably be replaced with a Context (that descends from top level poller's context)

lib/status_poller.go
```diff
@@ -276,30 +209,102 @@ func (sp *StatusPoller) poll(subs []*subPoller) ResolveState {
...
 	}
-	sp.status = max
+
+	// All completed successfully in first cycle.
+	if firstCycleMax == ResolveComplete && firstCycleMin == ResolveComplete {
+		return ResolveComplete
+	}
+
+	// At least one resolution in second cycle has failed.
+	if lastCycleMax == ResolveFailed {
```
nyarly: I *think* if lastCycleMax == ResolveNotVersion or ResolveNotStarted, that should also report an early exit - means that the version we intended to deploy isn't accepted in the GDM (and so won't ever deploy)

lib/status_poller.go
```diff
@@ -374,13 +386,13 @@ func (sub *subPoller) stateFeatures(kind string, rezState *ResolveStatus) (*Depl
 // computeState takes the servers intended deployment, and the stable and
 // current DiffResolutions and computes the state of resolution for the
 // deployment based on that data.
-func (sub *subPoller) computeState(srvIntent *Deployment, current *DiffResolution) ResolveState {
+func (sub *subPoller) computeState(srvIntent *Deployment, current *DiffResolution) (ResolveState, error) {
```
nyarly: I'm not convinced adding an error return value here is a good idea. (see below)

lib/status_poller.go
```diff
@@ -408,7 +420,7 @@ func (sub *subPoller) computeState(srvIntent *Deployment, current *DiffResolutio
 		// transition and be ready to receive a new Deploy)
 		if IsTransientResolveError(current.Error) {
 			logging.Log.Debugf("%s: received resolver error %s, retrying", sub.ClusterName, current.Error)
-			return ResolveErredRez
+			return ResolveErredRez, current.Error
```
nyarly: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully#handle-errors-once

We logged the error - it's transient and we're retrying. Error is handled. Why return it?

lib/status_poller.go
```diff
@@ -426,22 +438,22 @@ func (sub *subPoller) computeState(srvIntent *Deployment, current *DiffResolutio
 			}
 		}
 		logging.Log.Warn.Printf("Deployment of %s to %s failed: %s", subject, sub.ClusterName, current.Error.String)
-		return ResolveFailed
+		return ResolveFailed, current.Error
```
nyarly: Likewise, we logged the error, it can't be fixed, and we're returning a failure code. Why return the error as well?

lib/status_poller.go
```diff
@@ -426,22 +438,22 @@ func (sub *subPoller) computeState(srvIntent *Deployment, current *DiffResolutio
...
+		return ResolveComplete, nil
 	}
 
 	if current.Desc == ComingDiff {
-		return ResolveTasksStarting
+		return ResolveTasksStarting, nil
 	}
 
-	return ResolveInProgress
+	return ResolveInProgress, nil
```
nyarly: All the other values of the error return are nil. So... why return the error?

I also note: the error gets packed into the pollStatus, and then ignored, which I think is the result of the logic in this feedback.

lib/status_poller.go
```diff
@@ -55,94 +66,16 @@ type (
...
 	}
 
-	// A ResolveState reflects the state of the Sous clusters in regard to
-	// resolving a particular SourceID.
-	ResolveState int
-
-	statPair struct {
-		url  string
-		stat ResolveState
+	pollResult struct {
```
nyarly: Much better name than `statPair`

lib/status_poller_test.go
```diff
@@ -102,52 +104,145 @@ func TestStatusPoller_updateState(t *testing.T) {
...
-	assertStatus := func(status ResolveState) {
-		assert.Equal(sp.status, status, "StatusPoller total state was %s, expected %s", sp.status, status)
+	assertStatus := func(expected ResolveState) {
+		actual := sp.status
+		if actual != expected {
+			t.Errorf("got %s; want %s (after %d results):\n%s",
+				actual, expected, len(resultsSoFar), resultsSoFarStr())
+		}
+	}
+	result := func(clusterName, resolveID string, status ResolveState) {
```
nyarly: I really like this refinement. It makes the whole story clearer.

lib/status_poller_test.go
```diff
@@ -102,52 +104,145 @@ func TestStatusPoller_updateState(t *testing.T) {
...
+	result("two", first, ResolveComplete)
 	assertStatus(ResolveComplete)
 
-	sp.pollChans["one"] = ResolveComplete
-	sp.pollChans["two"] = ResolveFailed
+	// One moves to ResolveFailed in first cycle, overall ResolveInProgress
+	result("one", first, ResolveFailed)
+	assertStatus(ResolveInProgress)
 
-	assert.True(sp.finished(), "StatusPoller reported NOT finished: %s", sp.status)
```
nyarly: Why did you remove the "finished" tests? IMO it's important that the poller actually *stop* in the right states.

lib/status_poller_test.go
```diff
@@ -102,52 +104,145 @@ func TestStatusPoller_updateState(t *testing.T) {
...
+	var gdmJSON, serversJSON, statusJSON, statusJSON2 []byte
+
+	statusCalled := false
+	handleMutex := sync.Mutex{}
 
 	h := func(rw http.ResponseWriter, r *http.Request) {
+		// For testing purposes, we want to ensure we handle
+		// responses one at a time since statusCalled must
+		// be false on the first call and true on the second.
+		// The race detector picked up this issue.
```
nyarly: :+1:

util/logging/applicationid.go
```diff
@@ -48,9 +49,13 @@ func collectAppID(version semv.Version, env map[string]string) *applicationID {
 	return &id
 }
 
+// seq returns the next sequence number in a concurrency-safe manner.
+func (id *applicationID) seq() uint64 {
+	return atomic.AddUint64(&id.sequenceNumber, 1)
```
nyarly: I was just thinking that there are logging-related mutex concerns. At the time I'd thought (mistakenly) "when logging becomes multithreaded..." but of course it is already.

We might consider having the logging machinery in its own gorountine and sending log messages over a channel to it. Then a lot of the locking stuff becomes superfluous. If we don't, I have the nagging suspicion that there are other places we need to lock.

lib/status_poller.go
samsalisbury: Hmm, trying to get my head around this. For `ResolveNotVersion` I get it, by second cycle if change is not in the GDM something's gone awry; but can't `ResolveNotStarted` happen again legitimately in the second cycle?

lib/status_poller.go
samsalisbury: Yeah I'll see if I can pull that out, might do a general refactor in fact as I think there are a few loose threads and badly named things that need cleaning up.

lib/status_poller.go
samsalisbury: The thought behind this is that we need the status poller to eventually surface actual error messages to the user in some circumstances. Namely, if their tasks actually fail due to code/config issues, we need to present more detailed message to the user (which is currently surfaced on the `/status` endpoint, albeit not with temporal accuracy at present). I would like polling to return some errors up to the CLI as "This is the specific error you're having and need to address," rather than just rely on general log output to inform them of the issue. We may want to log other transient errors along the way, but some errors will be specially relevant to the invoker of `sous deploy` so treating them differently seems reasonable. Right now nothing useful is being passed up here, granted, but I think we can start passing useful stuff up now... More replies on this below.

lib/status_poller.go
samsalisbury: I'm happy to not return this error.

lib/status_poller.go
samsalisbury: The idea is that the CLI might want to do something with this error, since ResolveFailed errors are likely to be specially relevant to the invoker of 'sous deploy'.

lib/status_poller.go
samsalisbury: It seems clean to return an error to the CLI when the operation as a whole failed. This means the CLI needs no knowledge of which ResolveStates are  failures or otherwise, it can just check if it receives an error, and if so ensure to exit with nonzero code.

lib/status_poller_test.go
samsalisbury: Ah, that was an oversight, I'll put them back in. I think I was confused at first because `TestSubPoller_computeState` never actually called `computeState` directly, but rather just called `finished` as part of an apparent assertion. This had the side effect of transitively calling `computeState`, which was the crux of getting everything in the right state, but as it was buried in an assertion that was not obvious. Now the test directly calls `computeState` but I'll add those `finished` assertions back as well.

lib/status_poller_test.go
```diff
@@ -441,11 +599,11 @@ func TestStatusPoller_MesosFailed(t *testing.T) {
 		if err != nil {
 			t.Errorf("Error starting poller: %#v", err)
 		}
-		t.Logf("Returned state: %#v", rState)
+		t.Logf("Returned state: %s", rState)
 		testCh <- rState
 	}()
 
-	timeout := 300 * time.Millisecond
+	timeout := 10 * PollTimeout
```
samsalisbury: Oops, `10x` is overkill here. The important thing was setting the timeouts in terms of the ticker duration, so we can change that and tests should just work. 3x `PollTimeout` should be sufficient to avoid any race conditions surfacing in these tests. On that note, we should set a context to now+PollTimeout on each `pollOnce` call. I'll do that.

samsalisbury: cc @nyarly FYI ^

lib/status_poller.go
samsalisbury: Hmm, thinking about it, when a context is cancelled for any reason, it [must return a non-nil error on `Err`](https://golang.org/src/context/context.go?s=3760:4073#L99) so maybe using it as a general control flow is abusing its semantics? Or maybe I'm overthinking it...

lib/status_poller.go
samsalisbury: OK, got my head around it... Fixed in b646e45e650bd3626260c6b8f9adb05138b12a25 I think `ResolveNotStarted` is a confusing name for its semantics as used, and `ResolveNotVersion` could possibly be clearer too. How about `ResolveNoChangesDetected` and `ResolveOtherVersionInProgress` ? @nyarly

samsalisbury: Noted this in the changelog as well.

PR #492, opened 2017-10-25
'sous config' allows fixing broken config

samsalisbury:
The issue was that NameCache was being injected unconditionally, and it requires a valid config, so validation was running and causing sous to barf with invalid config on all commands (including 'sous config' which should tolerate invalid config so it can be fixed).

The unconditional injection of the name cache has been removed, and changes made to both `cli` and `cmdr` packages. `cmdr` no longer forwards flags from commands higher up the tree, and now only calls the `Parsed` hook once, for the final command in the invocation. This means that where we relied on things being added to the graph by 'RegisterOn' methods higher in the tree, we now need to add them directly to the leaf command. Flags from the `Sous` command which were passed down are now implemented as global flags using a mechanism already present in cmdr package. Help output has been fixed to additionally show all global flags.

Some more work may be necessary to guarantee the singleton nature of the name cache, so this is work in progress until that problem is solved, which I'm working on presently. The aim is to simplify construction of the graph to happen exactly once in the `Parse` hook with no further `Add` calls.

**EDIT** This change also includes a couple of tweaks to psyringe:

* When calling `Add` and getting panic because the injection type is already added, the panic message now includes the file and line number where the type was already added previously.
* Set the env var `PSYRINGE_DEBUG_FILE=psyringe.debug` to have all injections in a sous invocation logged in that file, namely which structs and constructors are being injected into with which types.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -3,10 +3,12 @@
 package cli
 
 import (
+	"flag"
 	"fmt"
 	"io"
 	"os"
 
+	"github.com/opentable/sous/config"
 	"github.com/opentable/sous/graph"
 	"github.com/opentable/sous/util/cmdr"
 	"github.com/opentable/sous/util/logging"
@@ -104,6 +106,7 @@ func BuildCLIGraph(root *Sous, cli *CLI, out, err io.Writer) *graph.SousGraph {
 //	if g, ok := cli.scopedGraphs[cmd]; ok {
 //		return g
 //	}
+
 //
 //	parent := cli.scopedGraphs[under]
 //
@@ -121,6 +124,8 @@ func NewSousCLI(di *graph.SousGraph, s *Sous, out, errout io.Writer) (*CLI, erro
 	stdout := cmdr.NewOutput(out)
 	stderr := cmdr.NewOutput(errout)
 
+	var verbosity config.Verbosity
+
 	cli := &CLI{
 		CLI: &cmdr.CLI{
 			Root: s,
@@ -131,8 +136,19 @@ func NewSousCLI(di *graph.SousGraph, s *Sous, out, errout io.Writer) (*CLI, erro
 			// uses the standard flag.ErrHelp value to decide whether or not to show
 			// this.
 			HelpCommand: os.Args[0] + " help",
+			GlobalFlagSetFuncs: []func(*flag.FlagSet){
+				func(fs *flag.FlagSet) {
+					fs.BoolVar(&verbosity.Silent, "s", false,
+						"silent: silence all non-essential output")
+					fs.BoolVar(&verbosity.Quiet, "q", false,
+						"quiet: output only essential error messages")
+					fs.BoolVar(&verbosity.Loud, "v", false,
+						"loud: output extra info, including all shell commands")
+					fs.BoolVar(&verbosity.Debug, "d", false,
+						"debug: output detailed logs of internal operations")
+				},
+			},
 		},
-
 		baseGraph: di,
 	}
 
@@ -147,6 +163,7 @@ func NewSousCLI(di *graph.SousGraph, s *Sous, out, errout io.Writer) (*CLI, erro
 	//	return nil
 	//}
 	cli.Hooks.Parsed = func(cmd cmdr.Command) error {
+		rootGraph.Add(&verbosity)
 		if registrant, ok := cmd.(interface {
 			RegisterOn(Addable)
 		}); ok {
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 
 	"github.com/opentable/sous/config"
 	"github.com/opentable/sous/graph"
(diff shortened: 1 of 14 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

LGTM

</p>

cli/cli.go
```diff
@@ -128,38 +110,41 @@ func NewSousCLI(di *graph.SousGraph, s *Sous, out, errout io.Writer) (*CLI, erro
...
-	cli.scopedGraphs = map[cmdr.Command]*graph.SousGraph{s: rootGraph}
+	addVerbosityOnce := sync.Once{}
 
 	cli.Hooks.Parsed = func(cmd cmdr.Command) error {
-		chain = append(chain, cmd)
+		addVerbosityOnce.Do(func() {
+			cli.graph.Add(&verbosity)
+		})
+		if registrant, ok := cmd.(interface {
+			RegisterOn(Addable)
```
nyarly: I'm not sure what purpose making this interface anonymous serves?

cli/cli.go
```diff
@@ -128,38 +110,41 @@ func NewSousCLI(di *graph.SousGraph, s *Sous, out, errout io.Writer) (*CLI, erro
...
 
-	chain := []cmdr.Command{}
-	rootGraph := BuildCLIGraph(s, cli, out, errout)
-	s.RegisterOn(rootGraph)
+	buildCLIGraph(s, cli, out, errout)
 
-	cli.scopedGraphs = map[cmdr.Command]*graph.SousGraph{s: rootGraph}
+	addVerbosityOnce := sync.Once{}
 
 	cli.Hooks.Parsed = func(cmd cmdr.Command) error {
```
nyarly: Ultimately, I think this change is for the good. We weren't getting much out of the nested chain of Add() calls, and they were a giant headache.

cli/sous.go
```diff
@@ -30,18 +25,7 @@ type Sous struct {
...
-
-	/*
-		This ensures the singularity of the field types - otherwise, if they're
-		injected twice and we have issues.
-
-		This is a temporary fix ahead of transitioning to a simpler DI.
-		 - jdl 9/28/17
-	*/
-	// added as a field here so that it will be singleton for the app
-	SingletonNameCache *docker.NameCache
```
nyarly: This fix is for a real issue. How is the namecache singularity guaranteed now?

graph/graph.go
```diff
@@ -545,12 +545,6 @@ func NewCurrentState(sr StateReader, log LogSink) (*sous.State, error) {
 
 // NewCurrentGDM returns the current GDM.
 func NewCurrentGDM(state *sous.State) (CurrentGDM, error) {
-	if state == nil {
```
nyarly: Likewise: how is the nil state guarded now?

util/cmdr/cli.go
```diff
@@ -190,20 +190,14 @@ func (c *CLI) prepare(cmd Command, cmdArgs []string, flagAddFuncs []func(*flag.F
 	cmdArgs = cmdArgs[1:]
 	// Add and parse flags for this command.
 	if cmdWithFlags, ok := cmd.(AddsFlags); ok {
-		if flagAddFuncs == nil {
-			flagAddFuncs = []func(*flag.FlagSet){}
-		}
-		// add these flags to the agglomeration
```
nyarly: :+1:

cli/cli.go
```diff
@@ -128,38 +110,41 @@ func NewSousCLI(di *graph.SousGraph, s *Sous, out, errout io.Writer) (*CLI, erro
...
 		return nil
 	}
 
-	// Before Execute is called on any command, inject it with values from the
-	// graph.
+	// Before Execute is called on any command, inject its dependencies.
 	cli.Hooks.PreExecute = func(cmd cmdr.Command) error {
-		// Create the CLI dependency graph.
-
-		for n, c := range chain {
```
nyarly: :+1: :heart:

cli/cli.go
samsalisbury: That just came out of refactoring, was quicker to type this anonymous interface than look for the interface name. It's only used in one place (here) so not sure the named interface is useful, but happy to put it back if that makes it more readable to you.

cli/sous.go
samsalisbury: As noted in the main PR comment, I am not 100% sure that the singleton nature of NameCache is guaranteed, need to verify this. (I thought I had applied the `wip` label to the PR but apparently not..

graph/graph.go
samsalisbury: I believe nil state is now impossible since fixing psyringe issue, but I could be wrong in this case I'll check.

samsalisbury: None of the real implementations of `ReadState() (*sous.State, error)` can return `nil, nil` they all guarantee at least one non-nil output. This means that the nil check here is no longer necessary since psyringe now guarantees that "non-nil error from constructor" -> "do not call dependent constructors". The only implementation that allows `nil, nil` return is `DummyStateManager` which is only used in tests. Therefore, I don't think this nil check is useful any more.

samsalisbury:
Just realised I had not committed the newly vendored psyringe with file:line where previously added type was added, done that and reinstated `Registrant` interface. The problem of conflicting code could be an issue though, let's look at that later.

PR #496, opened 2017-10-27
Build strategies design doc

nyarly:
Contrary to usual practice,
if you disagree or would like to add to the document,
please counter-PR with your changes.
We can go back and forth until we're happy,
or at least it's in an intermediate state we can tolerate.

### DIFF
--- a/doc/build_strategies_design.md
+++ b/doc/build_strategies_design.md
@@ -41,27 +41,27 @@ it documents our design process and ideas.*
 Each manifest has a `Build` stanza describing how sous should build its images.
 ```
   Build:
-    Type: mount-run-split
+    Strategy: mount-run-split
     Image: docker.internal.com/our-maven:latest
 ```
 
-Possible `Type` values could be: 
+Possible `Strategy` values could be: 
 
 * `mount-run-split` meaning we run the `Image` with various directories from
   the local machine mounted. The result is that one of those directories is
   populated with built runnable artifacts. We then split those artifacts
   amongst various output containers ready for deployment.
 * `simple-dockerfile` meaning we simply build the `Dockerfile` in the project's
-  source code repository (actually `SourceLocation`). This `Type` would not
+  source code repository (actually `SourceLocation`). This `Strategy` would not
   use the `Image` field at all and would is capable of producing only one
   artifact at a time (the docker image produced by `docker build`).
 
-Simple-dockerfile type builds are already supported by Sous, and are currently the
+Simple-dockerfile strategy builds are already supported by Sous, and are currently the
 default when a Dockerfile is present in the working directory when executing
 'sous build'.
 
 The rest of this document describes in more detail the `mount-run-split` build
-type, which we think should be the standard type used by most projects.
+strategy, which we think should be the standard strategy used by most projects.
 
 ### mount-run-split builds
 
@@ -109,14 +109,8 @@ However, using a Dockerfile for this purpose allows for the "small changes" requ
 
 It should be possible for Sous
 to add labels related to the build process
-to the final deployed images,
-which would give us the data required for audits -
-effectively, we could chase the long tail of the histogram
-of "build image digest"
-(because variant builds would have a different digest -
-I'm 85% sure that "conformant" builds would all have the same digest),
-and other labels would provide enough data
-to track down the actual built dockerfile.
+to the final build images
+for auditing purposes.
 
 We could add a /sous directory to built deploy artifacts
 and store e.g. the build Dockerfile there for reference purposes.
@@ -124,7 +118,7 @@ and store e.g. the build Dockerfile there for reference purposes.
 In order to allow for easy development of new build containers,
 and to effect transition to existing ones,
 `sous build` would be extended with two new flags:
-`-dev` would allow a number of non-production-suitale changes
+`-dev` would allow a number of non-production-suitable changes
 to the Sous build process, like
 caching of intermediate artifacts,
 flags to build programs to output more logging.
@@ -135,7 +129,7 @@ instead of consulting the manifest.
 
 Right after getting the manifest to build from,
 Sous should issue `git config` commands to store
-the type and
+the strategy and
 the image
 that it discovers.
 Then, if on a future build it can't get the manifest
@@ -163,7 +157,7 @@ a signal that we should look for a local Dockerfile
 
 We might have a period where we encourage users to switch
 to having a Build section
-even if it's to have `Type: simple-dockerfile`.
+even if it's to have `Strategy: simple-dockerfile`.
 Then, we'd trust what we find in the manifest,
 but fall back to looking for a Dockerfile.
 
--- a/doc/build_strategies_design.md
+++ b/doc/build_strategies_design.md
@@ -99,19 +99,19 @@ see below.)
 
 We then `docker run` the built image with four mounted volumes:
 
-* `/input` for the source artifacts
+* `/input` for the source (i.e. code and other things checked into the repo)
 * `/vendor` for externally fetched dependencies
 * `/working` for intermediate products
 * `/output` for output products (e.g. jar files, directories, executables etc.)
 
-Sous is responsible for choosing
-the host directories to mount.
-The `/input` directory is mounted from the current Sous context.
-The `/output` directory will be a temporary directory created per build.
+Sous is responsible for making a copy of files from the current sous context,
+and copying them into the `/input` directory inside the container.
 
-The `/vendor` and `/working` directories will depend on invocation,
-but we anticipate them being under
-a Sous-specific subdirectory of
+The `/output` directory will be a temporary on the sous user's machine,
+mounted inside the container. A new `/output` dir will be created per build.
+
+The `/vendor` and `/working` directories will be mounted from the local filesystem,
+we anticipate them being under a Sous-specific subdirectory of
 `$XDG_DATA_HOME`
 or
 `$XDG_CACHE_HOME`
@@ -119,6 +119,7 @@ or
 `~/.local/share/sous`
 and
 `~/.cache/sous`.)
+
 Directories should be created for each unique project,
 although arguably the `/vendor` directory should
 be per build image,
@@ -161,25 +162,54 @@ A runspec looks like this:
 }
 ```
 
-Any number of objects can be in the `images` object.
+The `images` object is keyed by `offset` which corresponds to a subdirectory in the repo.
+The `images` object can therefore contain up to one item per subdirectory in the repo (recursively).
+Acceptable `offset` values include sub-sub-directories and so-forth, e.g.: `service` `src/service` `src/service/api` etc).
+
 Sous will convert this runspec into
-a series of synthetic dockerfiles like
+a series of synthetic Dockerfiles based on the runspec data, e.g.:
+
 ```Dockerfile
 FROM microsoft/aspnetcore:2.0
 COPY /srv /
 CMD ["dotnet", "/srv/service.dll"]
 ```
 
-and `build` them in the directory mounted on `/output`.
+It will `build` them in the directory mounted on `/output`.
+Note that at this point all compilation has taken place,
+and the 'build' should be merely a `COPY` operation.
+
 The resulting images will be labelled with Sous metadata,
 with the offset pulled from the name of the image object
 (in this case: `service`.)
 
 It is the responsibility of the build image
-and its client projects
-to coordinate their offset sub-projects.
-E.g. Maven projects have the idea of "submodules"
-which can be hooked on to define offset images.
+to produce at most one offset per subdirectory,
+and to determine which subdirectories represent runnable items.
+
+For example, Maven projects have the idea of "submodules"
+defined in their pom.xml file, which correspond to
+subdirectories. At present there is nothing inside the `pom.xml` that indicates
+if a particular module corresponds to a runnable/deployable artifact.
+We will need to solve this problem, some potential options are:
+
+1. Require that any deployable modules are first `sous init`ed and check for the existence
+   of a manifest for each.
+2. Require a file in the root of the source repo listing deployable offsets.
+3. Establish a rule that `pom.xml` files must add some metadata to the `modules` section about
+   which offsets are to be considered deployable. (Other ecosystems would need an equivalent
+   rule in this case.)
+
+> I (Sam) favour option 1 as that ensures that we have deployment configuration for these
+offsets as well, so that we can actually deploy them. 'sous init' needs to provide guidance
+about 'sous init'ing offsets, and 'sous build' / 'sous deploy' need to print very clear messages
+about each item they are building and deploying so that if any are missing they can be added
+by the user.
+
+> **Note** we could then allow recognising tags for differing offsets using tag prefixes to support
+versioning offsets differently where necessary. In the absence of prefixed-tags we could assume that
+the entire repo is versioned as one. We would probably need to be explicit about which model is 
+used in the 'project' definition in the GDM.
 
 ### Observations
 
--- a/doc/build_strategies_design.md
+++ b/doc/build_strategies_design.md
@@ -32,24 +32,44 @@ sous build should be able to
 
 ## Current Proposal
 
-The named configuration is configured using the one-line-dockerfile - which points to a build container.
+The manifest for the project has a sub-entry like:
+```
+  Build:
+    Type: mount-run-split
+    Image: docker.internal.com/our-maven:latest
+```
 
-`sous build` "builds" this container,
-and then `docker run`s it with three mounted volumes:
+`mount-run-split` is the goal design here.
+We might bless `simple-dockerfile` and `build-copy-split` as well.
+For the time being, the type is superfluous - if there's
+a non-empty image field, we proceed with this proposal.
+It provides an avenue for future alternatives,
+however.
+
+We start by synthesizing a small Dockerfile like
+```
+FROM docker.internal.com/our-maven:latest
+```
+`sous build` "builds" this container.
+(We could simply `docker pull` the image,
+but we get the advantage
+of having a single code path for one-off experiments.)
+It then `docker run`s it with three mounted volumes:
 one for fetched dependences,
 one for intermediate products
 and one for output products (i.e. jar files).
-The runspec is delivered into the output volume,
+The runspec is delivered into the output volume
+by the build tools and scripts in the build image,
 and Sous uses the runspec to scatter products in the output into run images.
-The actual build images will be responsible
-for arranging the volumes into their needed directory structure
-(e.g. symlinks or mount -o bind)
-However, part of the recognition process
-for "mount-run-split" images will be enviroment variables specifying those three mount points.
-(The output mount will be required, the caches will be optional.)
+The actual build images will also be responsible
+for arranging the volumes into the directory structure
+required by the build tools
+(e.g. symlinks or mount -o bind for ~/.mvn2).
+
+### Observations
 
 Because the build image will generally just be downloaded,
-the `docker build` of the initial container will generally be (almost) a no-op -
+the `docker build` of the initial container will usually be (almost) a no-op -
 the first time there'll be download of data,
 and thereafter there'll be a check of the local daemon's cache of the build image.
 However, using a Dockerfile for this purpose allows for the "small changes" required.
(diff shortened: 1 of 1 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

LGTM

</p>

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+  (e.g. Maven's .m2 or node's node_modules or Ruby gems).
+* Cache intermediate build products
+  (for local development.)
+* Produce small-as-possible product image.
+* Produce more than one product image.
+* Users should be able to make small overrides to the build process,
+  if only so that quick experiments are possible
+* Sous controllers should be able to audit the build processes in place,
+  so that they can be consolidated -
+  i.e. "small overrides" should be reasonably public so that we can find audit divergences.
```
samsalisbury: Not sure what "Sous controllers" means? Maybe define it here?

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+## Current Proposal
+
+The manifest for the project has a sub-entry like:
+```
+  Build:
+    Type: mount-run-split
+    Image: docker.internal.com/our-maven:latest
+```
```
samsalisbury: 👍

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+
+`mount-run-split` is the goal design here.
+We might bless `simple-dockerfile` and `build-copy-split` as well.
+For the time being, the type is superfluous - if there's
+a non-empty image field, we proceed with this proposal.
+It provides an avenue for future alternatives,
+however.
+
+We start by synthesizing a small Dockerfile like
```
samsalisbury: I think the dockerfile for this purpose should not be called `Dockerfile` as the convention is for `Dockerfile` to produce a runnable image on `docker build`; Maybe we call this `Dockerfile.build` or similar to allow projects to provide standard `Dockerfile` as well? I would rather we did not have this sous-specific file inside the repo, but I can live with it if we make this change. An alternative would be to allow adding a Dockerfile to the manifest itself, and ensure it's really east to access and edit (e.g. `sous build -edit`) akin to `sous manifest edit`. This would keep Sous well and truly out of code repos, which makes our assertion that configuration should not be in those repos stronger. (If we allow customising the sous build here, why not the deploy configs as well?)

Since Manifests are YAML, we can use string literals to embed a Dockerfile in the YAML. As long as we make it extremely east to edit, I can't see an issue with having this in the manifest. You could just specify `Image` like the example above, or `Dockerfile` as proposed here. We already have a mechanism for detecting whether specific manifest changes result in re-deploy. Build Dockerfile changes would not cause a deployment, but rather impact future builds.

One objection to this approach is that it means connection to Sous server is required for `sous build` however we can remedy this by sous client maintaining a local cache of your manifest. Further, it should be possible to make local edits to the manifest as cached, and only push changes to `Dockerfile` section once you've had a chance to run the build locally. I think there is a bigger discussion here, as local caching of the entire GDM could actually be useful for future features (like being able to spin up a portion of OpenTable as a whole to do local integration testing between services).

I would suggest we first make the simple change and require connection to sous server to do `sous build` and later address the offline builds/caching issue.

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+one for fetched dependences,
+one for intermediate products
+and one for output products (i.e. jar files).
+The runspec is delivered into the output volume
+by the build tools and scripts in the build image,
+and Sous uses the runspec to scatter products in the output into run images.
+The actual build images will also be responsible
+for arranging the volumes into the directory structure
+required by the build tools
+(e.g. symlinks or mount -o bind for ~/.mvn2).
```
samsalisbury: 👍

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+It should be possible for Sous
+to add labels related to the build process
+to the final deployed images,
+which would give us the data required for audits -
+effectively, we could chase the long tail of the histogram
+of "build image digest"
+(because variant builds would have a different digest -
+I'm 85% sure that "conformant" builds would all have the same digest),
+and other labels would provide enough data
+to track down the actual built dockerfile.
```
samsalisbury: If the build Dockerfile is part of the manifest, this becomes a very simple issue to solve.

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+and to effect transition to existing ones,
+`sous build` would be extended with two new flags:
+`-dev` would allow a number of non-production-suitale changes
+to the Sous build process, like
+caching of intermediate artifacts,
+flags to build programs to output more logging.
+`-override-build-image <Dockerfile>`
+would only be available in `-dev` mode,
+and would use the named Dockerfile
+instead of consulting the manifest.
```
samsalisbury: 👍 `-override-build-image` might not be necessary if we use `sous build -edit` as proposed above (or a better named command).

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+
+Right after getting the manifest to build from,
+Sous should issue `git config` commands to store
+the type and
+the image
+that it discovers.
+Then, if on a future build it can't get the manifest
+(e.g. no network connection, Sous server is down)
+it can consult the local git configuration
+to recover that information.
```
samsalisbury: Ah, some kind of local cache, I like it!

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+
+At some point in the future,
+we can convert to a notarized platform
+where the build images would need to be notarized
+in order to be used
+(or at least "unsigned build image" becomes an advisory)
+which would encourage/require
+that updates to the build images be reviewed and approved.
+In the meantime, the `Build>Image` fields on manifests
+would form a catalogue of build images.
```
samsalisbury: 👍

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+I think we can consider layers vs. image a *specification* detail -
+but for the purpose of requirements,
+I want the build images to
+a) occupy as little space on disk as possible and
+b) to take as little time to transmit
+from artifact repository
+to execution agents
+as possible.
+Overall small images,
+and small unique layers both address those requirements.
```
samsalisbury: That's pretty much the wording I was looking for, I'll update PR to this.

samsalisbury:
I'll prepare a counter-PR with these changes.

doc/build_strategies_design.md
nyarly: I'm honestly not sure either. "Whoever it is in charge of auditing build processes" - which for OT means our team.

doc/build_strategies_design.md
nyarly: By "synthesizing a small Dockerfile" I mean "Sous produces it in a bytes.Buffer and uses it as the stdin to exec.Command("docker", "build") - the file itself wouldn't ever live on a filesystem.

By limiting the *manfiest* entry to an image, it means that "blessed" build images will need to be built and pushed external to a particular project. (Our internal set of build images being an example of that process.) I *don't* want to start inlining Dockerfiles into the manifest - that seems like an auditing nightmare.

See `-override-build-dockerfile` below for the "but I want to do one-off tweaks"

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+The manifest for the project has a sub-entry like:
+```
+  Build:
+    Type: mount-run-split
+    Image: docker.internal.com/our-maven:latest
+```
+
+`mount-run-split` is the goal design here.
+We might bless `simple-dockerfile` and `build-copy-split` as well.
+For the time being, the type is superfluous - if there's
```
adamzr: I'm thinking type in this example would be `docker-parent`, if sous sees `docker-parent` it creates an implicit Dockerfile using the value of `Image`. Another type might be `docker-internal` which uses the Dockerfile in the repo of the project. In that case there is no `Image` property. There might be other types which then have other fields that clarify details (maybe `Type: shell-script` and then `File: build.sh`, etc.)

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+For the time being,
+it documents our design process and ideas.*
+
+## Design Goals
+
+sous build should be able to
+
+* Build based on a named configuration
+  (which we expect to be a build docker image name).
+* Cache fetched dependencies
```
adamzr: 🎉

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+  (e.g. Maven's .m2 or node's node_modules or Ruby gems).
+* Cache intermediate build products
+  (for local development.)
+* Produce small-as-possible product image.
+* Produce more than one product image.
```
adamzr: Why?

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+but we get the advantage
+of having a single code path for one-off experiments.)
+It then `docker run`s it with three mounted volumes:
+one for fetched dependences,
+one for intermediate products
+and one for output products (i.e. jar files).
+The runspec is delivered into the output volume
+by the build tools and scripts in the build image,
+and Sous uses the runspec to scatter products in the output into run images.
+The actual build images will also be responsible
```
adamzr: The build images should be discarded and not saved as artifacts.

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+and other labels would provide enough data
+to track down the actual built dockerfile.
+
+We could add a /sous directory to built deploy artifacts
+and store e.g. the build Dockerfile there for reference purposes.
+
+In order to allow for easy development of new build containers,
+and to effect transition to existing ones,
+`sous build` would be extended with two new flags:
+`-dev` would allow a number of non-production-suitale changes
```
adamzr: Should be "suitable"

doc/build_strategies_design.md
adamzr: 👍  But, this should produce a warning.

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+which would encourage/require
+that updates to the build images be reviewed and approved.
+In the meantime, the `Build>Image` fields on manifests
+would form a catalogue of build images.
+
+## Discussion
+
+### Migration
+
+At time of introduction, a missing or empty Build section should be treated as
```
adamzr: The build could update the manifest to add this during the build itself.

doc/build_strategies_design.md
```diff
@@ -0,0 +1,163 @@
...
+We might have a period where we encourage users to switch
+to having a Build section
+even if it's to have `Type: simple-dockerfile`.
+Then, we'd trust what we find in the manifest,
+but fall back to looking for a Dockerfile.
+
+Finally, we'd come to a point where we require the `Build` section,
+and error out if it's missing.
+
+`sous init` would need to be updated to default to adding a `Build`
```
adamzr: It should prompt you with some known options and a best guess

doc/build_strategies_design.md
adamzr: Maybe there could be a type definition spec file. For example a type might be "java-spring-boot-service". The type definition file would then say what docker image and build strategy to use. It could also define health checks, environment variables, and other manifest defaults. The type definition file could also tell sous what to look for in a repo that would indicate it's a "java-spring-boot-service", this could power your best guess system.

doc/build_strategies_design.md
samsalisbury: I think chasing down the long-tail of the histogram of "build image digest" would be a lot worse ;)

Removing that suggestion and adding commentary to the effect that 'one-off tweaks' are strictly a local dev machine phenomena and would not result in deployable artifacts due to new advisories to that effect.

doc/build_strategies_design.md
```diff
@@ -0,0 +1,202 @@
...
+to the Sous build process, like
+caching of intermediate artifacts,
+flags to build programs to output more logging.
+`-override-build-image <Dockerfile>`
+would only be available in `-dev` mode,
+and would use the named Dockerfile
+instead of consulting the manifest.
+
+Right after getting the manifest to build from,
+Sous should issue `git config` commands to store
```
samsalisbury: I'm not sure about using local git config to store this without getting the user's approval. I would rather store a map of absolute directory path to last known strategy/image somewhere under `$XGD_DATA_HOME` for example.

doc/build_strategies_design.md
```diff
@@ -6,62 +6,95 @@ it documents our design process and ideas.*
...
+  the local machine mounted. The result is that one of those directories is
+  populated with built runnable artifacts. We then split those artifacts
+  amongst various output containers ready for deployment.
+* `simple-dockerfile` meaning we simply build the `Dockerfile` in the project's
+  source code repository (actually `SourceLocation`). This `Type` would not
+  use the `Image` field at all and would is capable of producing only one
+  artifact at a time (the docker image produced by `docker build`).
+
+Simple-dockerfile type builds are already supported by Sous, and are currently the
+default when a Dockerfile is present in the working directory when executing
```
nyarly: Not true

samsalisbury: Really? Could you elaborate?

doc/build_strategies_design.md
nyarly: The config would be namespaced to sous, and local to this instance of the project. Conversely, a map of absolute paths would break if any part of the directory structure does. Given that I have tools that I use that have a similar approach, I can speak from experience and say that that does happen and is very confusing.

nyarly: What objection to foresee to adding sections to `$project/.git/config`?

doc/build_strategies_design.md
samsalisbury: Hi, can you elaborate on this?

doc/build_strategies_design.md
```diff
@@ -45,74 +47,155 @@ Each manifest has a `Build` stanza describing how sous should build its images.
...
 
 ### Observations
 
 Because the build image will generally just be downloaded,
 the `docker build` of the initial container will usually be (almost) a no-op -
 the first time there'll be download of data,
 and thereafter there'll be a check of the local daemon's cache of the build image.
-However, using a Dockerfile for this purpose allows for the "small changes" required.
+However, using a Dockerfile for this purpose
+allows for us to fulfill the "small changes" requirement.
```
samsalisbury: This needs expanding I just realised.

doc/build_strategies_design.md
```diff
@@ -45,74 +47,155 @@ Each manifest has a `Build` stanza describing how sous should build its images.
...
 
-Each builder image compatible with `mount-run-split` must contain a `runspec.json`
-file (typically at the root of the filesystem, although this can be overridden by
-adding the line `ENV SOUS_RUN_IMAGE_SPEC={some-other-path}`).` This runspec file
-specifies how runnable artifacts produced by running the build container will be
-executed once they are delivered into runnable docker images (see below).
+and `build` them in the directory mounted on `/output`.
+The resulting images will be labelled with Sous metadata,
+with the offset pulled from the name of the image object
+(in this case: `service`.)
```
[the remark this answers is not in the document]
samsalisbury: To add to the above, there is also the lesser concern that this would bind all sub-projects in a repo to be built and versioned together at the same time. That seems fairly reasonable to me, but I can imagine a use case where separate modules might want to be tagged, versioned and built separately, just as they can already be released separately. Our support of tag prefixes lays the groundwork for allowing individually-versioned maven modules in a single repo. I would like to know if there are any good use cases for this, or if this is a non-concern? cc @stevenschlansker @pennello @SegFaultAX do you know if we have a reason to allow separately versioned maven modules in a single repo at OpenTable at present? I can see that e.g. discovery and buzzsaw version all modules together as 'discovery-parent-{semver}' and 'buzzsaw-parent-{semver}'.

doc/build_strategies_design.md
```diff
@@ -45,74 +47,155 @@ Each manifest has a `Build` stanza describing how sous should build its images.
...
+and is not itself a deployable artifact.
 (We could simply `docker pull` the image,
 but we get the advantage
-of having a single code path for one-off experiments, see below.)
+of having a single code path for one-off experiments,
+see below.)
+
+We then `docker run` the built image with four mounted volumes:
+
+* `/input` for the source artifacts
```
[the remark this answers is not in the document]
nyarly: I used artifacts instead of code with e.g. front end images in mind.

doc/build_strategies_design.md
```diff
@@ -45,74 +47,155 @@ Each manifest has a `Build` stanza describing how sous should build its images.
...
+* `/vendor` for externally fetched dependencies
+* `/working` for intermediate products
```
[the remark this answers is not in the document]
nyarly: Hrm. I think you're right, but we will need to do something akin to Docker's copy-to-daemon behavior to maintain a (cleanish) input directory. For Sous builds, I don't want to retain intermediate products by default.

doc/build_strategies_design.md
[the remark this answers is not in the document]
pennello: All of the multi-module repos I've built and maintained here have had sync'd up versions.  This design certainly simplifies matters, although of course Maven supports independent version maintenance.

I recall @stevenschlansker mentioned experience with the independently-versioned approach at Ness, and there, they explicitly moved away from it due to the added complexity.  It just wasn't worth it.

I wouldn't be averse to laying down a _de facto_ rule by virtue of Sous's capabilities that we keep multi-module repo versions in sync, although it does worry me.  Maybe we run into a legit use case down the line, and our deployment tooling cuts us off.  : /

If it was sufficiently simple to do, I'd say let's go for it.  If not, I might want to ask why, although I think the answer might be complicated.  Happy to dive in if you guys think it might be helpful.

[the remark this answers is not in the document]
nyarly: Sam and I discussed this yesterday, and I have notes that need to be incorporated here. I don't want to lose the thread, so I guess I'll make some progress on the doc tomorrow.

The upshot is that mount-run-split (which perhaps needs a better name) does a single sous build per *project* which can generate (at least) one "Deployable" per offset. The implication is that we have a problem in our model design: we need a "Project" that encompasses 1 or more Manifests, which is, at least, where the build strategy lives.

Other build strategies might later be employed to manage `sous build`:Deployable scenarios (which would support independent versioning).

A further possibility would be that `sous build` consider its `-offset` flag more carefully when it runs - an explicit offset would imply filtering the deploy artifacts created and pushed. No flag would mean "do whatever the strategy says" - it's likely that each strategy (= code in sous) would handle this case differently.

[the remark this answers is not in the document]
nyarly: Expanding briefly on Projects: we'd need to add some validation to Sous server such that we include a Project section in each manifest when we send them, and update the projects if any manifest changed them.

Which breaks our "one file per update" rule in git, I just realize.

Also possible: we accept that "Project" information needs to be kept in sync by policy (i.e. humans might break that...) until we can get off of git as a datastore.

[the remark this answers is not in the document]
stevenschlansker: One example we have today -- we built a discovery load generator which naturally lives in the discovery project, but does not get released or built with the default build.  So it ends up with a separate version and release cycle, while existing as a subdirectory of the discovery project.

doc/build_strategies_design.md
```diff
@@ -161,25 +162,54 @@ A runspec looks like this:
...
-to coordinate their offset sub-projects.
-E.g. Maven projects have the idea of "submodules"
-which can be hooked on to define offset images.
+to produce at most one offset per subdirectory,
+and to determine which subdirectories represent runnable items.
+
+For example, Maven projects have the idea of "submodules"
+defined in their pom.xml file, which correspond to
+subdirectories. At present there is nothing inside the `pom.xml` that indicates
+if a particular module corresponds to a runnable/deployable artifact.
```
nyarly: Implicit in a lot of my commentary here is that there are _built_ artifacts, and a subset of those that are _deployable_ artifacts. Examples are uploader images, or smoke testers, or other tested artifacts. (We discussed one project that has a very complicated build pipeline, which I think we could support by producing many artifacts with different subsets of their build products...)

doc/build_strategies_design.md
```diff
@@ -161,25 +162,54 @@ A runspec looks like this:
...
+We will need to solve this problem, some potential options are:
```
nyarly: I contend that, in the scope of the build strategies **engine**, no we don't. We need to provide for a build image to _fail_ and emit explanations about why. We probably need to establish best practices about how a build image determines that it's source is "acceptable" - e.g. has deployable outputs.

Arguably, Sous should check the runspec.json and confirm that every offset described in the GDM is present (modulo `sous build -offset ...` filtering) but not until we have lifecycle phases implemented. Otherwise, you have to build an image for every offset you ever mentioned to Sous.

doc/build_strategies_design.md
```diff
@@ -161,25 +162,54 @@ A runspec looks like this:
...
+
+1. Require that any deployable modules are first `sous init`ed and check for the existence
+   of a manifest for each.
+2. Require a file in the root of the source repo listing deployable offsets.
+3. Establish a rule that `pom.xml` files must add some metadata to the `modules` section about
+   which offsets are to be considered deployable. (Other ecosystems would need an equivalent
```
nyarly: POM files and "other ecosystems" are exactly why I feel that the build engine should be agnostic about this behavior.

samsalisbury:
New requirement: some builders/plugins e.g. Maven plugins require the .git repo to be present inside the builder container. Should we always copy/mount including .git or make this an option of the builder image? cc @nyarly @JC1738

samsalisbury:
We will copy for sure because it's unsafe to do otherwise.

samsalisbury:
Closing this out to remove visual noise. It's already linked from JIRA ticket, so we can come back to this conversation when necessary.

nyarly:
I thought we'd agreed to pull this into the docs directory with a caveat at the head of the document? A link in Jira seems too easy to miss.

PR #500, opened 2017-10-31
Output from status polling SOUS-259

nyarly:
The Update action and the Status pollers now emit console messages,
which means that `sous deploy` is much more descriptive of what it's doing.

Note that I also found and fixed a bug where status was checked in a very tight loop
and was issuing far too many update messages on kafka.

### DIFF
--- a/lib/messages.go
+++ b/lib/messages.go
@@ -27,18 +27,18 @@ type (
 	}
 )
 
-func reportPollerResolved(logsink logging.LogSink, sp *StatusPoller, status ResolveState, err error) {
-	msg := newPollerResolvedMessage(sp, status, err)
-	logging.Deliver(msg, logsink)
-}
-
 func reportPollerStart(logsink logging.LogSink, poller *StatusPoller) {
 	msg := newPollerStartMessage(poller)
 	logging.Deliver(msg, logsink)
 }
 
-func reportPollerStatus(logsink logging.LogSink, poller *StatusPoller) {
-	msg := newPollerStatusMessage(poller)
+func reportPollerResolved(logsink logging.LogSink, sp *StatusPoller, status ResolveState, err error) {
+	msg := newPollerResolvedMessage(sp, status, err)
+	logging.Deliver(msg, logsink)
+}
+
+func reportPollerStatus(logsink logging.LogSink, poller *StatusPoller, old ResolveState) {
+	msg := newPollerStatusMessage(poller, old)
 	logging.Deliver(msg, logsink)
 }
 
@@ -126,10 +126,11 @@ func (msg *pollerResolvedMessage) EachField(f logging.FieldReportFn) {
 	f("deploy-status", msg.status.String())
 }
 
-func newPollerStatusMessage(poller *StatusPoller) *pollerStatusMessage {
+func newPollerStatusMessage(poller *StatusPoller, old ResolveState) *pollerStatusMessage {
 	return &pollerStatusMessage{
-		callerInfo: logging.GetCallerInfo("sous/lib/messages"),
-		poller:     poller,
+		callerInfo:    logging.GetCallerInfo("sous/lib/messages"),
+		poller:        poller,
+		previousState: old,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

lib/status_poller.go
```diff
@@ -182,17 +184,6 @@ func (sp *StatusPoller) poll(subs []*subPoller) ResolveState {
...
-	for {
-		select {
-		default:
-			return
-		case update := <-sp.results:
-			sp.nextSubStatus(update)
-		}
-	}
-}
-
```
samsalisbury: Ah, thanks I added this to test something, didn't mean to leave it in.

lib/resolvestate.go
```diff
@@ -88,6 +88,40 @@ func (rs ResolveState) String() string {
...
+	case ResolveFailed:
+		return "The attempt to resolve this service failed"
+	case ResolveHTTPFailed:
+		return "HTTP connection to the Sous server has failed"
+	case ResolveComplete:
+		return "Deployment resolution is complete"
+	case ResolveMAX:
+		return "resolve maximum marker - not a real state, received in error?"
+	}
+}
```
samsalisbury: I like it.

lib/messages.go
```diff
@@ -148,6 +155,17 @@ func (msg *pollerStatusMessage) EachField(f logging.FieldReportFn) {
...
+func (msg *pollerStatusMessage) WriteToConsole(console io.Writer) {
+	if msg.poller.status == msg.previousState {
+		console.Write([]byte{'.'})
+		return
+	}
+	fmt.Fprintf(console, "\n%s", msg.poller.status.Prose())
+	if msg.poller.finished() {
+		console.Write([]byte{'\n'})
+	}
+}
```
samsalisbury: 👍

lib/resolvestate.go
```diff
@@ -88,6 +88,40 @@ func (rs ResolveState) String() string {
...
+	default:
+		return "unknown (oops)"
+	case ResolveNotPolled:
+		return "No data from server yet"
+	case ResolveNotStarted:
+		return "Waiting for server to begin resolution"
+	case ResolvePendingRequest:
+		return "Waiting for request to be made to cluster"
+	case ResolveNotVersion:
+		return "Waiting for server to acknowledge new version"
```
samsalisbury: I would like this status to be accompanied by the version currently deployed or in progress when presented to the user. I wonder what the best way to achieve that is? Swift provides parameterised enums, so different statuses can have different attributes (e.g. `ResolveNotVersion(inProgress: version, currentlyDeployed: version)`)  which would be handy here, doing that in go would require using something other than integers though & we currently rely on integer calculus. Other statuses likewise could benefit from additional info in their prose.

lib/resolvestate.go
samsalisbury: I suspect these would be easier to use if they all began with lowercase characters and all formed the same part of speech, so they could be inserted in a sentence e.g. if we imagine them all to begin with the phrase "Sous is ...". Some of them already work e.g. `ResolveNotStarted` but others would need to be changed e.g. `ResolveTasksStarting` could be "waiting for instances to boot up".

lib/status_poller.go
samsalisbury: Actually there was a point to this... It should mean we waste fewer cycles of the outer loop in the goroutine spun up by `StatusPoller.poll` as there's no point calculating `subPoller.finished` if there are further statuses already received waiting to be read. Not that I care about that optimisation much, happy to leave it out if it makes logging easier.

lib/resolvestate.go
```diff
@@ -91,34 +93,35 @@ func (rs ResolveState) String() string {
...
+		return "receiving an unexpected HTTP response from Sous server"
 	case ResolveErredRez:
-		return "Transient error on server, retrying"
+		return "re-trying after experiencing a transient error"
 	case ResolveTasksStarting:
-		return "Cluster has accepted deploy request, awaiting service boot"
+		return "waiting for instances to start on Singularity"
 	case ResolveNotIntended:
-		return "Sous server does not intend to deploy this version - probably another deploy request received"
+		return "not intending to perform this deployment (attempting to deploy a different version)"
```
samsalisbury: After discussion we will add note that this is prob caused by another deploy coming in like the original did. Update coming.

PR #505, opened 2017-11-07
Optional injection #trivial

samsalisbury:
Injection is now mandatory for all struct fields injected into in the CLI.

You can add `inject:"optional"` tag to any field to disable this.

Trivial because this is dev-facing change only. We should run the CLI tests if poss to verify this doesn't break anything. Integration + unit tests passing, and casual use of this build verifies it's at least mostly working still.


sous-machine-account:
LGTM

</p>

PR #507, opened 2017-11-09
Adding Schedule field #SOUS-8 #SOUS-753

nyarly:
Manifests now get a Schedule field for cron-style scheduled tasks.
Note that because of how serialization works, this field will be added to all manifests -
watch for "more than one file"  errors during deploy.
(The GDM may need to be renormalized.)

Also appearing: a lightly annotated diff-log of the changes required,
and the change/test process I go through to add a field.
Should make adding other fields more reasonable.


sous-machine-account:
LGTM

</p>

PR #509, opened 2017-11-10
Logging: CLI respects Logging.Basic.Level SOUS-756

samsalisbury:
- Before, we always had a logSet configured to use os.Stderr
- Now we use the graph, including config to create a log set, which therefore respects config: `Logging.Basic.DisableConsole`

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -130,8 +130,8 @@ const (
 
 // BuildGraph builds the dependency injection graph, used to populate commands
 // invoked by the user.
-func BuildGraph(vrsn semv.Version, ls *logging.LogSet, in io.Reader, out, err io.Writer) *SousGraph {
-	graph := BuildBaseGraph(vrsn, ls, in, out, err)
+func BuildGraph(vrsn semv.Version, in io.Reader, out, err io.Writer) *SousGraph {
+	graph := BuildBaseGraph(vrsn, in, out, err)
 	AddFilesystem(graph)
 	AddNetwork(graph)
 	AddState(graph)
@@ -151,12 +151,12 @@ func newSousGraph() *SousGraph {
 }
 
 // BuildBaseGraph constructs a graph with essentials - intended for testing
-func BuildBaseGraph(vrsn semv.Version, ls *logging.LogSet, in io.Reader, out, err io.Writer) *SousGraph {
+func BuildBaseGraph(vrsn semv.Version, in io.Reader, out, err io.Writer) *SousGraph {
 	graph := newSousGraph()
 	// stdout, stderr
 	graph.Add(
 		vrsn,
-		ls,
+		//ls,
 		func() InReader { return in },
 		func() OutWriter { return out },
 		func() ErrWriter { return err },
@@ -180,6 +180,7 @@ type adder interface {
 // AddLogs adds a logset to the graph.
 func AddLogs(graph adder) {
 	graph.Add(
+		newLogSet,
 		newLogSink,
 		newMetricsHandler,
 	)
@@ -311,10 +312,24 @@ func newRegistryDumper(r sous.Registry) *sous.RegistryDumper {
 	return sous.NewRegistryDumper(r)
 }
 
+func newLogSet(v semv.Version, config LocalSousConfig) (*logging.LogSet, error) {
+	if config.Logging.Basic.DisableConsole {
+		return logging.NewLogSet(v, "", "", ioutil.Discard), nil
+	}
+	ls := logging.NewLogSet(v, "", "", os.Stderr)
+	cerr := ls.Configure(config.Logging)
+	if cerr != nil {
+		return ls, initErr(cerr, "validating logging configuration")
+	}
+	cerr = logging.Log.Configure(config.Logging)
+	return ls, initErr(cerr, "validating logging configuration")
+}
+
 func newLogSink(v *config.Verbosity, set *logging.LogSet) LogSink {
 	set.Configure(v.LoggingConfiguration())
 
 	//logging.Log.Warn.Println("Normal output enabled")
+	set.Info("Info debugging enabled")
 	set.Vomitf("Verbose debugging enabled")
 	set.Debugf("Regular debugging enabled")
 	return LogSink{set}
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -312,17 +312,28 @@ func newRegistryDumper(r sous.Registry) *sous.RegistryDumper {
 	return sous.NewRegistryDumper(r)
 }
 
-func newLogSet(v semv.Version, config LocalSousConfig) (*logging.LogSet, error) {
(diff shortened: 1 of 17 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

graph/graph.go
```diff
@@ -311,10 +312,24 @@ func newRegistryDumper(r sous.Registry) *sous.RegistryDumper {
 	return sous.NewRegistryDumper(r)
 }
 
+func newLogSet(v semv.Version, config LocalSousConfig) (*logging.LogSet, error) {
```
nyarly: This will lead to a regression of the `sous config` problem....

graph/local_config.go
```diff
@@ -40,17 +40,9 @@ func newPossiblyInvalidLocalSousConfig(u config.LocalUser, defaultConfig Default
 	return v, initErr(err, "getting configuration")
 }
 
-func newLocalSousConfig(pic PossiblyInvalidConfig, ls *logging.LogSet, notReallyButPsyringeSequencingDep LogSink) (v LocalSousConfig, err error) {
+func newLocalSousConfig(pic PossiblyInvalidConfig) (v LocalSousConfig, err error) {
 	v.Config, err = pic.Config, pic.Validate()
```
nyarly: ...because this validation is still in place...

main.go
```diff
@@ -25,13 +25,21 @@ func main() {
...
-	ls := logging.NewLogSet(Sous.Version, "sous", "", os.Stderr)
-	defer ls.AtExit()
-	di := graph.BuildGraph(Sous.Version, ls, os.Stdin, os.Stdout, os.Stderr)
-	c, err := cli.NewSousCLI(di, Sous, ls, os.Stdout, os.Stderr)
+	di := graph.BuildGraph(Sous.Version, os.Stdin, os.Stdout, os.Stderr)
+	type logSetScoop struct {
+		*logging.LogSet
+	}
+	lss := &logSetScoop{}
+	di.MustInject(lss)
```
nyarly: ...and is triggered here, before we even construct the CLI, much less invoke `sous config`.

graph/graph.go
```diff
@@ -311,10 +312,24 @@ func newRegistryDumper(r sous.Registry) *sous.RegistryDumper {
...
+	if config.Logging.Basic.DisableConsole {
```
nyarly: As discussed, DisableConsole does something completely different (although aparently not intuitive or well documented) c.f. logSet.Configure

samsalisbury:
Thanks for the review, yeah I can see that 'sous config' regression now. I really want to make the graph a bit simpler, which I think this PR does to a small degree. I think the smallest change to this PR will be to allow possibly invalid config into the LogSet constructor and to validate it/provide defaults there (and print a warning to stderr if not valid), as well as not using `DisableConsole` in this way. I'm not against using `Flaws` for the config, but I think this suggested change is smaller for the same effect for now.

samsalisbury:
That test failure is because a new version of [staticcheck](honnef.co/go/tools/cmd/staticcheck) has a new rule:

    lib/manifestkind.go:15:1: only the first constant has an explicit type (SA9004)

Annoying, I like that brevity of specifying the type only once per `const` block, and I haven't seen it cause issues before. CC @nyarly

samsalisbury:
Looks like a real thing, I'll fix it, see: https://github.com/dominikh/go-tools/blob/3948fdd1819b44ad40a265f8f86e67a4c7c8425a/cmd/staticcheck/docs/checks/SA9004

samsalisbury:
And just verified it's true with... https://play.golang.org/p/kiSS7iA3lj

samsalisbury:
Those coverage reports look totally bogus, choosing to ignore.

main.go
```diff
@@ -20,29 +20,33 @@ var Sous = &cli.Sous{
...
-	if err != nil {
-		die(err)
+	di := graph.BuildGraph(Sous.Version, os.Stdin, os.Stdout, os.Stderr)
+	type logSetScoop struct {
+		*logging.LogSet
 	}
+	lss := &logSetScoop{}
+	if err := di.Inject(lss); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		return 70
```
SegFaultAX: I'm always wary of having magic numbers embedded in code like this. Can you create a constant called `ERR_FAILED_LOGGING_CONFIGURATION` or whatever to clearly indicate what this is supposed to mean?

main.go
```diff
@@ -20,29 +20,33 @@ var Sous = &cli.Sous{
...
 
-	result := c.Invoke(os.Args)
-	exitCode := result.ExitCode()
-
-	os.Exit(exitCode)
-}
+	c, err := cli.NewSousCLI(di, Sous, os.Stdout, os.Stderr)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		return 70
```
SegFaultAX: Same as above.

main.go
```diff
@@ -20,29 +20,33 @@ var Sous = &cli.Sous{
...
-	di := graph.BuildGraph(Sous.Version, ls, os.Stdin, os.Stdout, os.Stderr)
-	c, err := cli.NewSousCLI(di, Sous, ls, os.Stdout, os.Stderr)
-	if err != nil {
-		die(err)
+	di := graph.BuildGraph(Sous.Version, os.Stdin, os.Stdout, os.Stderr)
+	type logSetScoop struct {
+		*logging.LogSet
 	}
+	lss := &logSetScoop{}
+	if err := di.Inject(lss); err != nil {
```
SegFaultAX: Am I correctly understanding that this line is the main thrust of the change? Specifically, that it moves the `LogSet` initialization into the graph, then forces the `LogSet` to be initialized extremely early (in this case, even before the CLI has been created)?

main.go
samsalisbury: Hey, thanks for the review. Yeah I agree, I copied this over uncritically from #510 as that PR contained ideas needed here, but was not easily mergeable. I'll address this and then get it merged tomorrow morning.

main.go
samsalisbury: Yes, this part allows valid logging configuration to configure early logging (before you were stuck only with defaults for logging prior to config reading). There is another important part which makes LogSet constructor depend on unvalidated (aka "PossiblyInvalid") config. That is necessary because we must log something in spite of bad config. Also see the rest of #510 on which the latter half of this PR was based.

SegFaultAX: Ah, I see, yea. Thank you!

main.go
SegFaultAX: Cool. I signed off on the rest of it, so feel free to merge this down once you clean that up. Thank you!

PR #510, opened 2017-11-11
Very elementary changes to clean up output

nyarly:
We need something like #509 so that we can get early logging,
but see comments there.

In the meantime, this change simply increases the unconfigured log level to critical.
Once the config loads, it'll move down to whatever is configured.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,30 +19,26 @@ var Sous = &cli.Sous{
 }
 
 func main() {
+	exitCode := action()
+	os.Exit(exitCode)
+}
+
+func action() int {
 	defer handlePanic()
-	// Eventually, these should become flags on the top level application
-	//logging.Log.Info.SetOutput(os.Stderr)
-	//logging.Log.Debug.SetOutput(os.Stderr)
 	log.SetFlags(log.Flags() | log.Lshortfile)
 
 	ls := logging.NewLogSet(Sous.Version, "sous", "", os.Stderr)
 	defer ls.AtExit()
+
 	di := graph.BuildGraph(Sous.Version, ls, os.Stdin, os.Stdout, os.Stderr)
 	c, err := cli.NewSousCLI(di, Sous, ls, os.Stdout, os.Stderr)
 	if err != nil {
-		die(err)
+		fmt.Fprintln(os.Stderr, err)
+		return 70
 	}
 
 	result := c.Invoke(os.Args)
-	exitCode := result.ExitCode()
-
-	os.Exit(exitCode)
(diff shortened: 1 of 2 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

main.go
```diff
@@ -48,6 +48,10 @@ func handlePanic() {
 	if os.Getenv("DEBUG") == "YES" {
 		return
 	}
+	if pmsg := recover(); pmsg == nil {
```
samsalisbury: The reason we didn't recover here is because we still want to panic overall, and print out the usual stack trace etc. The idea is to add that OOPS banner to the top of the stack trace so that users know where to report it. This `recover` swallows it up.

main.go
```diff
@@ -19,30 +19,26 @@ var Sous = &cli.Sous{
 }
 
 func main() {
+	exitCode := action()
+	os.Exit(exitCode)
+}
```
samsalisbury: This change makes sense now we rely on `ls.AtExit`. Just for background, the way this worked was that it relied on `os.Exit` to prevent defers from firing, and thus to avoid printing the `OOPS` banner unless it panicked (and thus `os.Exit` was never called.

main.go
samsalisbury: I think we need to move `defer HandlePanic` back to `main()` and remove this line, that way we can rely on `os.Exit` preventing this defer from firing again.

util/logging/logging.go
```diff
@@ -91,7 +91,7 @@ func NewLogSet(version semv.Version, name string, role string, err io.Writer) *L
 
 	bundle := newdb(version, err, lgrs)
 
-	ls := newls(name, role, DebugLevel, bundle) //level constrains Kafka output
+	ls := newls(name, role, CriticalLevel, bundle) //level constrains Kafka output
```
samsalisbury: Thanks, this was the bit I was looking for. I was surprised that setting `Logging.Basic.Level: critical` in configuration did not have the effect, so I thought things were more broken than they now appear to be. I'm still not 100% sure why setting that in config didn't work, but glad to see this change resolves the situation and now altering config has the desired effect.

samsalisbury:
Important changes from here incorporated into #509, and panic banner behaviour restored so closing this in favour of that.

PR #511, opened 2017-11-15
Sous config test #trivial

samsalisbury:
This will prevent us breaking the ability of 'sous config' to fix broken config again.

PR #512, opened 2017-11-15
flags: -s, -q, -v, -d now set verbosity correctly

samsalisbury:
CLI flags take precedence over config file options, so the flags should now work as expected.

*Update* now safe to merge.

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,8 +34,11 @@ func main() {
 // Sous fails (e.g. when we can't set up logging or other elementary issues).
 const InitializationFailedExitCode = 70
 
-func guessVerbosity(cliArgs []string) *config.Verbosity {
-	var v, d bool
+func sniffVerbosity(cliArgs []string) graph.VerbosityOverride {
+	var s, q, v, d bool
+	// We want to give higher verbosity precedence in the case more than
+	// one flag is set, so first check for them all then return the most
+	// verbose one detected.
 	for _, a := range cliArgs {
 		if a == "-d" {
 			d = true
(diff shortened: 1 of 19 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

samsalisbury:
This now includes #511 (closed) and adds a fix for it.

main.go
```diff
@@ -33,32 +37,91 @@ func main() {
...
 		return InitializationFailedExitCode
 	}
-	c.LogSink = lss.LogSet
+	// This LogSink is replaced after the Parsed event fires on the CLI.
+	c.LogSink = preParseLogSet
 
 	return c.Invoke(os.Args).ExitCode()
 }
 
+func earlyLoggingVerbosityOverride(cliArgs []string) graph.VerbosityOverride {
```
SegFaultAX: This feels like a real hack to me. If I've understood the content of this PR correctly, the issue is that the verbosity command line parameters aren't taking affect early enough in Sous startup so that some work may be done without properly initializing the logging loglevel to the desired level. This function partially parses the command line looking for one of the available verbosity flags and, if it finds one, injects a `VerbosityOverridden` object into the graph so that the logging config's loglevel is correctly determined when we force it's initialization above in `getLogSet`.

So assuming I've got that right, the thing I take issue with is using this parser to process the command line arguments. It feels to me like there are too many places where the graph is being mutated. And it feels doubly weird to have random places in the code arbitrarily mutate the state of the graph based on their own private semantics of parsing the command line options. Ideally, there would be exactly 1 place where all command line parsing is done, and very few places where the graph is updated (at least during initialization). This spreads out the surface area of both command line parsing and graph initialization in a way that feels, well, like a hack.

Does that make sense?

PR #514, opened 2017-11-23
Fixing log and metrics shipping.

nyarly:


sous-machine-account:
LGTM

</p>

config/verbosity.go
```diff
@@ -7,40 +7,19 @@ type Verbosity struct {
...
-			return &Verbosity{Silent: true}
-		}
-		return &Verbosity{Quiet: true}
-	case "debug":
-		return &Verbosity{Debug: true}
-	case "extradebug1":
-		return &Verbosity{Loud: true}
+		set.BeChatty()
+	case v.Debug:
+		set.BeHelpful()
```
samsalisbury: This is a general point not directly related to this PR, but I think we should clearly define what "-d debug" vs "-v verbose" means. To my mind, "verbose" should mean info-level messages related to the user action that produced them. Verbose should be useful to the user in understanding what process has happened to their stuff as a result of their own actions, and what if any mistakes they have made, and should produce a fairly small, informative story that the sous operator can follow and feel good about understanding clearly.

-d debug should be a finer grained level of detail that might be useful to the Sous team in understanding problems with Sous itself, but is not something we would generally expect users to want. All the same, -d debug logs should be filtered to be specific to the user action that produced them (i.e. scoped to -repo -offset -cluster -flavor context, whichever are appropriate for the command).

Maybe we should add a -dd "double debug" flag to spit out non-contextualised logs (like we do now) should they be a useful thing we want to keep.

util/logging/logging.go
```diff
@@ -334,39 +336,54 @@ func (ls LogSet) Warnf(f string, as ...interface{}) {
...
+// it also sets console output to Discard
+//.Note that as implemented, console output cannot be recovered -
+// it's assumed that BeSilent will be called once per execution.
+func (ls *LogSet) BeSilent() {
 	ls.level = 0
 	ls.imposeLevel()
+	ls.dumpBundle.err = ioutil.Discard
+}
+
+// BeQuiet gets the LogSet to discard all its output
```
samsalisbury: I think "BeQuiet" and "BeSilent" names should be reversed :) (silent is quieter than quiet right?)

samsalisbury: Or maybe it's just the comment that needs fixing...

util/logging/logging.go
```diff
@@ -334,39 +336,54 @@ func (ls LogSet) Warnf(f string, as ...interface{}) {
...
-	ls.level = 1
+func (ls *LogSet) BeTerse() {
+	ls.level = WarningLevel
 	ls.imposeLevel()
 }
 
 // BeHelpful gets the LogSet to print debugging output
-func (ls LogSet) BeHelpful() {
-	ls.level = 2
+func (ls *LogSet) BeHelpful() {
```
samsalisbury: Be helpful to whom? I know I'm being picky, but prefer to think in terms of the CLI operator rather than us sous developers. I don't think our users get much value from our debug logs right now as they're too verbose.

PR #516, opened 2017-11-28
Enable profiling environment variable.

nyarly:
The profiling boolean was getting injected by value before checking the environment.
Fixed that, and also extracted Server as an Action while I was there.

I've tested the profiling configuration locally. If we can get this into production ASAP,
we can start experimenting with the profiling output to see if it tells us anything
about the rectification cycle.


sous-machine-account:
LGTM

</p>

PR #520, opened 2017-12-07
Refactor ensureGDMExists tests

JC1738:
There appear to be 5 different tests run on the ensureGDMExists method.  Three of them dealt with a path and a git repo.  So consolidated those in one test case that has a setup loop through test cases and teardown.  Then the other two seemed unrelated and didn't need setup and teardown.

### DIFF
--- a/cli/actions/server_test.go
+++ b/cli/actions/server_test.go
@@ -1,6 +1,7 @@
 package actions
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -8,8 +9,17 @@ import (
 	"testing"
 )
 
-// TODO: This would be better as separate tests for each case.
 func TestEnsureGDMExists(t *testing.T) {
+	testCases := []struct {
+		path       string
+		createFile bool
+	}{
+		{"does-not-exist", false},
+		{"empty-source-location", false},
+		{"nonempty-source-location", true},
+	}
+
+	//Setup Test Data
 	testDataDir := "testdata/gen/test-ensure-gdm-exists"
 
 	if err := os.RemoveAll(testDataDir); err != nil {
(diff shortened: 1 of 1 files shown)


sous-machine-account:
LGTM

</p>

samsalisbury:
Thanks @JC1738 reviewing now. If possible could you use a separate branch per PR over in your fork? That makes it super convenient for others to check out your code locally. I just added your fork as a remote, and after 'git remote update' which shows me work in other forks I could usually just spot the branch name and 'git checkout branchname' locally then. Also makes it easier to add commits to the PR based on review comments, thanks.

samsalisbury:
Might also be worth cleaning up some of your local branches with `git branch -d $(git branch --merged master | grep -v master)` some of those on your fork look ancient..

samsalisbury:
Those stale branches are of course our fault for not tidying up... I've just deleted about 40 stale and ancient branches, we now just have these... https://github.com/opentable/sous/branches/all

cli/actions/server_test.go
```diff
@@ -60,43 +59,47 @@ func TestEnsureGDMExists(t *testing.T) {
...
 	}
-	// GDM source repo is not a repo.
-	{
-		err := ensureGDMExists(someFile, "", t.Logf)
-		if err == nil {
-			t.Error("expected error when GDM repo does not exist")
-		}
+}
+
+func TestGDMNotARepo(t *testing.T) {
```
samsalisbury: I usually name tests similarly to how [example tests](https://blog.golang.org/examples) have to be named, as it makes them easier to find, especially when using the `-run` flag on `go test` to run a subset of tests. In this case a good name might be `TestEnsureGDMExists_notARepo`

cli/actions/server_test.go
```diff
@@ -60,43 +59,47 @@ func TestEnsureGDMExists(t *testing.T) {
...
+	testDataDir := "testdata/gen/test-ensure-gdm-exists"
+	someFile := path.Join(testDataDir, "a-file")
+
+	err := ensureGDMExists(someFile, "", t.Logf)
+	if err == nil {
+		t.Error("expected error when GDM repo does not exist")
 	}
+}
 
+func TestEnsureNoGDM(t *testing.T) {
```
samsalisbury: Similarly I would probably go for `TestEnsureGDMExists_pathDoesNotExist` or similar.

cli/actions/server_test.go
```diff
@@ -60,43 +59,47 @@ func TestEnsureGDMExists(t *testing.T) {
...
-	}
+	//Test
+	for _, testCase := range testCases {
+		t.Run(fmt.Sprintf("path: %s", testCase.path), func(t *testing.T) {
+			testPath := path.Join(testDataDir, testCase.path)
+			if err := os.MkdirAll(testPath, 0777); err != nil {
+				t.Fatal(err)
+			}
 
-	// SourceLocation dir exists and is empty.
```
samsalisbury: One thing I miss in this PR is these descriptions of the situation being modelled in the test. Having those separate test cases split out definitely helps, but may be worth keeping some of this descriptiveness somewhere?

PR #528, opened 2017-12-12
#trivial Split diff deploy phase 1: Single channel handles DeployablePairs.

samsalisbury:
I'll add commentary on individual lines.


ext/singularity/deployer.go
```diff
@@ -79,7 +79,7 @@ func stripDeployID(in string) string {
 }
 
 // NewDeployer creates a new Singularity-based sous.Deployer.
-func NewDeployer(c rectificationClient, ls logging.LogSink, options ...DeployerOption) *deployer {
+func NewDeployer(c rectificationClient, ls logging.LogSink, options ...DeployerOption) sous.Deployer {
```
samsalisbury: This has been complaining for a while, I would be happier exporting `singularity.deployer` but this got the linter off my back for now. There is now a type assertion in a test because of this, nothing else had to change.

ext/singularity/deployer.go
```diff
@@ -93,27 +93,65 @@ func OptMaxHTTPReqsPerServer(n int) DeployerOption {
 	return func(d *deployer) { d.ReqsPerServer = n }
 }
 
-// RectifyCreates implements sous.Deployer on deployer
-func (r *deployer) RectifyCreates(cc <-chan *sous.DeployablePair, errs chan<- sous.DiffResolution) {
+// Rectify implements sous.Deployer on deployer
+func (r *deployer) Rectify(cc <-chan *sous.DeployablePair, errs chan<- sous.DiffResolution) {
```
samsalisbury: This method is ripe for various tidy-ups. It's copypasta from the old `RectifyXXX` methods wrapped in a big switch.

ext/singularity/deployer.go
```diff
@@ -93,27 +93,65 @@ func OptMaxHTTPReqsPerServer(n int) DeployerOption {
 	return func(d *deployer) { d.ReqsPerServer = n }
 }
 
-// RectifyCreates implements sous.Deployer on deployer
-func (r *deployer) RectifyCreates(cc <-chan *sous.DeployablePair, errs chan<- sous.DiffResolution) {
```
samsalisbury: Copied into new `deployer.Rectify` func.

ext/singularity/deployer.go
```diff
@@ -155,20 +193,6 @@ func (r *deployer) RectifySingleCreate(d *sous.DeployablePair) (err error) {
 	return r.Client.Deploy(*d.Post, reqID)
 }
 
-func (r *deployer) RectifyDeletes(dc <-chan *sous.DeployablePair, errs chan<- sous.DiffResolution) {
```
samsalisbury: Copied into new `deployer.Rectify` func.

ext/singularity/deployer.go
```diff
@@ -185,29 +209,6 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
 	//return r.Client.DeleteRequest(d.Cluster.BaseURL, requestID, "deleting request for removed manifest")
 }
 
-func (r *deployer) RectifyModifies(
```
samsalisbury: Copied into new `deployer.Rectify` func.

ext/singularity/deployer_test.go
```diff
@@ -607,34 +607,34 @@ func TestModificationOfFailed(t *testing.T) {
...
 		},
-		Prior: &sous.Deployable{
+		Post: &sous.Deployable{
 			BuildArtifact: &sous.BuildArtifact{
 				Name: "build-artifact",
 				Type: "docker",
 			},
 			Deployment: dpl.Clone(),
-			Status:     sous.DeployStatusActive,
+			Status:     sous.DeployStatusFailed,
```
samsalisbury: Just reordered Prior/Post assignments.

ext/singularity/deployer_test.go
```diff
@@ -607,34 +607,34 @@ func TestModificationOfFailed(t *testing.T) {
...
 	rezCh := make(chan sous.DiffResolution)
 
-	go deployer.RectifyModifies(dpCh, rezCh)
+	go deployer.Rectify(dpCh, rezCh)
 	dpCh <- dp
 	close(dpCh)
 
 	rez := <-rezCh
 
-	assert.Equal(t, rez.Desc, sous.ModifyDiff)
```
samsalisbury: This was out-of-order actual/expected resulting in confusing test results. Man I am not a fan of this assert library, it's super easy to get bogus actual/expected ordering, and the library isn't even internally consistent. E.g. `assert.Equal` wants `(expected, actual)` but `assert.Len` wants `(actual slice, expected len)`, For me it always make tests just a little harder to understand than using plain Go.

ext/singularity/deployer_test.go
```diff
@@ -644,15 +644,15 @@ func TestModificationOfFailed(t *testing.T) {
 }
 
 func TestOptMaxHTTPReqsPerServer(t *testing.T) {
-	deployer := NewDeployer(nil, logging.SilentLogSet())
-	if deployer.ReqsPerServer != DefaultMaxHTTPConcurrencyPerServer {
+	d := NewDeployer(nil, logging.SilentLogSet()).(*deployer)
```
samsalisbury: This cast was the only impact of `singularity.NewDeployer` returning `sous.Deployer` instead of `singularity.deployer`.

ext/singularity/deployer_test.go
```diff
@@ -644,15 +644,15 @@ func TestModificationOfFailed(t *testing.T) {
...
+	if d.ReqsPerServer != DefaultMaxHTTPConcurrencyPerServer {
 		t.Fatal("Not using default")
 	}
 	const x = 512
 	if x == DefaultMaxHTTPConcurrencyPerServer {
 		t.Fatal("Bad test, please specify a non-default value.")
 	}
-	deployer2 := NewDeployer(nil, logging.SilentLogSet(), OptMaxHTTPReqsPerServer(x))
+	deployer2 := NewDeployer(nil, logging.SilentLogSet(), OptMaxHTTPReqsPerServer(x)).(*deployer)
```
samsalisbury: And this one...

lib/deployable_chans.go
```diff
@@ -8,40 +8,16 @@ type (
 	// A DeployableChans is a bundle of channels describing actions to take on a
 	// cluster
 	DeployableChans struct {
-		Start, Stop, Stable, Update chan *DeployablePair
```
samsalisbury: This is where it all began...

lib/deployable_chans.go
```diff
@@ -8,40 +8,16 @@ type (
...
-		Errs                        chan *DiffResolution
+		Pairs chan *DeployablePair
```
samsalisbury: I mean here...

lib/deployable_chans.go
```diff
@@ -8,40 +8,16 @@ type (
...
-		name         DeploymentID
-		ExecutorData interface{}
-	}
-
-	// DeployablePairKind describes the disposition of a DeployablePair
-	DeployablePairKind int
-
 	diffSet struct {
-		New, Gone, Same, Changed DeployablePairs
+		Pairs DeployablePairs
```
samsalisbury: We can lose `diffSet` now.

lib/deployable_chans.go
```diff
@@ -8,40 +8,16 @@ type (
...
-	// SameKind prior is unchanged from post
-	SameKind DeployablePairKind = iota
-	// AddedKind means an added deployable - there's no prior
-	AddedKind
-	// RemovedKind means a removed deployable - no post
-	RemovedKind
-	// ModifiedKind means modified deployable - post and prior are different
-	ModifiedKind
-)
-
```
samsalisbury: This has been moved to `lib/deployable_pair.go`

lib/deployable_chans.go
```diff
@@ -50,60 +26,41 @@ func NewDeployableChans(size ...int) *DeployableChans {
...
 
+func (ds diffSet) Filter(predicate func(*DeployablePair) bool) []*DeployablePair {
+	var result []*DeployablePair
+	for _, dp := range ds.Pairs {
+		if predicate(dp) {
+			result = append(result, dp)
+		}
+	}
+	return result
+}
```
samsalisbury: This method is just for test support at the moment so may move it to a `_test` file.

lib/deployable_chans.go
```diff
@@ -50,60 +26,41 @@ func NewDeployableChans(size ...int) *DeployableChans {
...
-	case dp.Prior == nil:
-		return AddedKind
-	case dp.Post == nil:
-		return RemovedKind
-	case len(dp.Diffs) == 0:
-		return SameKind
-	default:
-		return ModifiedKind
-	}
-}
```
samsalisbury: Moved to `lib/deployable_pair.go`

lib/deployable_chans.go
```diff
@@ -118,3 +75,20 @@ func (kind DeployablePairKind) String() string {
...
+	case SameKind:
+		return "take no action"
+	case AddedKind:
+		return "create new deployment"
+	case RemovedKind:
+		return "delete existing deployment"
+	case ModifiedKind:
+		return "update existing deployment"
+	}
+}
```
samsalisbury: This method drys up some of our textual logging. Those logs still need to be updated to structured...

lib/deployable_chans_test.go
```diff
@@ -151,10 +151,11 @@ func (nrs *NameResolveTestSuite) TestResolveNameStartChannelUnresolved() {
 func (nrs *NameResolveTestSuite) TestResolveNameStopChannelUnresolved() {
 	nrs.reg.FeedArtifact(nil, fmt.Errorf("not found"))
 	nrs.depChans = nrs.diffChans.ResolveNames(context.Background(), nrs.reg)
-	nrs.diffChans.Stop <- nrs.makeTestDepPair(nrs.makeTestDep(), nil)
+	nrs.diffChans.Pairs <- nrs.makeTestDepPair(nrs.makeTestDep(), nil)
 
 	select {
-	case stopped := <-nrs.depChans.Stop:
+	case stopped := <-nrs.depChans.Pairs:
+		nrs.True(stopped.Kind() == RemovedKind, "got %s; want RemovedKind", stopped.Kind)
```
samsalisbury: This assertion didn't used to be necessary but now we can't be sure of kind based on channel.

samsalisbury: There may be other places that should do this.

lib/deployer.go
```diff
@@ -26,11 +24,7 @@ func (dd *DummyDeployer) RunningDeployments(reg Registry, from Clusters) (Deploy
...
-func (dd *DummyDeployer) RectifyCreates(<-chan *DeployablePair, chan<- DiffResolution) {}
+// Rectify implements Deployer
+func (dd *DummyDeployer) Rectify(<-chan *DeployablePair, chan<- DiffResolution) {
 
-// RectifyDeletes implements Deployer
-func (dd *DummyDeployer) RectifyDeletes(<-chan *DeployablePair, chan<- DiffResolution) {}
-
-// RectifyModifies implements Deployer
-func (dd *DummyDeployer) RectifyModifies(<-chan *DeployablePair, chan<- DiffResolution) {}
+}
```
samsalisbury: Random whitespace.

lib/deployment_diff.go
```diff
@@ -25,12 +24,16 @@ type (
 	}
 )
 
+// Diffs returns the diffs in this pair, from prior to post.
+// TODO: Cache the result.
+func (dp *DeploymentPair) Diffs() Differences {
+	_, differences := dp.Prior.Diff(dp.Post)
+	return differences
+}
```
samsalisbury: I'll add a patch to cache this.

lib/deployment_diff.go
```diff
@@ -41,13 +44,15 @@ func (dp *DeploymentPair) ID() DeploymentID {
...
-	}(difr, other)
-
-	return difr.DeployableChans
+	differ := newStateDiffer(d)
+	go func(differ *stateDiffer, other Deployments) {
+		// Promote "other" (i.e. intended) to "Active".
+		// We always intend "Active".
+		activeDeployments := other.promoteAll(DeployStatusActive)
+		differ.diff(activeDeployments)
+	}(differ, other)
```
samsalisbury: Renames & comment for my own benefit.

lib/deployment_diff.go
```diff
@@ -86,68 +91,51 @@ func newDiffer(intended Deployments) *differ {
...
+	}
+	prior := &Deployable{
+		Deployment: &existingDS.Deployment,
+		Status:     existingDS.Status,
+	}
+
+	return &DeployablePair{
+		name:         id,
+		ExecutorData: executorData,
+		Prior:        post,
```
samsalisbury: ???

lib/deployment_diff.go
```diff
@@ -86,68 +91,51 @@ func newDiffer(intended Deployments) *differ {
...
+		Post:         prior,
```
samsalisbury: ??? there is something funny going on with naming in this file, I think at some point wires get crossed resulting in this oddity. I'll try and make that more consistent in a follow-up patch.

lib/deployment_diff.go
```diff
@@ -86,68 +91,51 @@ func newDiffer(intended Deployments) *differ {
...
-				Status:     intendDS.Status,
-			},
-			Post: &Deployable{
-				Deployment: &existingDS.Deployment,
-				Status:     existingDS.Status,
-			},
-		}
 	}
 
+	// Deployable pairs for deleted deployments.refcitifation
```
samsalisbury: Ah, refcitifation, yes.

samsalisbury: This file is ripe for refactoring as well I think.

lib/loggingprocessor_test.go
```diff
@@ -99,9 +98,8 @@ func TestDiffMessages(t *testing.T) {
 	logging.AssertMessageFields(t, msg, logging.StandardVariableFields, fields)
 
 	msg.pair.Post.Deployment.SourceID.Version = semv.MustParse("0.0.2")
-	msg.pair.Diffs = Differences{"version not the same (this is the test message)"}
 
-	fields["sous-deployment-diffs"] = "version not the same (this is the test message)"
+	fields["sous-deployment-diffs"] = "source id; this: \"github.com/opentable/example,0.0.1\"; other: \"github.com/opentable/example,0.0.2\""
```
samsalisbury: We can no longer provide any old diff, unfortunately this makes the test more brittle, but I'm not hugely worried.

lib/loggingprocessor_test.go
```diff
@@ -111,7 +109,6 @@ func TestDiffMessages(t *testing.T) {
 func TestDiffMessages_knownpanic(t *testing.T) {
 	msg := &deployableMessage{
 		pair: &DeployablePair{
-			Diffs: Differences{},
```
samsalisbury: Now a method.

lib/nameresolver.go
```diff
@@ -19,73 +19,47 @@ func (d *DeployableChans) ResolveNames(ctx context.Context, r Registry) *Deploya
...
+	case AddedKind, ModifiedKind:
+		logging.Log.Vomit.Printf("%s deployment processed, needs artifact: %#v", dp.Kind(), intended)
+		if err := newImageNameResolution; err != nil {
+			logging.Log.Info.Printf("Unable to %s %q: %s", action, intended.ID(), err)
+			logging.Log.Debug.Printf("Failed to %s %q: % #v", action, intended.ID(), intended)
+			return nil, err
+		}
+		if newImageName == nil {
+			logging.Log.Info.Printf("Unable to %s deployment %q: no artifact for SourceID %q", action, intended.ID(), intended.SourceID)
+			logging.Log.Debug.Printf("Failed to %s %q: % #v", action, intended.ID(), intended)
```
samsalisbury: All the above logging needs structuring up.

samsalisbury: Separate PR for that I think.

lib/nameresolver.go
```diff
@@ -19,73 +19,47 @@ func (d *DeployableChans) ResolveNames(ctx context.Context, r Registry) *Deploya
...
-	return da
+	return &DeployablePair{ExecutorData: dp.ExecutorData, name: dp.name, Prior: dp.Prior, Post: newImageName}, nil
 }
 
 func resolveName(r Registry, d *Deployable) (*Deployable, *DiffResolution) {
+	if d == nil {
+		return nil, &DiffResolution{
+			Error: &ErrorWrapper{error: fmt.Errorf("nil deployable")},
+		}
+		//panic("NIL Deployable")
```
samsalisbury: Should have removed this line...

ext/singularity/deployer_test.go
nyarly: I would absolutely prefer a fluent assertion interface (assertThat(actual).equals(expected)) but there's two things I strongly prefer about using some assertion library: the intent of the test becomes clearer faster (oh, Sam is saying that slice should have 5 elements), the written code is about the test, rather than the mechanics of the test ("these are equal", not "I compare them for equality, and if they aren't equal let me format the error output"). There's also a ton of utility to the actual output - calling out the context and reporting diffs etc is very valuable to me.

Granted, the "assertEqual(one, theother)" leads to these out of order issues happen super frequently, and I share your frustration over them.

lib/deployment_diff.go
nyarly: The diffing logic *in general* has been kind of screwy. At the root, there are `(one Type) Diff(other Type) TypePair` definitions that report changes as if other is the base (i.e. features of other appear as `prior`. I'm not happy with any of that, but hunting it down has seemed onerous, and getting it wrong is fiddly and catastrophic when it there's mistakes.

If you want to tackle correcting the sense of the diff vector, I'd be all in, but I think it needs two careful, well-informed painstaking pairs of eyes on it.

lib/deployment_diff.go
nyarly: To be fair, it does seem like "refcitifation" is what we do :)

lib/deployment_diff.go
```diff
@@ -167,34 +155,32 @@ func (d *differ) diff(existing Deployments) {
...
+			d.Pairs <- &DeployablePair{
 				name:  id,
 				Prior: nil,
 				Post:  &Deployable{Deployment: existingDeployment, Status: DeployStatusActive},
 			}
 			continue
 		}
 		delete(d.from, id)
-		different, differences := existingDeployment.Diff(&intendedDeployment.Deployment)
+		different, _ := existingDeployment.Diff(&intendedDeployment.Deployment)
```
nyarly: I think, unless we're going to write a real `IsDifferent` method (i.e. early return in the event of a difference), we should start factoring out the ([]diffs, bool) returns from things.

lib/nameresolver.go
nyarly: Makes sense. I'd come behind and do it, but it seems good to ensure that I'm not the only one who understands how structured logging works.

lib/nameresolver.go
```diff
@@ -19,73 +19,47 @@ func (d *DeployableChans) ResolveNames(ctx context.Context, r Registry) *Deploya
...
+func (names *nameResolver) HandlePairs(dp *DeployablePair) (*DeployablePair, *DiffResolution) {
+	intended := dp.Post
+	deployID := dp.ID()
+	action := dp.Kind().ResolveVerb()
+	newImageName, newImageNameResolution := resolveName(names.registry, intended)
+	switch dp.Kind() {
+	default:
+		panic(fmt.Errorf("Unknown kind %v", dp.Kind()))
+
+	case SameKind, RemovedKind:
```
nyarly: This is always going to fail for resolving the Post name of a RemovedKind - since the Post should be nil.

nyarly: resolveName is always going to fail for the `Post` of a `Removed` pair - because it'll be nil.

nyarly: We need to be able to resolve the `Post` of added and changed pairs. 

For logging, it's nice to be able to resolve the `Prior` where they exists and the `Post` for samesies. We might decide not to do this, or only to resolve this if we can resolve from our local cache. (IMO, if we drop this entirely, we're jinxed to have to debug an operational issue related to not knowing the prior image...)

lib/nameresolver.go
```diff
@@ -19,73 +19,47 @@ func (d *DeployableChans) ResolveNames(ctx context.Context, r Registry) *Deploya
...
+			logging.Log.Debug.Printf("Failed to %s %q: % #v", action, intended.ID(), intended)
+			return nil, err
+		}
+		if newImageName == nil {
+			logging.Log.Info.Printf("Unable to %s deployment %q: no artifact for SourceID %q", action, intended.ID(), intended.SourceID)
+			logging.Log.Debug.Printf("Failed to %s %q: % #v", action, intended.ID(), intended)
+			return nil, &DiffResolution{
+				DeploymentID: dp.ID(),
+				Desc:         "not created",
+				Error:        WrapResolveError(errors.Errorf("Unable to create new deployment %q: no artifact for SourceID %q", intended.ID(), intended.SourceID)),
```
nyarly: I seriously wonder about the relationship between log entries and errors here. Arguably, the ResolveError should be well structured and logged once further up.

lib/pipeline.go
```diff
@@ -17,7 +22,7 @@ func (d *DeployableChans) Pipeline(ctx context.Context, proc DeployableProcessor
 	d.Add(1) // for the errs channel
 
 	wg := sync.WaitGroup{}
-	wg.Add(4)
+	wg.Add(1)
```
nyarly: Is there a reason not to `d.Add(1)` for the pairs channel here? Or roll this up with L22 and make it `d.Add(2)`? I can't see a reason not to.

lib/resolver.go
```diff
@@ -37,30 +36,7 @@ func NewResolver(d Deployer, r Registry, rf *ResolveFilter, ls logging.LogSink)
...
-		}
-		if dep.Status == DeployStatusPending {
-			rez.Desc = ComingDiff
-		}
-		if dep.Status == DeployStatusFailed {
-			rez.Error = WrapResolveError(&FailedStatusError{})
-		}
-		results <- rez
-	}
+	d.Rectify(dcs.Pairs, results)
```
nyarly: :+1:

lib/deployment_diff.go
samsalisbury: I agree. Part of the problem may be that there is at least one Diff method (I'll try and find it) that actually assumes that either one or other is post or prior when that method should not make this assumption. I have considered adding labels to the diff methods, for the sake of logging. Right now we log "this X other Y" would be great to log "actual X intended Y" or similar by passing labels e.g. `x.Diff("actual", y, "intended")`.

ext/singularity/deployer_test.go
samsalisbury: Hmm, I can see the attraction of this library, but the trouble is it's not written well enough to be completely trustworthy, and in tests in particular, I don't want to be left second-guessing what the library is doing. I find well-abstracted Go code very quick to parse in my mind. It ends up being longer in LOC but shorter in reading time for me personally, if done right, and I believe code should be optimised for the reader first. Avoiding masses of formatting code is important, and needs to be carefully abstracted away. In fact, using libraries like spew makes a lot of sense to me for this. That posited fluent interface would solve the problem of actual/expected, but still necessitates additional complexity. A generic equality method is very hard to get right, and even the best effort will have edge cases where the apparent simplicity of the invocation is belied by an [intricate reflection-driven dance](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L64-L72) underneath. Even [simple assumptions, (like this surprising one)](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L303) that are made by the library need to be learnt and memorised by both the writer and reader of the code. I would rather not have to learn a testing library author's determination of what equality means, when Go already has a [well-defined equality operator](https://golang.org/ref/spec#Comparison_operators) for simple cases. For more complex cases I would rather be able to see custom equality code nearby (e.g. an `Equal` method) than have to trawl through vendored code. For the sake of 3 lines of code rather than one, I am not yet persuaded that assertion libraries like this are helpful in Go tests.

lib/nameresolver.go
samsalisbury: Yes it will always fail, this was a slightly lazy thing to do but I was in a hurry not to let the broad picture of the refactor slip. In fact, if this changes a log message then I'll need to add a changelog entry as well... Good spot.

Just an FYI this refactor was mostly just mechanical, I tried not to understand why the code was doing what it was doing, and just concentrated on the code, as algebra, working out how to get all the tests passing again after reducing 4 channels to 1. I think in the wake of this PR I could happily spend another full day tying up loose ends and probably deleting a fair amount of redundant code, so that's what I'll plan to do tomorrow, until the big seam we are going for starts to reveal itself more clearly.

samsalisbury: I am not sure which is the best way to go with this. Is knowing the previous image useful at all right now? I do think we will need it if we want to start doing transactional deployments, but I'm not sure of its utility in the logs. My thought is to avoid doing anything unless there is a clear need for it, as it's just more code to maintain. Unless there's a use case I'm not aware of?

lib/pipeline.go
samsalisbury: I can't see a reason not to either... worth tidying up, good spot.

samsalisbury:
>  I could be okay with "only resolve the two cases we must have, put in an issue for log-facing resolves".

I vote for putting this in as an issue; I consider this hack n slash "refactor" I've done slightly risky, so delaying getting it merged/deployed just compounds the risk that any further work on top of it will be wasted.

lib/nameresolver.go
nyarly: Yeah, I'd honestly be happy with "on delete or same, no resolve. On add or modify, resolve only post" which seems like a v small change here.

samsalisbury: Yep I'm happy with that, if you open a PR with this change I'll certainly be happy to merge tomorrow, thanks.

ext/singularity/deployer_test.go
nyarly: Fair enough, so long as you've heard: it's not a LOC concern for me: it is clearer to me, as a reader, to used named assertions than ad hoc conditionals.

PR #531, opened 2017-12-15
Allow config.yaml loading to support overriding siblingings url by re…

JC1738:
…flecting map[string]string from env variable

Also fixed some linting in messages.go


sous-machine-account:
LGTM

</p>

util/configloader/configloader_test.go
```diff
@@ -89,3 +93,24 @@ func TestLoad_EmptyEnv(t *testing.T) {
...
+	if err := cl.Load(&c, "test_map_config.yaml"); err != nil {
+		t.Fatal(err)
+	}
+
+	if _, ok := c.SiblingURLs["env3"]; ok {
+		t.Errorf("Failed to override map from environment variable")
+	}
+
+	if val := c.SiblingURLs["env1"]; val != "foo" {
+		t.Errorf("Value not expected")
```
samsalisbury: In general I always print out the actual and expected values in the error message as well, as it's often helpful in quickly identifying the cause (e.g. typos). But this case is pretty simple so I can live with this.

util/configloader/test_map_config.yaml
```diff
@@ -0,0 +1,4 @@
+SiblingURLs:
+  env1: http://foo.com
+  env2: http://bar.com
+  env3: http://example.com
```
samsalisbury: In general it's a good idea to put data files like this in a `testdata` sub directory, as this is treated specially by the go toolchain and will be explicitly ignored when compiling etc. For one file it makes little difference, but prob worth doing anyway as it's good practice. (See `go help packages`)

JC1738:
@samsalisbury Made changes suggested by review and ready for re-review/merge

samsalisbury:
Thanks @JC1738

PR #532, opened 2017-12-18
Make targets and liquibase scripts

nyarly:
This PR adds make targets that allow for a development database to be set up quickly.

It currently requires that you have Postgres 10 and Liquibase installed (with
the 42.1 version of the Postgres JDBC driver). I'm afraid this requirement will
be onerous for other devs, so I want to check that assumption.

Once you have the required software installed, these commands should work:

`make postgres-start`
This is not only start development Postgres server,
it will also build the requisite data directory,
and use the Liquibase changelog to update its schema.

`make postgres-connect`
Connects psql to the dev database. Just a convenience -
it means you don't have to know or remember the connection parameters.

`make postgres-stop`
Shuts down the development server.


sous-machine-account:
LGTM

</p>

PR #534, opened 2017-12-21
#trivial add location of sous config

JC1738:
add location of config.yaml to sousConfigHelp

### DIFF
--- a/cli/sous_config.go
+++ b/cli/sous_config.go
@@ -14,7 +14,7 @@ type SousConfig struct {
 
 func init() { TopLevelCommands["config"] = &SousConfig{} }
 
-const sousConfigHelp = `view and edit sous configuration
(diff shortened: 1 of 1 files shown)


sous-machine-account:
LGTM

</p>

nyarly:
Strictly speaking, we compute a path for the config file based on environment variables and conventions (https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html) *Usually* that means the config file is at `~/.config/sous/config.yaml` but if it's somewhere different, it'd be good if `sous config` could accurately report that.

I'd be okay merging this with an eye towards refining it in a near-future UX concentrated session.

samsalisbury:
This will be true 99% of the time, so merging for now, but we should make it dynamic, or tell the user how to work it out themselves...

PR #538, opened 2017-12-27
#trivial Fix cli/actions package lint warnings

JC1738:
Clear all messages for actions package

### DIFF
--- a/cli/actions/server.go
+++ b/cli/actions/server.go
@@ -44,20 +44,21 @@ func (ss *Server) Do() error {
 }
 
 func ensureGDMExists(repo, localPath string, log func(string, ...interface{})) error {
-	s, err := os.Stat(localPath)
-	if err == nil && s.IsDir() {
-		files, err := ioutil.ReadDir(localPath)
-		if err != nil {
+
+	if s, err := os.Stat(localPath); err == nil && s.IsDir() {
+		if files, err := ioutil.ReadDir(localPath); err != nil {
 			return err
-		}
-		if len(files) != 0 {
+		} else if len(files) != 0 {
+
 			// The directory exists and is not empty, do nothing.
 			if repo != "" {
 				log("not pulling repo %q; directory already exist and is not empty: %q", repo, localPath)
 			}
+
 			return nil
 		}
 	}
+
 	if err := config.EnsureDirExists(localPath); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


sous-machine-account:
LGTM

</p>

cli/actions/server.go
```diff
@@ -44,20 +44,21 @@ func (ss *Server) Do() error {
...
-	if err == nil && s.IsDir() {
-		files, err := ioutil.ReadDir(localPath)
-		if err != nil {
+
+	if s, err := os.Stat(localPath); err == nil && s.IsDir() {
+		if files, err := ioutil.ReadDir(localPath); err != nil {
 			return err
-		}
-		if len(files) != 0 {
+		} else if len(files) != 0 {
```
samsalisbury: Changes to this file took me a while to understand properly. I'm having trouble getting `:GoMetaLinter` in vim to respect the gometalinter.json file (tried [renaming it to .gometalinter.json](https://github.com/alecthomas/gometalinter#configuration-file) but no dice). Anyway, test pass so I'm happy, just wanted to understand what the linter was complaining about... Me and my beloved shadowed vars :P

cli/actions/update.go
```diff
@@ -67,21 +67,25 @@ func updateRetryLoop(ls logging.LogSink,
...
+
+		if state, err = sm.ReadState(); err != nil {
 			return sous.NewDeployments(), err
 		}
-		manifest, ok := state.Manifests.Get(mid)
-		if !ok {
-			err := fmt.Errorf("No manifest found for %q - try 'sous init' first.", mid)
+
+		if manifest, ok = state.Manifests.Get(mid); !ok {
+			err = fmt.Errorf("no manifest found for %q - try 'sous init' first", mid)
```
nyarly: The reasoning for "errors should be lowercased and unpunctuated" has to do with formatting errors for the user. But here we're using the error directly. I don't know that we can *just* change the formatting of these errors without a corresponding formatting on output.

(on the other hand, "try sous init first" turns out to be confusing a reasonable percentage of the time.)

cli/actions/server.go
nyarly: I'm ambivalent about this change - on the one hand, I'm not a particular fan of the inlined variable assignment in if statements - it's clever, but the construct is so fragile that it often doesn't apply and I wind up wrangling the separate assignment.

On the other hand, the defense of not doing it this way is to preserve the "early return, don't indent" rule, which I'm also lukewarm on.

In the end, if this is the linter's advice, and it'll be consistent, then let's do that.

samsalisbury:
The lack of pre-existing test coverage for this change still makes me nervous. I think either we need to add tests that cover expected behaviour, and make this change in the knowledge that behaviour is unchanged, or else close the PR if it's not high enough priority to warrant that work right now.

PR #539, opened 2018-01-02
Bugfix: Always order owners

samsalisbury:
See [associated changelog entry](https://github.com/opentable/sous/compare/master...samsalisbury:always-order-owners?expand=1#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR15) for details.

### DIFF
--- a/ext/storage/disk_state_manager_test.go
+++ b/ext/storage/disk_state_manager_test.go
@@ -34,6 +34,39 @@ func TestWriteState(t *testing.T) {
 	}
 }
 
+func TestWriteState_out_of_order_owners(t *testing.T) {
+
+	const repo = "github.com/opentable/sous"
+	s := exampleState(func(s *sous.State) {
+		m, ok := s.Manifests.Single(func(m *sous.Manifest) bool {
+			return m.Source.Repo == repo
+		})
+		if !ok {
+			t.Fatalf("no manifest with repo %q found", repo)
+		}
+		// Switch owners around.
+		m.Owners[0], m.Owners[1] = m.Owners[1], m.Owners[0]
+	})
+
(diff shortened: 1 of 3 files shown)


sous-machine-account:
LGTM

</p>

ext/storage/disk_state_manager_test.go
```diff
@@ -79,14 +112,17 @@ func TestReadState_empty(t *testing.T) {
 	}
 }
 
-func exampleState() *sous.State {
+// exampleState produces a canonical state. If you pass one or more modify
+// funcs, each will be applied in order before the state is returned.
```
nyarly: Concerning modify funcs: I think you can just manipulate the State returned. Passing funcs in to do this seems over-complicated.

It doesn't, yet, significantly impact readability, though, so I don't want to block a bugfix on the merge.

nyarly: @samsalisbury Just want your ack on this.

samsalisbury: Hey @nyarly, thanks for the review. My thought was to allow a single expression in the test itself to get the state under test. By passing in a func, we don't need to pollute the top-level scope in the test func with a bunch of state manipulation code, so to me it reads cleaner this way by encapsulating state setup away from other test code. I acknowledge that this is more complex than just manipulating state in the test itself, and in general complexity is worse than simplicity, but it was done for this reason. It could have been encapsulated by any old func wrapping state manipulation, or scoped by an arbitrary code block, but by making this a documented feature of the much-reused `exampleState` func, I hope it encourages that one pattern in other tests rather than each going their own way. There are pros and cons, but I'm happy with this pattern at the moment.

PR #541, opened 2018-01-04
#trivial rectification queue per DeploymentID

samsalisbury:
This large PR adds a job (rectification) queue per DeploymentID. This is an internal change, no external change to behaviour should be observed.

`Resolver.queueDiffs` synchronously waits for each rectification to complete before moving on to the next. This satisfies the integration tests as they stand, but will cause performance issues ~once rectifications are responsible for waiting on tasks becoming healthy~. I will address this along with that change in the next PR.

~**EDIT:** This will cause performance issues right now, as we already wait for the rectification to complete. Follow-up commits coming.~

**EDIT 2:** Ready for review/merge.

Note that the type `R11nQueue` maintains a reference to each rectification it contains in its buffered channel. This unusual implementation was decided on because it allows us to utilise channels for signalling (thus no need to poll our own queue) and will enable a feature in the next story where we want to be able to query the queue to see what it's going to work on next.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 985 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
+package sous
+
+import (
+	"sync"
+
+	"github.com/pborman/uuid"
+)
+
+// R11nQueue is a queue of rectifications.
+type R11nQueue struct {
```
nyarly: Overall, I'm finding the queueing to be very complicated. It seem like R11nQueueSet could be a sync.Map of chans and done. What am I missing?

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	rq.refs[id] = qr
+	rq.queue <- qr
+	return qr
+}
+
+// PushIfEmpty adds an item to the queue if it is empty, and returns the wrapper
+// added and true if successful. If the queue is not empty, or is full, it
+// returns nil, false.
+func (rq *R11nQueue) PushIfEmpty(r *Rectification) (*QueuedR11n, bool) {
```
nyarly: I think this is part of my concern. We only want PushIfEmpty for as long as we're retaining the current diff-based behavior, right? For that duration, a map of Rectifications suffices, with goroutines to delete them.

Once we actually want channels, PushIfEmpty doesn't make sense, and we don't need the refs map with R11nQueue.

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
...
+	rq.Lock()
+	defer rq.Unlock()
+	// We look at refs since we only delete the ref after handling has happened.
+	// If we are busy handling a r11n, then we consider the queue non-empty.
+	if len(rq.refs) != 0 {
```
nyarly: For that matter, couldn't a `select` work here? Have a 1 element channel, and have a `case ch <- r11n` where the default clause returns `nil, false` ?

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
...
+			delete(rq.refs, qr.ID)
+		}
+	}()
+	return results
+}
+
+// Wait waits for a particular rectification to be processed then returns its
+// result. If that rectification is not in this queue, it immediately returns a
+// zero DiffResolution and false.
+func (rq *R11nQueue) Wait(id R11nID) (DiffResolution, bool) {
```
nyarly: This is the other requirement for a map at the R11nQueue level - but the use case is in `resolver.go` and looks like "push, then immediately wait". The calling code has the Rectification to wait on already.

lib/r11n_queue_set.go
```diff
@@ -0,0 +1,45 @@
+package sous
+
+import "sync"
+
+// R11nQueueSet is a concurrency-safe mapping of DeploymentID to R11nQueue.
+type R11nQueueSet struct {
+	set  map[DeploymentID]*R11nQueue
+	opts []R11nQueueOpt
+	sync.RWMutex
```
nyarly: sync.Map?

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	qr := <-rq.queue
+	rq.Lock()
+	defer rq.Unlock()
+	rq.handlePopped(qr.ID)
+	return qr
+}
+
+// handlePopped assumes rq is locked.
+func (rq *R11nQueue) handlePopped(id R11nID) {
```
nyarly: If the position of R11ns is really important, a couple of atomic counters (numberIn, numberOut) on the queue would handle that neatly as well. IIRC, the only place the design requires it is informationally.

lib/resolver.go
```diff
@@ -32,13 +34,42 @@ func NewResolver(d Deployer, r Registry, rf *ResolveFilter, ls logging.LogSink)
...
+				qr.Rectification.Begin(r.Deployer)
+				return qr.Rectification.Wait()
+			}))
+	}
+
+	var wg sync.WaitGroup
 	for p := range dcs.Pairs {
-		sr := NewSingleRectification(*p)
-		results <- sr.Resolve(r.Deployer)
+		p := p
```
nyarly: ?

lib/resolver.go
```diff
@@ -32,13 +34,42 @@ func NewResolver(d Deployer, r Registry, rf *ResolveFilter, ls logging.LogSink)
...
+		sr := NewRectification(*p)
+		queued, ok := globalQueueSet.PushIfEmpty(sr)
+		if !ok {
+			r.ls.Warnf("dropping rectification; queue not empty for %q", sr.Pair.ID())
+			continue
+		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			result, ok := globalQueueSet.Wait(p.ID(), queued.ID)
```
nyarly: e.g. (from above) this could be sr.Wait(), right?

lib/resolvestatus.go
```diff
@@ -102,14 +103,15 @@ func NewResolveRecorder(intended Deployments, f func(*ResolveRecorder)) *Resolve
...
+	// Execute the main function (f) over this resolve recorder.
 	go func() {
 		f(rr)
 		rr.write(func() {
 			rr.status.Finished = time.Now()
 			if rr.err == nil {
 				rr.status.Phase = "finished"
 			}
-			close(rr.Log)
+			//close(rr.Log)
```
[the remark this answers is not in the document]
nyarly: I see - it happens at https://github.com/opentable/sous/pull/541/files#diff-37e220231d21cfe9bbe0ad2f3c03daa3R126 now.

Please delete this line.

lib/r11n_queue.go
samsalisbury: The design as I understand it is that we will always want to rectify diffs based on intended/actual deployments, but only when no diffs instantiated by 'sous deploy' are queued or in progress, so we will always need `PushIfEmpty`.

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
...
+	"sync"
+
+	"github.com/pborman/uuid"
+)
+
+// R11nQueue is a queue of rectifications.
+type R11nQueue struct {
+	cap   int
+	queue chan *QueuedR11n
+	refs  map[R11nID]*QueuedR11n
```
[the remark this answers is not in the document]
samsalisbury: `sync.Map` has specific use cases and treats everything as `interface{}` so I do not believe it would be appropriate here. See https://golang.org/pkg/sync/#Map

lib/r11n_queue.go
samsalisbury: What would be the benefits of using `sync.Map` there? We would no longer have to use a mutex but would have to replace it with lots of type assertions.

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
...
+		cap: R11nQueueCapDefault,
+	}
+	for _, opt := range opts {
+		opt(rq)
+	}
+	return rq.init()
+}
+
+// R11nQueueOpt is an option for configuring an R11nQueue.
+type R11nQueueOpt func(*R11nQueue)
```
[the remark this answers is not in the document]
samsalisbury: I agree we probably don't need this pattern here since this doesn't need to be exported, however it still makes the code more readable so I prefer to leave this in. Yes I wrote the queue as a more generic type than strictly necessary, however I don't see the harm in maintaining it like that since it is fully documented and the general structure could find application elsewhere.

lib/r11n_queue.go
samsalisbury: This is true, in fact this entire PR is geared only to making way for the next story, and its requirements are unusually code-level requirements. So, simply examining the functionality of this PR, it is more complicated than it needs to be (it adds complexity but does not change external behaviour). Please consider this code in light of its stated goals: https://opentable.atlassian.net/browse/SOUS-791 (sorry non-opentablers can't access that link).

lib/r11n_queue.go
samsalisbury: If we use a "1 element channel" (I read this as "channel with buffer length 1") then we will at most be able to queue a single rectification, it is implied by our discussed spec that we may want to queue more than one element, but don't yet know how many exactly which is why that figure is configurable. As the comment here says, we aren't relying on the channel length (current number of items buffered) because the currently in progress rectification still counts as non-empty). Of course, the entire implementation could be different, in which case selects may have been appropriate, but that's not how it ended up, this is a working implementation with an eye towards the next goal.

lib/r11n_queue_set.go
samsalisbury: I do not believe using sync.map it would have any notable benefit in this instance. We would just swap a mutex for type assertions.

lib/resolver.go
```diff
@@ -32,13 +34,42 @@ func NewResolver(d Deployer, r Registry, rf *ResolveFilter, ls logging.LogSink)
 	}
 }
 
-// rectify takes a DiffChans and issues the commands to the infrastructure to
-// reconcile the differences.
-func (r *Resolver) rectify(dcs *DeployableChans, results chan DiffResolution) {
+// globalQueueSet is a temporary solution to having a queue of rectifications
+// per Deployment. This will be replaced with a persistent queue.
```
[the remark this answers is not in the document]
samsalisbury: We discussed putting these into a persistent queue, but I'll remove the comment as it's not necessary to make this statement here.

lib/resolver.go
samsalisbury: This creates an interation-scoped copy of the value of the outer (range-scoped) `p` for use inside the iteration. Note line 65 where we use p inside a goroutine. Removing this line is a race condition ;)

lib/resolver.go
```diff
@@ -32,13 +34,42 @@ func NewResolver(d Deployer, r Registry, rf *ResolveFilter, ls logging.LogSink)
...
+			}))
+	}
+
+	var wg sync.WaitGroup
 	for p := range dcs.Pairs {
-		sr := NewSingleRectification(*p)
-		results <- sr.Resolve(r.Deployer)
+		p := p
+		sr := NewRectification(*p)
+		queued, ok := globalQueueSet.PushIfEmpty(sr)
```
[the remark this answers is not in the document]
samsalisbury: My understanding is that we will always have 'sous depoy' driven deployments and 'actual no longer matches intended' driven deployments (like all deployments are now), the latter should always use the 'push if empty' behaviour. Is this not captured in a ticket? Hmm, we should capture that if not.

lib/resolver.go
```diff
@@ -32,13 +34,42 @@ func NewResolver(d Deployer, r Registry, rf *ResolveFilter, ls logging.LogSink)
...
+		if !ok {
+			r.ls.Warnf("dropping rectification; queue not empty for %q", sr.Pair.ID())
```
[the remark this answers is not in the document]
samsalisbury: I concur, I'll structure this up.

lib/resolver.go
samsalisbury: Of course, this entire PR is moot from the point of behaviour, it just re-implements current behaviour in a more complex way. Since we will eventually (soon) want to be able to long-poll on completion of a rectification based on its ID, best to get part of that mechanism in use now so we know we can rely on it later.

lib/resolver.go
```diff
@@ -32,13 +34,42 @@ func NewResolver(d Deployer, r Registry, rf *ResolveFilter, ls logging.LogSink)
...
+			continue
+		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			result, ok := globalQueueSet.Wait(p.ID(), queued.ID)
+			if !ok {
+				panic(fmt.Sprintf("waiting on non-existent r11n %q", queued.ID))
```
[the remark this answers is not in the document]
samsalisbury: Sure that makes sense.

lib/resolvestatus.go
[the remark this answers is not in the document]
samsalisbury: Of course.

lib/r11n_queue.go
samsalisbury: How would using atomic counters be simpler or more explicit? Queue position is relevant to future UX stories.

lib/resolver.go
samsalisbury: @nyarly see the last code example here: https://golang.org/doc/faq#closures_and_goroutines this is a common pattern in Go, I use it frequently.

lib/resolver.go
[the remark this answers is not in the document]
samsalisbury: I've opened a private PR to add "diff-source" fields to sous-diff-resolution OTL. I'll send link via JIRA.

ext/singularity/deployer.go
```diff
@@ -94,64 +94,64 @@ func OptMaxHTTPReqsPerServer(n int) DeployerOption {
...
+		return result
+	case sous.RemovedKind:
+		result := sous.DiffResolution{DeploymentID: d.ID()}
+		if err := r.RectifySingleDelete(d); err != nil {
+			result.Error = sous.WrapResolveError(&sous.DeleteError{Deployment: d.Prior.Deployment.Clone(), Err: err})
+			result.Desc = "not deleted"
+		} else {
+			result.Desc = sous.DeleteDiff
+		}
+		Log.Vomit.Printf("Reporting result of delete: %#v", result)
```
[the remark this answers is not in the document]
samsalisbury: I agree but I consider making further changes to this file outside the scope of this PR (this file was only changed to reflect the new `Rectify` signature). I'm adding a story to fix logging in `ext/singularity` now, as there is a lot of it that needs updating there.

lib/r11n_queue.go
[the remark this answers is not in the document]
nyarly: YAGNI. The queues don't need *any* configuration except speculatively.

lib/r11n_queue.go
nyarly: I think that's the story for this PR. Which story do you mean?

lib/resolver.go
nyarly: Okay. Now that you explain it, I recognize the problem that it solves. I find this construct very confusing to read. It certainly doesn't expose anything about your intention. Could we instead pass the iteration variable to the goroutine function, a la:
```
for p := range dcs.Pairs {
   // ...
   go func(pair *DeployablePair) {
     // ...   
   }(p)
}
```
?

lib/r11n_queue.go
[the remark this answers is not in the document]
samsalisbury: We could remove it, but later when we want to tweak queue size we'll have to put it back in. The 'StartWithHandler' code makes the operation of the queue explicit and visible at the crucial point of use in `lib/resolver.go`. I'm trying to locate related code together so it can be read in context, and keep that somewhat smelly code wrapping `Rectify` calls isolated.

lib/resolver.go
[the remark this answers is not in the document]
nyarly: The behavior of multiple change requests is tricky, and I think we can to some resolution... I just don't remember what. :) I'd like to be able to iterate on it, but I suspect it's a deep and subtle point of behavior that we may not be able to hill-climb usefully.

lib/r11n_queue.go
samsalisbury: Oh, this one: SOUS-792 (the `{DeployID}/{DeployActionID}` endpoint).

lib/resolver.go
samsalisbury: It's more verbose but I'm happy to make this change if that make it more readable.. update coming.

lib/resolver.go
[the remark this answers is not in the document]
samsalisbury: OK, I thought we had reached consensus on how to approach this for the first pass (what I described above), that's certainly what SOUS-791 says. We can of course iterate on it, but it seems like a reasonable behaviour to start with.

lib/resolver.go
samsalisbury: I did https://github.com/opentable/sous/pull/541/commits/5bc8b4a4647eac0fe5bdb6a7e385f0b9b8150e76 instead, as that seemed even clearer in a way. Making the func explicitly accept the range var to close over it read oddly when all the other local vars declared in the iteration were equally important but just happened to be safe to use in the closure without adding parameters. By capturing just the ID, the `p := p` line no longer exists. However in this case I found the original easier to read, as that odd looking construct can only really mean one thing... Meh, feel free to change it to however looks most readable if that gets it merged :)

lib/resolver.go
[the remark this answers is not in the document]
nyarly: My point being, we should capture that design somewhere so that we're all working toward the same goal.

It's bothering me because I feel like we came to a non-obvious conclusion on this point.

lib/resolver.go
nyarly: Ok. I think after all I'm starting to see why these are more fiddly than they might seem to need to be.

lib/r11n_queue.go
nyarly: The QueuedR11ns wouldn't need to be updated after being put on the queue - they'd get their "butcher's number" when they're created, and their queue position is always their number - the "now serving" number. It'd also decrease the size and duration of critical sections.

lib/r11n_anomaly_message.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"fmt"
+
+	"github.com/opentable/sous/util/logging"
+)
+
+// r11nAnomalyMessage is a specialisation of diffRezMessage, so we embed that
+// anonymously and override Message and DefaultLevel.
+type r11nAnomalyMessage struct {
```
nyarly: :+1:

lib/r11n_queue.go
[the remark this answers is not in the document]
nyarly: Yeah. That's disappointing.

lib/r11n_queue.go
nyarly: Reviewing sync.Map I start to see the objections. That's frustrating.

What I was missing is that you're targeting future functionality (e.g. being able to query deployments and their status.)

lib/r11n_queue.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// Start starts applying handler to each item on the queue in order.
+func (rq *R11nQueue) Start(handler func(*QueuedR11n) DiffResolution) <-chan DiffResolution {
+	results := make(chan DiffResolution, 100)
+	go func() {
+		for {
+			qr := rq.next()
+			results <- handler(qr)
+			delete(rq.refs, qr.ID)
```
nyarly: Based on my current understanding, refs will be used to answer `GET /deploystatus?deployment_id=[...]` right?

Once we get there, we'll need to stop deleting the Rectifications when they're started. And figure out some other criteria to delete them :/

nyarly:
I'd still rather not be using the option functions pattern for the queues, but I'm okay with this as it stands.

lib/r11n_queue.go
samsalisbury: Yup that's true, or else grow unbounded in memory. Alternatively we could store these in postgres and avoid dealing with cleanup policy for a little longer...

samsalisbury: I suspect we'll have to do that in fact, as I think we'll want to keep these things around beyond restarts. Keeping hold of them once they've been processed could also be useful in stable state detection.

samsalisbury:
OK, fixed the author on some of those commits, and test coverage should now be good.

lib/r11n_anomaly_message.go
```diff
@@ -0,0 +1,88 @@
...
+	default:
+		return "reason unknown"
+	case r11nDroppedQueueNotEmpty:
+		return "queue not empty"
+	case r11nDroppedQueueFull:
+		return "queue full"
+	}
+}
+
+func (msg *r11nAnomalyMessage) Message() string {
```
nyarly: Structured messages need to conform to at least one of the structured logging interfaces: logging.LogMessage, logging.MetricsMessage or logging.ConsoleMessage. LogMessage is most common, and has 3 methods: DefaultLevel, Message and EachField.

It's advised to use the logging.AssertMessageFields helper to test that fields are generated correctly by the message, as well.

samsalisbury: `r11nAnomalyMessage` conforms to the same interfaces as `diffRezMessage`, namely `logging.LogMessage`. In case this was non-obvious I added a comment to explain this up at https://github.com/opentable/sous/pull/541/files/82f98c997654d50188f53d8070e003b0c62d5e9c#diff-09799ec3fd889c06f42beac2189d8619R9

AssertMessageFields seems like a good idea, I'll see about adding that now.

lib/r11n_queue.go
samsalisbury: I see, yep that would be more efficient, I really can't see this being a bottleneck, but we should keep this optimisation in mind if we ever do hit perf issues around this.

PR #542, opened 2018-01-05
Using placeholders for data-sourced values

nyarly:
Things like table and column names are still being interpolated, since those
are all controlled by the program and known at compile/testing time.

Tests exercise ~30 placeholder parameters. I guess we'll see if there's a hard
upper limit. I wasn't able to find an documentation to the effect.

### DIFF
--- a/ext/storage/postgresstatemanager_write.go
+++ b/ext/storage/postgresstatemanager_write.go
@@ -5,10 +5,12 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"regexp"
 	"strings"
 	"text/template"
 	"time"
 
+	"github.com/lib/pq"
 	sous "github.com/opentable/sous/lib"
 	"github.com/opentable/sous/util/logging"
 )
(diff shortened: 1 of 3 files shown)


sous-machine-account:
LGTM

</p>

ext/storage/postgresstatemanager_write.go
```diff
@@ -323,22 +329,38 @@ func (f fields) noncandidates() string {
...
+	placeIdx := 0
 
 	lines := []string{}
+	for range f.rows {
+		valpats := []string{}
+		for _, name := range f.colnames {
+			pat := f.coldefs[name].fmt
+			pat = placeholderQs.ReplaceAllStringFunc(pat, func(q string) string {
+				placeIdx++
+				return fmt.Sprintf("$%d", placeIdx)
```
samsalisbury: Neat, been down a rabbit hole looking into Postgres SQL to see if this is necessary, but I think it is... Nice that we don't have to maintain indexes manually!

PR #544, opened 2018-01-08
#trivial dev make command

JC1738:


sous-machine-account:
LGTM

</p>

Makefile
```diff
@@ -76,9 +76,18 @@ help:
 	@echo "make test-integration: integration tests"
 	@echo "make test-staticcheck: runs static code analysis against project packages."
 	@echo "make wip: puts a marker file into workspace to prevent Travis from passing the build."
+	@echo "make build-debug: builds a linux debug version "
 	@echo
 	@echo "Add VERBOSE=1 for tons of extra output."
 
+build-debug:
+	@if [[ $(SOUS_VERSION) != *"debug" ]]; then echo 'missing debug at the end of semv, please add'; exit -1; fi
```
samsalisbury: I find myself only wanting to run this command like so:

    SOUS_VERSION=$(git describe)-debug make build-debug

Is there a reason we might want to do something different? If not, maybe worth making this the default to save us having to think of version strings up-front...

samsalisbury: Oh, just looking at https://github.com/opentable/ot-sous-server-dev/pull/2 ... so we want to deploy specific tags? Ok, guess I just need more context on both these PRs.

PR #545, opened 2018-01-09
Fix validate check

JC1738:
If you don't specify brokers for kafka, the validation never actually catches error.

https://stackoverflow.com/questions/35866221/why-does-an-empty-array-returned-from-strings-split-have-a-length-of-1-in-golang


sous-machine-account:
LGTM

</p>

JC1738:
This is the line that was expecting to be 0 but since len always was 1 on call to split string on empty string.   

https://github.com/JC1738/sous/blob/f47b6bb09cb14e01fe09df1c20854edc9d671c49/util/logging/config.go#L148

PR #550, opened 2018-01-10
Runspec

JC1738:
Fix for https://opentable.atlassian.net/browse/SOUS-831

Tested with spring-boot-demo

`spring-boot-demo:>sous build -tag 3.0.0
Building with split container
  (Sous)> running docker build --pull .
  (Sous)>   Sending build context to Docker daemon  250.9kB
  (Sous)>   Step 1/1 : FROM docker.otenv.com/sous-otj-autobuild:latest
  (Sous)>   latest: Pulling from sous-otj-autobuild
  (Sous)>   5880d11c5767: Already exists
  (Sous)>   6ea46a0a461e: Already exists
  (Sous)>   b7081e06a7b4: Already exists
  (Sous)>   eb6d1afd0db3: Already exists
  (Sous)>   ff37601757e3: Already exists
  (Sous)>   7f42ca8af2c4: Already exists
  (Sous)>   5c8a126e9da9: Already exists
  (Sous)>   76bfe840e6cb: Already exists
  (Sous)>   177d2b41b6a7: Already exists
  (Sous)>   0bb09f3fd577: Already exists
  (Sous)>   829f323e67dc: Already exists
  (Sous)>   b995817f2749: Already exists
  (Sous)>   ba9c95be020a: Already exists
  (Sous)>   Digest: sha256:9ad7dcf9e9f31a5bbecfdbf93e00830f47648305192e71c3ae11211882d3366a
  (Sous)>   Status: Image is up to date for docker.otenv.com/sous-otj-autobuild:latest
  (Sous)>   # Executing 5 build triggers
  (Sous)>    ---> Using cache
  (Sous)>    ---> Using cache
  (Sous)>    ---> Using cache
  (Sous)>    ---> Using cache
  (Sous)>    ---> Using cache
  (Sous)>    ---> 2c50f0250f88
  (Sous)>   Successfully built 2c50f0250f88
  (Sous)> running docker create 2c50f0250f88
  (Sous)>   c9db5a47c33a7db133e56a8c576829c4a9fad87205145713313ebfc8023b34b2
  (Sous)> running docker cp c9db5a47c33a7db133e56a8c576829c4a9fad87205145713313ebfc8023b34b2:/run_spec.json /var/folders/b4/wly56yrs3nzbbc_vtb__t1l40000gp/T/sous-split-build215553755/runspec.json
Deploy spec invalid:
	No image was created, was .build-service marker file provided?: cannot be repaired.
spring-boot-demo:>`


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

Makefile
```diff
@@ -152,6 +152,9 @@ install-staticcheck:
 install-metalinter:
 	go get github.com/alecthomas/gometalinter
 
+install-liquibase:
+	brew install liquibase
```
samsalisbury: This assumes macOS with homebrew installed (other usages of brew are either explicit: `install-brew`, or ignore failure: `install-dev` calls `brew uninstall opentable/public/sous || true`). I'll add a check for brew installed so we can fail gracefully if this is called when homebrew not installed.

ext/docker/splitcontainer_runspec.go
```diff
@@ -62,12 +62,16 @@ func (ms *MultiImageRunSpec) Validate() []sous.Flaw {
 		}
 		fs = append(fs, ms.SplitImageRunSpec.Validate()...)
 	} else {
-		for idx, spec := range ms.Images {
-			sfs := spec.Validate()
-			for _, f := range sfs {
-				f.AddContext("image %d", idx)
+		if len(ms.Images) < 1 {
+			fs = append(fs, sous.FatalFlaw("No image was created, was .build-service marker file provided?"))
```
samsalisbury: This check applies to all builders, so the error message will be misleading for those not using the OTJ builder. We should link to the builder docs if possible so users can get specific help.

samsalisbury: Also, can return early here and avoid the `else` block. See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

ext/docker/splitcontainer_splitbuilder_test.go
```diff
@@ -78,6 +78,32 @@ func TestSplitBuilder_ValidateRunspec(t *testing.T) {
...
+			SplitImageRunSpec: &SplitImageRunSpec{
+				Files: []sbmInstall{{Source: sbmFile{"a"}, Destination: sbmFile{"a"}}},
+			},
+		},
+	}
+
+	builder.RunSpec.SplitImageRunSpec.Image.Type = "docker"
+	builder.RunSpec.SplitImageRunSpec.Image.From = "test"
+	exec := make([]string, 3)
+	exec = append(exec, "a", "b", "c")
```
samsalisbury: These two lines may not be doing what you expect. `make([]string, 3)` creates a slice with _length_ 3, so the append calls will be adding indexes 3, 4, 5, with 3 empty strings at 0, 1, 2. (See https://play.golang.org/p/B-ebj-cI9B5). It may work anyway, but a better approach might be to do:

    exec := []string{"a", "b", "c"}

Or

    var exec []string
    exec = append(exec, "a", "b", "c")

In Go, you can append to a nil slice, so it's common practice to use the second approach when you don't know how many items will be in the slice, see https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

Note that it is possible to allocate a larger slice without setting its _length_. To do this you provide the _capacity_  arg to make: `make([]string, 0, 3)`. However, there would be little point doing that here as we already know the exact contents of the slice. See also https://golang.org/pkg/builtin/#make (Go has generics! Who knew :P)

ext/docker/splitcontainer_splitbuilder_test.go
```diff
@@ -78,6 +78,32 @@ func TestSplitBuilder_ValidateRunspec(t *testing.T) {
...
+
+	assert.NoError(t, builder.validateRunSpec())
+}
+
+func TestSplitBuilder_ValidateRunspec_noImageError(t *testing.T) {
+	builder := splitBuilder{
+		RunSpec: &MultiImageRunSpec{},
+	}
+
+	assert.Error(t, builder.validateRunSpec(), "should error no Image present")
```
samsalisbury: For validation messages especially, it's worth asserting on the specific error message rather than just the presence of a non-nil error, since this is part of the UX of Sous and this change is primarily to improve on that. (If the message is likely to be wrapped in other messages, then `string.Contains` call might be more appropriate than strict equality, but in this case I would write the assertion like this:

```go
gotErr := builder.validateRunSpec()
wantErr := "desired error message"
if gotErr == nil {
  t.Fatalf("got nil error; want %q", wantErr)
}
if gotErr.Error() != wantErr {
  t.Fatalf("got error %q; want %q", gotErr, wantErr)
}
```

It's more verbose to write, but when this test fails there's a lot more info to go on. Note that for `%q` (and `%s`) the fmt package calls `error.Error() string` to get the string, as long as the error is not also a `fmt.Stringer` in which case it will call `err.String()` oh magic. I usually avoid reliance on this magic by first reading the string from `err.Error()` first and then comparing that. This adds another line to the test but makes it slightly more solid in the face of new methods added to the error later. (Digression: maybe it should be a linter check that types implementing both `Error() string` and `String() string` should always return the same string, and that they always return the same string on multiple calls given the same underlying value...)

Makefile
samsalisbury: If #551 gets merged, you  can use the new [`homebrew` target](https://github.com/opentable/sous/pull/551/files#diff-b67911656ef5d18c4ae36cb6741b7965R122) from that.

ext/docker/splitcontainer_runspec.go
```diff
@@ -63,15 +63,16 @@ func (ms *MultiImageRunSpec) Validate() []sous.Flaw {
...
 		if len(ms.Images) < 1 {
-			fs = append(fs, sous.FatalFlaw("No image was created, was .build-service marker file provided?"))
-		} else {
-			for idx, spec := range ms.Images {
-				sfs := spec.Validate()
-				for _, f := range sfs {
-					f.AddContext("image %d", idx)
-				}
-				fs = append(fs, sfs...)
+			fs = append(fs, sous.FatalFlaw("No image was created, please see https://github.com/opentable/sous-build-containers"))
```
samsalisbury: The issue here is that that's a private repo at present, so the link will be broken for non-opentablers. @nyarly suggested in SOUS-831 that a meaningful link could be extracted from the repo field inside the builder image metadata. I think that would be nicer, but if it's a lot of work then prob not worth it. Simply saying "No images present in runspec.json" and failing is enough here. We can be more specific, and fail much earlier by also putting a change inside the OTJ builder as suggested in SOUS-825 (epic). I'll extract that into its own ticket as well now...

samsalisbury: Added SOUS-838 to track this.

Makefile
nyarly: Since this isn't a dependency of other tasks, while I agree with adding the homebrew target to it, it seems reasonable to merge this as is and come back to it.

ext/docker/splitcontainer_splitbuilder_test.go
nyarly: Just to point out: `testify/assert` has EqualError which will let you specify a string like `wantError` above in a single line.

samsalisbury: Nice, I hadn't seen that before.

PR #551, opened 2018-01-11
Grab bag: makefile and error message improvements

samsalisbury:
Some incidental tidbits... In an attempt to keep other PRs clean and focussed, I've extracted these into their own unfocussed PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,6 +269,7 @@ artifacts/sous_$(SOUS_VERSION)_amd64.deb: artifacts/$(LINUX_RELEASE_DIR)/sous
 	mv sous_$(SOUS_VERSION)_amd64.deb artifacts/
 
 $(DEV_POSTGRES_DATA_DIR):
+	@if [ -d "$@" ]; then exit 0; fi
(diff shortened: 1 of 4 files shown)


samsalisbury:
I'll look into this tomorrow... If anyone can spot the error do let me know :)

Makefile
```diff
@@ -277,7 +282,7 @@ postgres-start: $(DEV_POSTGRES_DATA_DIR)/postgresql.conf
 
 postgres-test-prepare: $(DEV_POSTGRES_DATA_DIR)/postgresql.conf postgres-create-testdb
 
-postgres-create-testdb:
+postgres-create-testdb: postgres-start
```
nyarly: This is the problem you're having in Travis, I think.

In CI, we rely on Travis running postgres as a service, rather than spinning up our own instance of it. Adding this dependency collides with those assumptions (and incidentally triggers a dependency on initdb which I never figured out how to install on Travis.)

sous-machine-account:
LGTM

</p>

PR #552, opened 2018-01-11
New server log message

JC1738:
I wanted to get this PR out just so @nyarly and others can review and comment if this is the right direction.

Additional would like to add otpl fields for the deploy filters, version, listen address.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

cli/actions/server.go
```diff
@@ -121,3 +121,8 @@ func (msg serverMessage) Message() string {
 func (msg serverMessage) composeMsg() string {
 	return fmt.Sprintf("%s, server v%s at %s for %s: DeployFilter Flags %v", msg.msg, msg.version, msg.listenAddress, msg.deployFilterFlags.Cluster, msg.deployFilterFlags)
 }
+
+func (msg serverMessage) EachField(f logging.FieldReportFn) {
```
JC1738: @nyarly I was confused, logConfigurationError seems to also not have EachField, but is of type LogMessage according to my IDE, I thought to be of an interface need to support all methods of that interface?

JC1738: Tested and got generic  log message in kibana
monosnap 2018-01-11 15-52-02

nyarly: logConfigurationError doesn't need any fields beyond the CallerInfo backtrace - as such it can make use of the (fairly narrow) case of using an anonymous embed of CallerInfo to fulfill parts of the LogMessage interface - iow because CallerInfo is included as an anonymous member of logConfigurationError, lCE responds to EachField by forwarding to CallerInfo, and that's how it fulfills that part of the interface.

cli/actions/server.go
```diff
@@ -70,10 +73,54 @@ func ensureGDMExists(repo, localPath string, log func(string, ...interface{})) e
...
+	version           semv.Version
+	listenAddress     string
+}
+
+func logServerMessage(msg string, filterFlags config.DeployFilterFlags, version semv.Version, listenAddress string, log logging.LogSink) {
+	msgLog := serverMessage{
+		msg:               msg,
+		CallerInfo:        logging.GetCallerInfo(logging.NotHere()),
+		deployFilterFlags: filterFlags,
+		version:           version,
```
nyarly: The logger already includes the Sous version in every message it emits, so this is superfluous.

cli/actions/server.go
```diff
@@ -70,10 +73,54 @@ func ensureGDMExists(repo, localPath string, log func(string, ...interface{})) e
...
+func (msg serverMessage) DefaultLevel() logging.Level {
+	return logging.WarningLevel
+}
+
+func (msg serverMessage) Message() string {
+	return msg.composeMsg()
+}
+
+func (msg serverMessage) composeMsg() string {
+	return fmt.Sprintf("%s, server v%s at %s for %s: DeployFilter Flags %v", msg.msg, msg.version, msg.listenAddress, msg.deployFilterFlags.Cluster, msg.deployFilterFlags)
```
nyarly: I've been using a single unformatted string here, in general, but I don't have a strong argument one way or the other for that. Worth discussing?

cli/actions/server.go
```diff
@@ -70,10 +73,54 @@ func ensureGDMExists(repo, localPath string, log func(string, ...interface{})) e
...
+
+type serverMessage struct {
+	logging.CallerInfo
+	msg               string
+	deployFilterFlags config.DeployFilterFlags
+	version           semv.Version
+	listenAddress     string
+}
+
+func logServerMessage(msg string, filterFlags config.DeployFilterFlags, version semv.Version, listenAddress string, log logging.LogSink) {
```
nyarly: I've been using a pattern like "reportServerMessage", partly because the LogSink can also print messages to the console or collect metrics based on messages, partly to avoid reusing the word "log" any more than I had to.

I've got only a weak opinion about this, apart from wanting consistency in the names.

cli/actions/server.go
JC1738: @nyarly Totally forgot about the anonymous nature of struct, explains it, thanks!

cli/actions/server.go
JC1738: ah, will remove version

cli/actions/server.go
JC1738: ok, will change to make consistant

cli/actions/server.go
JC1738: wouldn't need to do this once the loglov message type supported, until then didn't know a good way to get that information to always come across.

cli/actions/server.go
```diff
@@ -30,20 +32,20 @@ type Server struct {
...
 // Do runs the server.
 func (ss *Server) Do() error {
-	if err := ensureGDMExists(ss.GDMRepo, ss.Config.StateLocation, ss.Log.Warnf); err != nil {
+	if err := ensureGDMExists(ss.GDMRepo, ss.Config.StateLocation, ss.DeployFilterFlags, ss.ListenAddr, ss.Log); err != nil {
 		return err
 	}
-	ss.Log.Warnf("Starting scheduled GDM resolution.")
-	ss.Log.Warnf("Filtering the GDM to resolve on this server to: %v", ss.DeployFilterFlags)
+
+	reportServerMessage("Starting scheduled GDM resolution.  Filtering the GDM to resolve on this server", ss.DeployFilterFlags, ss.ListenAddr, ss.Log)
```
samsalisbury: Whenever I see multiple fields of the same value (in this case ss) being passed in to a func, I think it might be neater as a method on that type, the result could be:

    ss.reportServerMessage("Starting scheduled GDM resolution.  Filtering the GDM to resolve on this server")

cli/actions/server.go
```diff
@@ -30,20 +32,20 @@ type Server struct {
...
 	ss.AutoResolver.Kickoff()
 
-	ss.Log.Warnf("Sous Server v%s running at %s for %s", ss.Version, ss.ListenAddr, ss.DeployFilterFlags.Cluster)
+	reportServerMessage("Sous Server Running", ss.DeployFilterFlags, ss.ListenAddr, ss.Log)
 
 	return server.Run(ss.ListenAddr, ss.ServerHandler)
 }
 
-func ensureGDMExists(repo, localPath string, log func(string, ...interface{})) error {
+func ensureGDMExists(repo, localPath string, filterFlags config.DeployFilterFlags, listenAddress string, log logging.LogSink) error {
```
samsalisbury: See calling of ensureGDMExists... This should probably always have been a method on `actions.Server` as well. Totally not blocking over these btw, just noticing potentially nice refactors as I go...

cli/actions/server.go
```diff
@@ -53,7 +55,8 @@ func ensureGDMExists(repo, localPath string, log func(string, ...interface{})) e
 		if len(files) != 0 {
 			// The directory exists and is not empty, do nothing.
 			if repo != "" {
-				log("not pulling repo %q; directory already exist and is not empty: %q", repo, localPath)
+				msg := fmt.Sprintf("not pulling repo %q; directory already exist and is not empty: %q", repo, localPath)
```
samsalisbury: Log messages like this can reasonably use sentence conventions (initial capital, final full-stop/other punctuation) I think because they constitute a single string field in the log; we don't expect it to be composed into another sentence, like error messages.

PR #553, opened 2018-01-11
Duplex state storage, copies state out to Postgres DB

nyarly:
The goal here is to keep the two states in sync until we're ready to switch over.

Note that if there's a problem connecting to the DB,
or saving state to it, there's only a log entry -
the designed intent here is that the secondary store
not influence the operation of Sous.

### DIFF
--- a/ext/storage/duplexstatemanager.go
+++ b/ext/storage/duplexstatemanager.go
@@ -1,0 +1,41 @@
+package storage
+
+// A DuplexStateManager echoes StateManager operation to a primary StateManager,
+// but also ensures that writes occur to the secondary one.
+type DuplexStateManager struct {
+	primary, secondary sous.StateManager
+	log                logging.LogSink
+}
+
+// NewDuplexStateManager creates a DuplexStateManager
+func NewDuplexStateManager(primary, secondary sous.StateManager, log logging.LogSink) *DuplexStateManager {
+	return &DuplexStateManager{
+		primary:   primary,
+		secondary: secondary,
+		log:       log,
+	}
+}
+
+// ReadState implements StateManager on DuplexStateManager
+func (dup *DuplexStateManager) ReadState() (*sous.State, error) {
+	start := time.Now()
+	state, err := dup.primary.ReadState()
+	if err != nil {
+		if err := dup.secondary.WriteState(state, user); err != nil {
+			logging.ReportError(dup.log, errors.Wrapf(err, "writing to secondary StateManager"))
+		}
+	}
+	reportReading(dup.log, start, state, err)
+	return state, err
+}
+
+// WriteState implements StateManager on DuplexStateManager
+func (dup *DuplexStateManager) WriteState(state *sous.State, user sous.User) error {
+	start := time.Now()
+	if err := dup.secondary.WriteState(state, user); err != nil {
+		logging.ReportError(dup.log, errors.Wrapf(err, "writing to secondary StateManager"))
+	}
+	err := dup.primary.WriteState(state, user)
+	reportWriting(dup.log, start, state, err)
+	return err
+}
--- a/ext/storage/duplexstatemanager.go
+++ b/ext/storage/duplexstatemanager.go
@@ -1,5 +1,13 @@
 package storage
 
+import (
+	"time"
+
+	sous "github.com/opentable/sous/lib"
+	"github.com/opentable/sous/util/logging"
+	"github.com/pkg/errors"
+)
+
 // A DuplexStateManager echoes StateManager operation to a primary StateManager,
 // but also ensures that writes occur to the secondary one.
 type DuplexStateManager struct {
@@ -18,9 +26,10 @@ func NewDuplexStateManager(primary, secondary sous.StateManager, log logging.Log
 
 // ReadState implements StateManager on DuplexStateManager
 func (dup *DuplexStateManager) ReadState() (*sous.State, error) {
+	user := sous.User{}
 	start := time.Now()
 	state, err := dup.primary.ReadState()
-	if err != nil {
+	if err == nil {
 		if err := dup.secondary.WriteState(state, user); err != nil {
 			logging.ReportError(dup.log, errors.Wrapf(err, "writing to secondary StateManager"))
 		}
--- a/ext/storage/duplexstatemanager_test.go
+++ b/ext/storage/duplexstatemanager_test.go
@@ -1,0 +1,131 @@
+package storage
+
+import (
+	"database/sql"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/opentable/sous/lib"
+	"github.com/opentable/sous/util/logging"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDuplexWrite(t *testing.T) {
+	s := exampleState()
+
+	clobberDir(t, "testdata/result")
+	PrepareTestGitRepo(t, s, "testdata/remote", "testdata/out")
+
+	db := setupDB(t)
+	defer db.Close()
+
+	log, _ := logging.NewLogSinkSpy()
+	gsm := NewGitStateManager(NewDiskStateManager("testdata/out"))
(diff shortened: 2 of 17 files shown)


sous-machine-account:
LGTM

</p>

samsalisbury:
Hmm, failing because of [`status_poller_test.go:808: Empty subpoller polling took more than 100ms`](https://travis-ci.org/opentable/sous/builds/327939387#L2355) again.. Re-running Travis, I wonder if that timeout needs to be upped (unlikely) or what the underlying issue is... I don't think I've seen this failure locally.

ext/storage/messages.go
```diff
@@ -54,3 +55,76 @@ func (msg *sqlMessage) EachField(fn logging.FieldReportFn) {
...
+		CallerInfo:      logging.GetCallerInfo(logging.NotHere()),
+		MessageInterval: logging.NewInterval(started, time.Now()),
+		state:           state,
+		direction:       dir,
+		err:             err,
+	}
+}
+
+func (msg *storeMessage) DefaultLevel() logging.Level {
+	return logging.DebugLevel
```
samsalisbury: I wonder if we should make this at least a warning when `msg.err != nil` ?

ext/storage/duplexstatemanager.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+// ReadState implements StateManager on DuplexStateManager
+func (dup *DuplexStateManager) ReadState() (*sous.State, error) {
+	user := sous.User{}
+	start := time.Now()
+	state, err := dup.primary.ReadState()
+	if err == nil {
+		if err := dup.secondary.WriteState(state, user); err != nil {
```
samsalisbury: Write-secondary-on-read-primary seems odd here but I may be misunderstanding. Should we instead read secondary and check it agrees with the read of primary (off in a goroutine to avoid slowing it down)? Or maybe this behaviour is preferable if we're just trying to keep secondary updated rather than verify it _behaves_ the same too...

ext/storage/duplexstatemanager.go
```diff
@@ -0,0 +1,50 @@
...
+	return state, err
+}
+
+// WriteState implements StateManager on DuplexStateManager
+func (dup *DuplexStateManager) WriteState(state *sous.State, user sous.User) error {
+	start := time.Now()
+	if err := dup.secondary.WriteState(state, user); err != nil {
+		logging.ReportError(dup.log, errors.Wrapf(err, "writing to secondary StateManager"))
+	}
+	err := dup.primary.WriteState(state, user)
```
samsalisbury: May be preferable to parallelize these two writes and block until they both complete to keep latency to a minimum?

ext/storage/duplexstatemanager_test.go
```diff
@@ -0,0 +1,131 @@
...
+		port = np
+	}
+	connstr := fmt.Sprintf("dbname=sous_test_template host=localhost port=%s sslmode=disable", port)
+	setupDB, err := sql.Open("postgres", connstr)
+	if err != nil {
+		t.Logf("Error setting up test database Error: %v. Did you already `make postgres-test-prepare`?", err)
+		t.FailNow()
+	}
+	// ignoring error because I think "no such DB is a failure"
+	setupDB.Exec("drop database sous_test")
```
samsalisbury: This kind of thing can end up being more confusing in future than it seems now, as there are other possible errors than "no such DB". I would first check if the DB exists, then report the error from this `Exec` call as a test failure if it fails despite the database existing.

ext/storage/duplexstatemanager_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if err != nil {
+		t.Logf("Setting up, error: %v", err)
+		t.FailNow()
+	}
+	return db
+}
+
+func assertStatesEqual(t *testing.T, oldState, newState *sous.State) {
+	t.Helper()
```
samsalisbury: :D

ext/storage/messages.go
```diff
@@ -54,3 +55,76 @@ func (msg *sqlMessage) EachField(fn logging.FieldReportFn) {
...
+	default:
+		return "Unknown state storage direction (shouldn't ever occur?)"
+	case read:
+		return "Reading state"
+	case write:
+		return "Writing state"
+	}
+}
+
+func (msg *storeMessage) EachField(fn logging.FieldReportFn) {
```
samsalisbury: I feel like this should include an explicit field about which stateReader/Writer is reading/writing. This should make it a bit easier to query than just relying on CallerInfo fields which are relatively likely to change with refactorings etc.

ext/storage/postgresstatemanager_read.go
```diff
@@ -28,12 +30,15 @@ func (m PostgresStateManager) ReadState() (*sous.State, error) {
 
 	state, err := loadState(context, m.log, tx)
 	if err != nil {
+		reportReading(m.log, start, state, err)
```
samsalisbury: Maybe consider using `errors.Wrap` or similar on all `reportReading` `err` arguments in this function to give the error message some context? Reading can fail 3 different ways here; the error messages may be clear about this, but that's presumably dependent on the particular db lib we're using...

ext/storage/duplexstatemanager_test.go
```diff
@@ -0,0 +1,131 @@
...
+	expected := exampleState()
+
+	sameYAML(t, actual, expected)
+
+	pstate, err := psm.ReadState()
+	require.NoError(t, err)
+	assertStatesEqual(t, expected, pstate)
+}
+
+func setupDB(t *testing.T) *sql.DB {
```
samsalisbury: Nice, does this obviate any of our Makefile test DB setup at all? At least for unit tests?

ext/storage/duplexstatemanager.go
nyarly: ATM, the design here is just to ensure that the secondary is updated correctly. Keep in mind that the primary in this case can be updated "behind the scenes" from the perspective of the local postgres store.

A confirmation that they're the same seems like it'd be a good idea as well, but I don't know if we want to run that in-line every time.

ext/storage/duplexstatemanager_test.go
nyarly: Not at all, as it relies on the DB being set up by liquibase ahead of time.

What we're doing here is using the "real" Sous database as a template for the `sous_test` DB, which lets us set up the test DB on a per-test run very quickly. (A technique wrested from long years of Rails development.)

ext/storage/messages.go
nyarly: :+1:

ext/storage/messages.go
nyarly: CallerInfo fields will refer to the full name of the function this was logged from, which will include the classname. We could include a "state manager type" field but I hesitate because it will either add extra unnecessary log latency to collect, or introduce the possibility for error without introducing new reliable information.

ext/storage/duplexstatemanager.go
samsalisbury: Cool, makes sense.

ext/storage/duplexstatemanager.go
nyarly: I'd consider it. At the moment, recording state is not where out latency goes, so my feeling is to keep parallelism to its own PR for careful review.

ext/storage/messages.go
samsalisbury: OK, as long as we bear in mind that innocent looking code-level method/type renames could break dashboards/saved searches based on this. I was really suggesting a string field with a set of blessed constant values e.g. `sous-state-reader-type: postgres|git`

ext/storage/duplexstatemanager.go
samsalisbury: Yep, makes sense.

ext/storage/duplexstatemanager_test.go
samsalisbury: OK, got it, tx.

PR #554, opened 2018-01-12
Fix r11n queue race

samsalisbury:
As per title, this fixes a data race picked up by the race detector (between `.init` and `.Start`) and addresses an issue where we would occasionally deadlock due to interaction between `next` and `pop` methods. The `pop` method was not in use and has been removed.


sous-machine-account:
LGTM

</p>

PR #555, opened 2018-01-12
Resource message

JC1738:
replacing logging.Log.Warn and .Vomit with structure


lib/resources.go
```diff
@@ -128,36 +129,106 @@ func (r Resources) Ports() int32 {
...
+	logging.CallerInfo
+	msg        string
+	ports      int32
+	cpus       float64
+	memory     float64
+	isDebugMsg bool
+}
+
+func reportResourceMessage(msg string, r Resources, log logging.LogSink, debug ...bool) {
+	debugStmt := false
```
JC1738: Kind of hackish, but one of the suggested ways of using optional parameters.  Could specify a reportResourceMessageDebug method instead, but liked the implicity, but can change to other suggestions.

lib/resources.go
```diff
@@ -128,36 +129,106 @@ func (r Resources) Ports() int32 {
...
+		isDebugMsg: debugStmt,
+	}
+	logging.Deliver(msgLog, log)
+}
+
+func (msg resourceMessage) WriteToConsole(console io.Writer) {
+	fmt.Fprintf(console, "%s\n", msg.composeMsg())
+}
+
+func (msg resourceMessage) DefaultLevel() logging.Level {
```
JC1738: I assume that each time Deliver is called this is evaluated, if that is not the case then need to refactor.

sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

lib/resources.go
samsalisbury: I can see the temptation to do this, feels like it makes it easy to quickly switch a log up to debug or not. However it makes it more difficult to search for debug level logs, and isn't as descriptive for the reader as a separate `reportDebugResourceMessage` or maybe `debugResourceMessage` for brevity...

lib/resources.go
samsalisbury: Yup, other things depend on this being the case as well.

lib/resources.go
```diff
@@ -128,36 +129,106 @@ func (r Resources) Ports() int32 {
...
 
 	if math.Abs(r.Cpus()-o.Cpus()) > 0.001 {
-		logging.Log.Vomit.Println("Cpus differ")
+		reportResourceMessage("Cpus differ", r, logging.Log, true)
 		return false
 	}
 
 	if math.Abs(r.Memory()-o.Memory()) > 0.001 {
-		logging.Log.Vomit.Println("Memory differ")
+		reportResourceMessage("Memory differ", r, logging.Log, true)
```
samsalisbury: I'm not sure if there is much value in keeping the logs in this method (`Resources.Equal`).

lib/resources.go
JC1738: @samsalisbury Yes, while easier to code, searching for these messages would be harder, okay, will move them apart.

JC1738:
Yes, looking at the code, these are validates as flaws and bubbled up, on DeployConfig Validate()  https://github.com/opentable/sous/blob/dea6d8d1b3317287d4f4455156533ad8f83b9964/lib/deploy_config.go#L60 so that flaws are generated after checking the resources of the DeployConfig as well as Startup.  

Think this round of refactoring is purely on the log message, while the other is an actual story / backlog of general how we want to continue on the flaws / errors / messaging aspect throughout the processes?

samsalisbury:
I concur, that tangent is out of scope for this PR. Should just add a changelog entry that we've structured these messages up, and good to merge, tx.

samsalisbury:
Just merged this accidentally, trying out a GitHub app “CodeHub” with confusing UI... Looks like Travis never finished, will check it passes on master & fix any issues.

samsalisbury:
...for completeness this merge passed on master anyway... https://travis-ci.org/opentable/sous/builds/329924646

PR #556, opened 2018-01-13
[WIP] Sous 824 structured logging on singularity

lachlanh:
first cut at structured logging in the singularity package, please review and let me know if this is on the right track. At the moment it is not using a specific otl, for the deployable pair, that will be the next revision

### DIFF
--- a/ext/singularity/deployer.go
+++ b/ext/singularity/deployer.go
@@ -211,53 +211,53 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
 func (r *deployer) RectifySingleModification(pair *sous.DeployablePair) (err error) {
 	different, diffs := pair.Post.Deployment.Diff(pair.Prior.Deployment)
 	if !different {
-		logDeployerMessage(fmt.Sprintf("Attempting to rectify empty diff for %q", pair.ID()),
+		reportDeployerMessage(fmt.Sprintf("Attempting to rectify empty diff for %q", pair.ID()),
 			pair, logging.WarningLevel, logging.Log)
 	}
 
-	logDeployerMessage(fmt.Sprintf("Rectifying modified %q; Diffs: %s", pair.ID(), strings.Join(diffs, "\n")),
+	reportDeployerMessage(fmt.Sprintf("Rectifying modified %q; Diffs: %s", pair.ID(), strings.Join(diffs, "\n")),
 		pair, logging.InformationLevel, logging.Log)
-	logDeployerMessage(fmt.Sprintf("Full prior and post deployments: %q: \n  %# v \n    =>  \n  %# v", pair.ID(), pair.Prior.Deployment, pair.Post.Deployment),
+	reportDeployerMessage(fmt.Sprintf("Full prior and post deployments: %q: \n  %# v \n    =>  \n  %# v", pair.ID(), pair.Prior.Deployment, pair.Post.Deployment),
 		pair, logging.DebugLevel, logging.Log)
 	defer rectifyRecover(pair, "RectifySingleModification", &err)
 
 	data, ok := pair.ExecutorData.(*singularityTaskData)
 	if !ok {
 		err := errors.Errorf("Modification record %#v doesn't contain Singularity compatible data: was %T\n\t%#v", pair.ID(), data, pair)
-		logDeployerMessage(err.Error(), pair, logging.WarningLevel, logging.Log)
+		reportDeployerMessage(err.Error(), pair, logging.WarningLevel, logging.Log)
 		return err
 	}
 	reqID := data.requestID
 
 	changesApplied := false
-	logDeployerMessage(fmt.Sprintf("Operating on request %q", reqID),
+	reportDeployerMessage(fmt.Sprintf("Operating on request %q", reqID),
 		pair, logging.ExtraDebug1Level, logging.Log)
 	if changesReq(pair) {
-		logDeployerMessage("Updating Request...", pair, logging.DebugLevel, logging.Log)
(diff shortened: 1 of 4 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

ext/singularity/deployermessage.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type deployerMessage struct {
+	logging.CallerInfo
+	logging.Level
+	msg string
+	pair *sous.DeployablePair
+}
+
+func logDeployerMessage(message string, pair *sous.DeployablePair, level logging.Level, logger logging.LogSink) {
```
samsalisbury: Very minor, other funcs of this nature are prefixed `report` rather than `log` so may be better to follow that convention here. AFAIK @nyarly chose 'report' because strictly these could do things other than write logs.

ext/singularity/deployermessage.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (msg deployerMessage) Message() string {
+	return msg.msg
+}
+
+func (msg deployerMessage) EachField(f logging.FieldReportFn) {
+	f("@loglov3-otl", "sous-generic-v1")
+	msg.CallerInfo.EachField(f)
+}
```
samsalisbury: This method is missing use of `msg.msg` and `msg.pair` atm. We need to use a more specific OTL than `sous-generic-v1` and add those fields here, as the old `Warnf` and friends were already ending up using this OTL. To make this more structured, we'll need to add more context to help with filtering, especially the fields from `sous-resolve-filter-v1` OTL.

ext/singularity/deployermessage_test.go
```diff
@@ -0,0 +1,24 @@
...
+	pair := baseDeployablePair()
+
+	logDeployerMessage("test", pair, logging.InformationLevel, logger)
+
+	logCalls := control.CallsTo("LogMessage")
+	require.Len(t, logCalls, 1)
+	assert.Equal(t, logCalls[0].PassedArgs().Get(0), logging.InformationLevel)
+
+	//weak check on WriteToConsole
+	consoleCalls := control.CallsTo("Console")
```
samsalisbury: Curious why you wouldn't just use `"WriteToConsole"` ?

ext/singularity/deployermessage_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func TestDeployerMessage(t *testing.T) {
+	logger, control := logging.NewLogSinkSpy()
+	pair := baseDeployablePair()
+
+	logDeployerMessage("test", pair, logging.InformationLevel, logger)
+
+	logCalls := control.CallsTo("LogMessage")
+	require.Len(t, logCalls, 1)
+	assert.Equal(t, logCalls[0].PassedArgs().Get(0), logging.InformationLevel)
```
samsalisbury: Nice, this will definitely break appropriately... I tend to write tests at a slightly higher level, often just against exported funcs/methods, slightly black-boxing them and not necessarily covering to this level of detail, also makes them less likely to break for "trivial" changes like changing level in this case. Not against this level of detail though, maybe we should use more of it...

ext/singularity/deployermessage.go
```diff
@@ -0,0 +1,42 @@
...
+		CallerInfo: logging.GetCallerInfo(logging.NotHere()),
+		Level: 		level,
+		msg:		message,
+		pair:		pair,
+	}
+	logging.Deliver(msg, logger)
+}
+
+func (msg deployerMessage) DefaultLevel() logging.Level {
+	return logging.InformationLevel
```
samsalisbury: Should return `msg.Level` here...

ext/singularity/deployer.go
```diff
@@ -211,47 +211,54 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
...
 	different, diffs := pair.Post.Deployment.Diff(pair.Prior.Deployment)
 	if !different {
-		Log.Warn.Printf("Attempting to rectify empty diff for %q", pair.ID())
+		logDeployerMessage(fmt.Sprintf("Attempting to rectify empty diff for %q", pair.ID()),
+			pair, logging.WarningLevel, logging.Log)
 	}
 
-	Log.Notice.Printf("Rectifying modified %q; Diffs: %s", pair.ID(), strings.Join(diffs, "\n"))
-	Log.Debug.Printf("Full prior and post deployments: %q: \n  %# v \n    =>  \n  %# v", pair.ID(), pair.Prior.Deployment, pair.Post.Deployment)
+	logDeployerMessage(fmt.Sprintf("Rectifying modified %q; Diffs: %s", pair.ID(), strings.Join(diffs, "\n")),
```
samsalisbury: Ideally with structured messages, we should pass in the raw data and have `Message`, `EachField` and `WriteToConsole` etc decide on specific formatting. This should shorten the logging lines slightly and allow customisation for the logging targets in question. See also note on `logDeployerMessage` below.

E.g. these are all 'rectifier' logs, so maybe we could just pass 'modified' and let the `Message` func decide on formatting the rest of the sentence.

ext/singularity/deployermessage_test.go
lachlanh: not really sure how to instrument the test to see the change on the deployermessage. With the console check I think I am testing that WriteToConsole gets called and delegated down to  the console via the rest of the logging infrastructure.

lachlanh:
closing due to switching off my fork

PR #557, opened 2018-01-13
Recording global sync plan

nyarly:


sous-machine-account:
LGTM

</p>

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+that that data is considered
+merely advisory.
+
+Sous clients communicate with a single server -
+although which server is a per-request determination.
+In the event of network isolation
+(or simply by preference)
+a user might configure their workstation to talk to their "closest" Sous server.
+The server handling a client request
+coordinates the response across the Sous infrastructure.
```
samsalisbury: I like that this also simplifies the client a little.

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Protocol Specifics
+
+During a `sous deploy`,
+the Sous client updates the Global Deploy Manifest.
+(Once the update is complete, the client monitors the resulting deploy.
```
samsalisbury: Monitors directly or via further calls to sous server?

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+When a coordinating node receives a
+`GET /gdm`
+it performs conditional
+`GETs /state/deployments`
+to the respective authoritative servers.
+It bases the `If-None-Match` headers on its own local state.
+If the GETs come back with 200 instead of 304,
+the coordinating node updates its local database.
+In the event of a network or HTTP error,
+the coordinating server replies with its local most-recent information.
```
samsalisbury: This model implicitly trusts any response from `/state/deployments` on authoritative servers. I wonder if there is value in detecting and rejecting (or maybe just flagging up via logging/alerts) certain suspicious deltas (e.g. half the deployments have been removed/more, or generally rate of change looks suspiciously high). We don't need this in the initial implementation necessarily, but may be something to bear in mind. I can imagine failure modes in sibling servers where they will incompletely report state for example...

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+`PUT /gdm`
+it determines the implied changes,
+and partitions the deployments based on cluster.
+Those changes are then submitted
+to the managing Sous servers via
+`PATCH /state/deployments`.
+Again, this is a conditional request:
+the `If-Match` is based on the local data
+of the coordinating server.
+Once those responses have come back with 204,
```
samsalisbury: Maybe add "indicating that the state change has been applied locally" or similar.

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+the "coordinating" server records to its local DB,
+and returns 200 to the client.
+
+The 204 reponse is used in the "happy path" case:
+where the authority's state matches the `If-Match` condition,
+and the contents of the request are successfully applied.
```
samsalisbury: OK, above point may be moot.

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+doesn't match the condition of the `PATCH`,
+but that applying the PATCH would have no effect -
+in other words, the data contained in the request are already
+true about the DeployIDs listed in the request.
+In this case, the authority should return
+a 200 with its new state, as if a
+`GET /state/deployments` has been issued.
+The coordinating node,
+in this case,
+updates its local database.
```
samsalisbury: It would be nice if we could let the operator know when this happens as well, e.g. "{DeployID} already at version {Version} in {ClusterName}". This has been requested by users already and this and other ongoing work should make this message easier to add.

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+
+In the event of the remote responding `412 Precondition Failed` to the `PATCH`
+(because the `If-Match` condition fails,
+but the request would have implied a change to the authoritative state.)
+the coordinating node makes a new
+`GET /state/deployments` from the remote,
+updates its local database and determine whether to try again.
+Specifically, it computes a difference of the new authoritative state
+and the state it knew for that cluster when the `PUT /gdm` began.
+If there's an intersection between that difference
```
samsalisbury: Do you have a rough implementation for this in mind? Are we just talking about DeploymentID-scoped intersections, or something more fine grained? May be worth expanding slightly to clarify this.

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+it is safe to retry the `PATCH`,
+and the coordinating node proceeds by doing so.
+
+If any of the authoritative servers return an error,
+in response to `PATCH /state/deployments`
+the coordinating server returns a 503 error
+the body of which indicates which authoritative server(s)
+were unable to service the request,
+which the client relates to the operator,
+so that they can potentially adjust and retry their request.
```
samsalisbury: We should probably re-try the request in some circumstances, especially in the case that the authoritative server did not respond within a timeout, maybe for certain HTTP errors as well?

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+out of band on to
+`PUT /advisory/gdm`
+to all its sibling servers.
+Each server receiving this `PUT` request
+uses the request body to update its database
+about Deployments for which it is *not* authoritative.
+The coordinating server, again, sends conditional requests,
+but disregards `412` responses.
+The non-authoritative servers that *produce* `412`s,
+however, issue their own `GET /state/deployments`
```
samsalisbury: ..."to the authoritative server"

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+
+If any of the authoritative servers return an error,
+in response to `PATCH /state/deployments`
+the coordinating server returns a 503 error
+the body of which indicates which authoritative server(s)
+were unable to service the request,
+which the client relates to the operator,
+so that they can potentially adjust and retry their request.
+
+Finally, the coordinating server
```
samsalisbury: It may make little difference, but why not have the authoritative server make these advisory PUT requests upon successful commit, echoing current state, since it is itself authoritative that may reduce the likelihood of receiving 412s back from the sibling server, and it's possible that the authoritative server has more up to date info at GDM scope. However I've just realised these endpoints sound like "entire GDM" endpoints. Should they not be `/advisory/deployments/{DeploymentID}` (and likewise `/state/deployments/{DeploymentID}`) ?

doc/global_storage.md
```diff
@@ -0,0 +1,207 @@
...
+
+As to actual implementation,
+this synchronization is managed with a decorator StateManager.
+The StateManager interface has two methods:
+ReadState and WriteState.
+The GlobalSyncStateManager wraps another StateMananger
+(i.e. PostgresStateManager)
+and manages the HTTP synchronization requests
+(the ones to /state/...  and /advisory/...)
+around Read/WriteState calls to the wrapped manager.
```
samsalisbury: 👍

doc/global_storage.md
nyarly: Via calls to the sous server (the new shepherd based endpoint.) But out of scope for this doc

doc/global_storage.md
nyarly: I appreciate the caution regarding sudden changes, but the authoritative servers are authoritative. I'm not sure how non-authoritative nodes can adjudicate their responses without crashing the whole infra

doc/global_storage.md
nyarly: This sequence of events indicates that the GDM already records that version - which is not the same as that version already being deployed.

As far as reporting a null GDM update goes, the only meaningful way to do that, I think, is for the client to report it at the outset when it gets the GDM in the first place. If, between getting the GDM and issuing the update, that update has become null (because e.g. another CI process already issued the update) we can still see if there's a deploy that needs to happen.

nyarly:
Very much appreciate the close attention, here, @samsalisbury. Important to get this right.

doc/global_storage.md
nyarly: I do have a rough implementation in mind.

Denote the cluster scoped subset of a GDM CLUSTER($cluster, $gdm).
Denote the members of a diff that are added, deleted or modified as NotStable($diff).
Call the state of the GDM that we (the coordinating node) knew when the PUT was recieved GDM-0.
Call the state of the GDM submitted in the PUT GDM-P.
Compute diff-P0 = DIFF(CLUSTER(x, GDM-0), CLUSTER(x, GDM-P)).
diff-P0 is the contents of the PATCH we send to node x.
Call the state as reported by the authoritative node GDM-A.
Compute diff-A0 = DIFF(CLUSTER(x, GDM-0), CLUSTER(x, GDM-A)).
Collect the set of DeploymentIDs in NotStable(diff-A0) as changed-A0.
Collect the set of DeploymentIDs in NotStable(diff-P0) as changed-P0.

If the intersection of changed-A0 and changed-P0 is not the null set, we issue an error (a 412) because there's a collision between this update and <whatever's been going on over there>.
If the intersection is null, we can safely reissue our PATCH, because there's no collision.

Funnily enough, that seemed way simpler in my head before I started writing it out.

doc/global_storage.md
nyarly: Retries should be safe, and easy enough to implement. There's no guarantee of completion though, so I'm tempted to leave retry up to the operator? In the case of CI, we can stick with our "try 5 times" thing. But I think, as a practical matter, the need for retry should decrease.

doc/global_storage.md
nyarly: No, that absolutely the right approach. I'd been futzing with how to structure those advisory PUTs - the authoritative server sending updates is exactly right. It simplifies the whole advisory update part of the protocol. 

The resource should be like `/advisory/deployments?cluster=...` though.

nyarly: Scoping these updates by deployment is definitely of value, but a whole extra effort. Once we have this is place, it'll be much easier to add deployment scoped updates.

doc/global_storage.md
samsalisbury: Sure, it would be hard to make assumptions strong enough to reject updates on without statistically significant data, and even then outright rejecting updates would come with its own problems. I think the thing to do here is to log enough info to be able to alert on unusual updates in future. I'll add a ticket for this log message now.

PR #565, opened 2018-01-23
Corrects itermittant failure in TestResolveRecorder #trivial

nyarly:
Slight simplification to ResolveRecorder interface (yagni'd performGauranteedPhase)
Transformed table tests to t.Run with helpers.

Sam and I agreed that there's no outside-facing change to this, and it shouldn't have a CL entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ with respect to its command line interface and HTTP interface.
 
 ### Changed
 * All: error parsing repo from SourceLocation now more informative.
-* Server: small change to private interface of ResolveRecorder
 
 ### Fixed
 * All: Data race in rectification queue.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ with respect to its command line interface and HTTP interface.
 
 ### Changed
 * All: error parsing repo from SourceLocation now more informative.
+* Server: small change to private interface of ResolveRecorder
 
 ### Fixed
 * All: Data race in rectification queue.
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -4,7 +4,6 @@ package integration
 
 import (
 	"fmt"
-	"os"
 	"sort"
 	"testing"
 	"time"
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ with respect to its command line interface and HTTP interface.
 
 ### Changed
 * All: error parsing repo from SourceLocation now more informative.
+* Server: small change to private interface of ResolveRecorder
```
samsalisbury: Would argue that this isn't changelog material since we have the commit log for internal changes. Seems that to make Changelog truly useful it should concentrate on significant higher-level changes and changes to the external API and/or behaviour of Sous itself. There are a multitude of small changes like this, most of which rightly go unrecorded in the Changelog, or else it would explode in size.

Since this PR makes a test more stable, and I think it should have no effect on behaviour of Sous, we should either remove the entry altogether, or at most just document that an integration test was made more stable. (Or else document the eternal behaviour that was changed, in case I missed it.)

integration/integration_test.go
```diff
@@ -166,6 +177,7 @@ func (suite *integrationSuite) deployDefaultContainers() {
 
 	// This deployment fails immediately, and never results in a successful deployment at that singularity request.
 	registerAndDeploy(ip, "test-cluster", "supposed-to-fail", "github.com/opentable/homer-says-doh", "fails-labels", "1-fails", []int32{}, nilStartup)
+	WaitForSingularity()
```
samsalisbury: Is this the ultimate fix for the reported bug?

lib/resolvestatus.go
```diff
@@ -178,10 +178,6 @@ func (rr *ResolveRecorder) performPhase(name string, f func() error) {
 	}
 }
 
-func (rr *ResolveRecorder) performGuaranteedPhase(name string, f func()) {
```
samsalisbury: Just curious why this was removed? The idea was to reduce 'return nil' noise at call sites.

lib/resolvestatus_test.go
```diff
@@ -2,153 +2,143 @@ package sous
...
+	t.Helper()
+
+	actual := len(r.late.Log)
+	if actual != expected {
+		t.Errorf("expected log of resolutions to have %d entries; has %d", expected, actual)
+	}
+}
+
+func TestResolveRecorder(t *testing.T) {
+	t.Run("No phases", func(t *testing.T) {
```
samsalisbury: Nice upgrade to sub tests, especially getting the name of the test in here. I like that this continues to work with standard 'go test' flags as well rather than relying on flags of testify.

nyarly:
@samsalisbury I believe I've addressed your concerns and merged @lachlanh's PR collisions (turned out not to really be a thing)

PR #569, opened 2018-01-25
Simple implementation of /health endpoint

nyarly:
Just returns 200 and the version.

Do we want to open a new ticket for DB and Singularity fields?
Or wait until they're actually something we should wait on in /health?

### DIFF
--- a/cli/actions/server_test.go
+++ b/cli/actions/server_test.go
@@ -108,7 +108,7 @@ func TestEnsureGDMExists_notARepoFile(t *testing.T) {
 	}
 
 	if err := os.RemoveAll(testDataDir); err != nil {
-		t.Fatal(err)
+		t.Log(err)
 	}
 
 }
@@ -137,7 +137,7 @@ func TestEnsureGDMExists_notARepoDir(t *testing.T) {
 	}
 
 	if err := os.RemoveAll(testDataDir); err != nil {
-		t.Fatal(err)
+		t.Log(err)
 	}
 
 }
--- a/server/handle_health_test.go
+++ b/server/handle_health_test.go
@@ -1,0 +1,21 @@
+package server
+
+import (
+	"testing"
+
+	sous "github.com/opentable/sous/lib"
+)
+
+func TestHandleServerList_Get(t *testing.T) {
+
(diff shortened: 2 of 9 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

cli/actions/server_test.go
```diff
@@ -108,7 +108,7 @@ func TestEnsureGDMExists_notARepoFile(t *testing.T) {
 	}
 
 	if err := os.RemoveAll(testDataDir); err != nil {
-		t.Fatal(err)
+		t.Log(err)
```
JC1738: Why change to Log vs Fatal, don't we want the tests to clean up after each of themselves?

cli/actions/server_test.go
```diff
@@ -137,7 +137,7 @@ func TestEnsureGDMExists_notARepoDir(t *testing.T) {
...
```
JC1738: @nyarly Same comment as above

server/handle_health_test.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+	data, stat := h.Exchange()
+
+	if stat != 200 {
+		t.Errorf("Expecting 200 status; got %d", stat)
+	}
+
+	rez, is := data.(Health)
+
+	if !is {
```
JC1738: If you get a fatal here, does that just stop processing the http request, but keeps sous server up and running (ie 500 server error)? @nyarly

cli/actions/server_test.go
nyarly: There is an issue where the testDataDir is in use somehow when this RemoveAll gets called, and it intermittently fails tests. It's not an issue, ultimately, since the directory gets recreated anyway. I haven't been able to track down the actual blocker here, either, and I'm somewhat at a loss for doing so.

Plus: these are tests for an implementation we're trying to leave, so I'm more willing to ignore the warts on these tests in the hope of excising them later.

PR #574, opened 2018-01-27
Sous 824 structured logging on singularity

lachlanh:
Moved pr over from my fork, includes judson's and my refactoring of the deployable_message.


ext/singularity/deployer.go
```diff
@@ -209,49 +209,50 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
...
 
 	changesApplied := false
-	Log.Vomit.Printf("Operating on request %q", reqID)
+	reportDeployerMessage("Operating on request", pair, data, logging.ExtraDebug1Level, logging.Log)
 	if changesReq(pair) {
-		Log.Debug.Printf("Updating Request...")
+		reportDeployerMessage("Updating request", pair, data, logging.DebugLevel, logging.Log)
 		if err := r.Client.PostRequest(*pair.Post, reqID); err != nil {
-			Log.Warn.Println(err)
+			reportDeployerMessage(err.Error(), pair, data, logging.WarningLevel, logging.Log)
```
samsalisbury: I'm uneasy about using `err.Error()` as the message title, as error strings often have little context. Maybe prefix with `"Error: "` at least? Or better add an 'error' field containing this message, and change the string here to "Error posting request to Singularity"... That would provide slightly nicer filtering as all errors returned here could be filtered on together, or could filter on the error field to differentiate errors from Singularity...

ext/singularity/deployer.go
```diff
@@ -209,49 +209,50 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
...
-		Log.Debug.Printf("Request %q does not require changes", reqID)
+		reportDeployerMessage("No change to request", pair, data, logging.DebugLevel, logging.Log)
 	}
 
 	if changesDep(pair) {
-		Log.Debug.Printf("Deploying...")
+		reportDeployerMessage("Deploying", pair, data, logging.DebugLevel, logging.Log)
 		if err := r.Client.Deploy(*pair.Post, reqID); err != nil {
-			Log.Warn.Println(err)
+			reportDeployerMessage(err.Error(), pair, data, logging.WarningLevel, logging.Log)
```
samsalisbury: Similar comment to the above regarding use of `err.Error()` for the message title.

ext/singularity/deployer.go
```diff
@@ -209,49 +209,50 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
...
 			return err
 		}
 		changesApplied = true
 	} else {
-		Log.Debug.Printf("Deploy at %q does not require changes", reqID)
+		reportDeployerMessage("No change to deploy for request", pair, data, logging.DebugLevel, logging.Log)
```
samsalisbury: Might read easier: "No changes to Singularity request/deployment required" ?

ext/singularity/deployer.go
```diff
@@ -209,49 +209,50 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
...
 	}
 
 	if !changesApplied {
-		Log.Warn.Printf("No changes applied to Singularity for %q", pair.ID())
+		reportDeployerMessage("No change to singularity", pair, data, logging.DebugLevel, logging.Log)
```
samsalisbury: Not sure we need this message at all, although it might be helpful to log something if we end up in the error handling branch starting on line 245?

ext/singularity/deployermessage_test.go
```diff
@@ -0,0 +1,24 @@
...
+	pair := baseDeployablePair()
+
+	reportDeployerMessage("test", pair, nil, logging.InformationLevel, logger)
+
+	logCalls := control.CallsTo("LogMessage")
+	require.Len(t, logCalls, 1)
+	assert.Equal(t, logCalls[0].PassedArgs().Get(0), logging.InformationLevel)
+
+	//weak check on WriteToConsole
+	consoleCalls := control.CallsTo("Console")
```
samsalisbury: How come we only check `"Console"` not `"WriteToConsole"` here?

lib/loggingprocessor.go
```diff
@@ -15,7 +15,17 @@ func (log loggingProcessor) HandlePairs(dp *DeployablePair) (*DeployablePair, *D
 
 func (log loggingProcessor) doLog(dp *DeployablePair) {
 	msg := &deployableMessage{
-		pair:       dp,
+		submessage: &DeployablePairSubmessage{
+			pair: dp,
+			priorSub: &DeployableSubmessage{
+				deployable: dp.Prior,
+				prefix:     "",
```
samsalisbury: Can drop setting prefix to zero value `""` I think..

lib/loggingprocessor.go
```diff
@@ -15,7 +15,17 @@ func (log loggingProcessor) HandlePairs(dp *DeployablePair) (*DeployablePair, *D
...
+			},
+			postSub: &DeployableSubmessage{
+				deployable: dp.Post,
+				prefix:     "",
```
samsalisbury: Same here.

ext/singularity/deployer.go
```diff
@@ -209,49 +209,50 @@ func (r *deployer) RectifySingleDelete(d *sous.DeployablePair) (err error) {
 }
 
 func (r *deployer) RectifySingleModification(pair *sous.DeployablePair) (err error) {
-	different, diffs := pair.Post.Deployment.Diff(pair.Prior.Deployment)
+	different, _ := pair.Post.Deployment.Diff(pair.Prior.Deployment) //TODO LH we will want to log the diffs (second param here)
```
samsalisbury: Just curious why diffs got left out? Are we waiting on an OTL update being merged, or is this coming in another PR?

ext/singularity/deployermessage_test.go
nyarly: The control returned by NewLogSinkSpy doesn't have calls to WriteToConsole available...

But! c.f. https://github.com/opentable/sous/blob/master/util/logging/testsupport.go#L56

iow, you could
```
control.Console.CallsTo("WriteToConsole")
```
instead

lib/loggingprocessor.go
nyarly: Actually, I think we *need* a prefix set here - `"sous-prior"` I think, right?

lib/loggingprocessor.go
nyarly: and `"sous-post"`

ext/singularity/deployer.go
nyarly: I think we might have another candidate submessage: c.f. https://github.com/opentable/sous/blob/master/util/logging/errors.go

I'd be okay with merging this as is, so long as there's a story entered about refactoring to an error submessage, since I'm sure there are other places we should do that as well.

PR #578, opened 2018-02-05
Fix (?) for ResolveRecorder race #trivial

nyarly:
Upshot here is that we were requiring the clients of the ResolveRecorder
to close the RR's Log channel for it.
The tests blocked on a channel after closing the log,
which meant there was a race between that `close(Log)`
and the tests of whether the RR was `Done()`.

Ultimately, it's better practice _anyway_ to have a channel be
closed by whoever created it - this PR simply moves the handling
of that that channel up to the ResolveRecorder, and removes the
places where it was closed in production/test code.

I **think** this will result in more stable tests.

### DIFF
--- a/lib/cluster_manager.go
+++ b/lib/cluster_manager.go
@@ -1,0 +1,31 @@
+package sous
+
+import "github.com/nyarly/spies"
+
+type (
+	// ClusterManager reads and writes deployments as scoped by cluster
+	ClusterManager interface {
+		ReadCluster(clusterName string) (Deployments, error)
+		WriteCluster(clusterName string, deps Deployments) error
+	}
+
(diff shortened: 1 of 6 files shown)


sous-machine-account:
LGTM

</p>

lachlanh:
Not sure I can make an informed comment on this other than it seems to be adding cluster_manager and handle_state_deployments as well as the channel closing fix, should these have a seperate pr ?

JC1738:
I agree with Lach, I think @nyarly that your PR leaked in some of the other work you were doing, probably makes sense to open a seperate PR for the resolve recorder?

nyarly:
Yep.

PR #580, opened 2018-02-06
Sous 824 structured loggin on singularity

lachlanh:
To fix log errors in prod, this will need  a more thorough review, in light of last problem.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

ext/singularity/deployermessage.go
```diff
@@ -2,53 +2,92 @@ package singularity
...
-	taskData *singularityTaskData
-	error    error
+	msg        string
+	submessage *sous.DeployablePairSubmessage
+	diffs      sous.Differences
+	taskData   *singularityTaskData
+	error      error
+}
+
+type diffResolutionMessage struct {
```
samsalisbury: Ideally put this type (and its methods) into their own file to aid discoverability & avoid having to move it in future when it gets expanded with more fields/methods. (I too often don't follow this advice and then regret it.)

ext/singularity/deployermessage_test.go
```diff
@@ -11,13 +13,138 @@ import (
...
+	}
+
+	logging.AssertMessageFields(t, logMessage, logging.StandardDeployerFields("sous-prior", "sous-post"), expectedFields)
+
+	//weak check on WriteToConsole
+	consoleCalls := control.CallsTo("Console")
+	require.Len(t, consoleCalls, 1)
+}
+
+func TestDeployerMessageNilCheck(t *testing.T) {
```
samsalisbury: This test and the two below look like good candidates to convert to a table-driven test since they only differ in the call to `reportDeployerMessage` and the `expectedFields` map. This will make it easier to add more test cases when fields are added...

util/logging/testsupport.go
```diff
@@ -165,6 +165,56 @@ var StandardVariableFields = []string{
...
+
+// StandardDeployerFields is used to create the standard logging fields
+// for the structured logging on the deployerMessage
+// takes the prior and post prefixes to generate the fields appropriately
+func StandardDeployerFields(prefix1 string, prefix2 string) []string {
+	deployerFields := unprefixedDeployerFields
+	deployerFields = append(deployerFields, StandardVariableFields...)
+	for _, str := range prefixedDeployerFields {
+		deployerFields = append(deployerFields, fmt.Sprintf(str, prefix1))
+		deployerFields = append(deployerFields, fmt.Sprintf(str, prefix2))
```
samsalisbury: I would be tempted to use string concatenation here, `fmt.Sprintf` seems like overkill when we know this is always "prefix a string" and is slightly slower than string concat. May not be an optimisation really worth making though unless it ever becomes a bottleneck which I doubt will happen...

```shell
face:sous ssalisbury$ go test -v -bench Sprintf .
goos: darwin
goarch: amd64
pkg: github.com/opentable/sous
BenchmarkSprintf/fmt.Sprintf-8           5000000               312 ns/op
BenchmarkSprintf/""_+_""-8              10000000               188 ns/op
```

```go
package main

import (
	"fmt"
	"testing"
)

// Results is global and exported to ensure compiler doesn't optimise out code
// under test.
var Results map[int]string

func BenchmarkSprintf(b *testing.B) {
	b.Run("fmt.Sprintf", func(b *testing.B) {
		Results = make(map[int]string, b.N)
		for i := 0; i < b.N; i++ {
			Results[i] = fmt.Sprintf("%s-blah", "hello")
		}
	})
	b.Run(`"" + ""`, func(b *testing.B) {
		Results = make(map[int]string, b.N)
		for i := 0; i < b.N; i++ {
			Results[i] = "hello" + "-blah"
		}
	})
}
```

lachlanh:
@nyarly can you cast another eye over this and if it looks good we will get it into preprod ?

ext/singularity/deployermessage_test.go
```diff
@@ -1,24 +1,216 @@
...
+	}
+
+	logging.AssertMessageFields(t, logMessage, logging.StandardVariableFields, expectedFields)
+
+	//weak check on WriteToConsole
+	consoleCalls := control.CallsTo("Console")
+	require.Len(t, consoleCalls, 1)
+}
+
+func merge(a, b map[string]interface{}) map[string]interface{} {
```
samsalisbury: Looking at usages of this, it seems overkill, as it would be possible to just add additional pairs to the map (see note above).

ext/singularity/deployermessage_test.go
```diff
@@ -1,24 +1,216 @@
...
 	"testing"
 
+	"github.com/opentable/sous/lib"
 	"github.com/opentable/sous/util/logging"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+var requestID = "12345"
+var defaultExpectedFields = map[string]interface{}{
```
samsalisbury: State like this that's shared between tests is generally best avoided to maintain isolation. I would make `defaultExpectedFields` a `func() map[string]interface{}` that returns a pristine copy of this map to each test that wants it.

ext/singularity/deployermessage_test.go
```diff
@@ -1,24 +1,216 @@
...
+func TestDeployerMessageError(t *testing.T) {
+	logger, control := logging.NewLogSinkSpy()
+	pair := baseDeployablePair()
+	taskData := &singularityTaskData{
+		requestID: requestID,
+	}
+	err := errors.New("Test error")
+	expectedFields := merge(defaultExpectedFields, map[string]interface{}{
+		"sous-error": "Test error",
+	})
```
samsalisbury: Here we rely on merge creating a _copy_ of the defaultExpectedFields map and adding `sous-error` to the copy. It would seem simpler to get a pristine copy of defaultExpectedFields (using func noted above), and just add a pair to the map, e.g.:

    expectedFields := defaultExpectedFields()
    expectedFields["sous-error"] = "Test error"

This would avoid the complexity of the merge function and make the test a bit quicker to understand to me...

lachlanh: great feedback, much cleaner way of doing what I doing, will make the change.

lachlanh:
@samsalisbury have made the changes as suggested, take a look and let me know if you have further thoughts, going to dismiss the review for now, so we can merge if others are ok with it. @JC1738 @nyarly

PR #582, opened 2018-02-07
SOUS-792 read-only rectification endpoints

samsalisbury:
This large PR adds new rectification queue endpoints, as per SOUS-792.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 1579 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

lib/r11n_queue.go
```diff
@@ -2,6 +2,7 @@ package sous
 
 import (
 	"container/ring"
+	"sort"
 	"sync"
 
 	"github.com/pborman/uuid"
```
nyarly: We have 3 different uuid packages now :(

satori/uuid.go, pborman/uuid and docker/distribution/uuid.

samsalisbury: Yikes, I'll add a ticket to address that in a future PR, since it wasn't added in this one. Sorry about the size of this PR btw, should have broken that story down more... Any blockers to merge @nyarly ?

util/restful/routemap.go
```diff
@@ -149,26 +150,61 @@ func (rm *RouteMap) SingleExchanger(factory ExchangeFactory, gf func() Injector,
...
+		if e.Name == name {
+			return e, true
+		}
+	}
+	return routeEntry{}, false
+}
+
+// URIFor returns the URI (relative to the root) for the named route using
+// pathParams and kv to fill out path parameters and query values respectively.
+func (rm RouteMap) URIFor(name string, pathParams map[string]string, kvs ...KV) (string, error) {
```
samsalisbury: This replaces `PathFor` and adds support for route/path parameters. We are not yet using path parameters because I had to pivot this PR to avoid them when shortcomings of httprouter's handling of encoded slashes in path segments were realised. However the old `PathFor` was untested as far as I could tell, so at least this feature is now covered extensively in `routemap_test.go`, and should make it easy to use path params in future.

server/handle_r11n.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Exchange returns the targeted r11nResponse and 200 if it exists, other
+// non-200 responses otherwise.
+func (h *GETR11nHandler) Exchange() (interface{}, int) {
+	if h.DeploymentIDErr != nil {
+		return nil, http.StatusNotFound
+	}
+	// Note that all queries and waiting should be done using the QueueSet
+	// itself, not the rectification.
```
nyarly: I'm not sure I understand this comment. If that's so, shouldn't the associated fields of the rectification be unexported?

server/handle_deployqueue.go
```diff
@@ -0,0 +1,77 @@
...
+		DeploymentIDErr error
+	}
+)
+
+func newDeployQueueResource(ctx ComponentLocator) *DeployQueueResource {
+	return &DeployQueueResource{context: ctx}
+}
+
+func deploymentIDFromRoute(r *http.Request) (sous.DeploymentID, error) {
+	didStr, err := url.QueryUnescape(r.URL.Query().Get("DeploymentID"))
```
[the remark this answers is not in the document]
JC1738: What is the alternative @nyarly, we do own both sides of the pipe.  So can pass anything.  Agree on that, though for the other consumers, while good to think about, not sure that is a big weighing factor.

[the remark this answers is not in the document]
nyarly: The alternative would be to use query parameters for the parts of the DeploymentID and lean on `net/http` to parse them.

PR #583, opened 2018-02-08
GET|PUT /state/deployments

nyarly:
There are a few points of interest in this PR,
but it's currently broken. Setting it down to address logging.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Debugging output: +	spew.Dump(“ExcludeMe”, ex)<br />
 (there may be others)</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 514 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

lib/fixtures.go
```diff
@@ -33,11 +51,30 @@ func deployConfigFixture(name string) DeployConfig {
 	}
 }
 
-func deploymentFixture(name string) *Deployment {
+// DeploymentFixture returns a fixtured Deployment. Stereotypes:
+// * (default) a repeating "example" repo
+// * sequenced-repo - uses sequences to return distinct deployments
+func DeploymentFixture(name string) *Deployment {
```
samsalisbury: Neat idea, although because the results are dynamic with "sequenced-repo", if we want to know the repo that was set (e.g. to assert on the identity of deployments later on) we now need to dig through the returned `Deployment`. Maybe it would be clearer to call this argument "repo" and just use whatever is passed as the repo name? At least then the identity is known a priori in the test. Alternatively, to maintain the pattern, could add another argument "repo" for this purpose, and allow the stereotype to still set other fixed values.

server/data_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	sous "github.com/opentable/sous/lib"
+)
+
+func TestGDMWrapperAddHeaders(t *testing.T) {
+	wrapper := GDMWrapper{
+		Deployments: []*sous.Deployment{
+			sous.DeploymentFixture("sequenced-repo"),
+			sous.DeploymentFixture("sequenced-repo"),
+			sous.DeploymentFixture("sequenced-repo"),
```
samsalisbury: From these usages, I can see that we don't care about identity of the deployments returned from DeploymentFixture, so no need to add this yet. However, for reuse of DeploymentFixture later it may be better to add the repo explicit now to avoid breaking tests later when it may become necessary...

server/data_test.go
```diff
@@ -0,0 +1,27 @@
...
+		},
+	}
+
+	headers := http.Header{}
+
+	wrapper.AddHeaders(headers)
+
+	if strings.Index(headers.Get("Etag"), "w/") != 0 {
```
samsalisbury: Could use `strings.HasPrefix` for concision?

server/handle_gdm.go
```diff
@@ -70,8 +68,6 @@ func (h *GETGDMHandler) Exchange() (interface{}, int) {
 		}
 		data.Deployments = append(data.Deployments, d)
 	}
-	etag, _ := h.GDM.GetEtag()
-	h.RzWriter.Header().Set("Etag", etag)
```
samsalisbury: Since we are now having headers added by checking the returned thing for `AddHeaders` (in `restful` I presume?) I wonder if we should also have the status code determined by the same process, to keep all HTTP handling in one place? E.g. `return nil` could map to a 404, `return error` to a 500 etc. Or alternatively centralise all HTTP handling in the handlers like it was before? I lean towards the latter approach as it is more explicit and flexible...

server/handle_state_deployments_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	ex := &PUTStateDeployments{
+		cluster:     cm,
+		clusterName: "test-cluster",
+		req:         req,
+	}
+
+	data, status := ex.Exchange()
+
+	if status != 202 {
```
samsalisbury: tupo?

test/server_integration_test.go
```diff
@@ -134,6 +136,17 @@ func (suite integrationServerTests) TestUpdateServers() {
...
 }
 
+func (suite integrationServerTests) TestUpdateStateDeployments() {
+	data := server.GDMWrapper{Deployments: []*sous.Deployment{}}
+	err := suite.client.Create("./state/deployments", nil, &data, nil)
+	suite.Error(err, `412 Precondition Failed: "resource present for If-None-Match=*!\n"`)
+
+	updater, err := suite.client.Retrieve("./state/deployments", nil, &data, nil)
+	suite.NoError(err)
+	suite.Equal(data, "a rabbit")
```
samsalisbury: I can't clearly see why this assertion should pass? Where does that string come from?

nyarly:
@samsalisbury I'm coming to the end of my last day for a week. This branch currently fails tests by timing out. All the paniced goroutines are in Psyringe - maybe they've deadlocked?

Could you check this out and see if you can get the tests to at least fail without timeout?

samsalisbury:
Hey @nyarly, got back to this this last night/this morning - those tests now fail appropriately. I've had to add to the graphs bits that are later added again in the actions in order to be able to run the tests. This is a little tricky and will cause tests to break as more items are added in the actions themselves. Trying to think of a better pattern, but at least for now you should be able to make progress.

cli/cli_test.go
```diff
@@ -22,6 +23,15 @@ func prepareCommand(t *testing.T, cl []string) (*CLI, *cmdr.PreparedExecution, f
 
 	s := &Sous{Version: semv.MustParse(`1.2.3`)}
 	di := graph.BuildTestGraph(semv.Version{}, stdin, stdout, stderr)
+	di.Add(&config.DeployFilterFlags{},
+		&config.PolicyFlags{},
+		graph.DryrunBoth,
+		&config.Verbosity{},
+		&config.OTPLFlags{},
+	)
+	if err := di.Test(); err != nil {
```
nyarly: @samsalisbury I'm carrying this forward, but wondering if di.Test() is actually appropriate here. There are a number of cases where we have partially provided graphs, because we don't need, e.g. `PolicyFlags` for most commands. `di.Test()` demands that we artificially `Add` them. Arguably, there's a concrete detriment to our tests, since we wind up testing a circumstance that is concretely different than the graph used in production.

cli/cli_test.go
```diff
@@ -22,6 +23,15 @@ func prepareCommand(t *testing.T, cl []string) (*CLI, *cmdr.PreparedExecution, f
 
 	s := &Sous{Version: semv.MustParse(`1.2.3`)}
 	di := graph.BuildTestGraph(semv.Version{}, stdin, stdout, stderr)
+	di.Add(&config.DeployFilterFlags{},
+		&config.PolicyFlags{},
+		graph.DryrunBoth,
+		&config.Verbosity{},
```
nyarly: @samsalisbury  Adding `Verbosity` here is now breaking `TestInvokeCommand` because of a double add. I'm not sure what changed that this is the case. Taking `Verbosity` out of this helper breaks the rest of the tests.

cli/cli_test.go
samsalisbury: I see what you mean, made a comment that they should likely be removed in the commit where I added them.

Because we are now passing the graph around and adding to it late, we can't always expect a complete graph in all tests. With that in mind I've made a quick change to psyringe that prevents adding any constructor that would cause a dependency cycle. I initially wanted to avoid doing that because of the cost of checking, however it makes sense as any dependency cycle will cause `Inject` to deadlock, so we never ever want that in the graph, even momentarily. The `Test` method still checks for graph completeness, but we can use that only in places where we do expect a complete graph. I'll update this PR now...

samsalisbury: OK @nyarly I've pushed [this branch (state_deployments_endpoint_nocycle)](https://github.com/opentable/sous/compare/state_deployments_endpoint_nocycle)  which resets back to 4ed6dd5 and vendors psyringe which fails on Adding any dependency cycle. (Thus those Test calls are gone, along with all the extra things added to make them pass.) I think the best thing might be to reset this branch to match that one so we preserve PR commentary all in one place. Tests now fail purely because of the dependency cycle, so fixing it doesn't also involve making graphs complete and should be a simpler job... https://github.com/opentable/sous/compare/state_deployments_endpoint_nocycle

cli/cli_test.go
samsalisbury: Hmm, that's frustrating, I think we have a good solution now and we can get rid of these extra calls, and most of the Test calls. See [comment above.](https://github.com/opentable/sous/pull/583#discussion_r169930399)

server/handle_state_deployments_test.go
nyarly: 202 is appropiate here, I think.

server/handle_gdm.go
nyarly: Mapping application returns to status codes is tough to do well. Many responses should have an empty body, and nil makes sense for that. A non-zero error doesn't necessarily mean that the cause of the error is server-side.

Conversely, there's a *lot* of a well-formed HTTP exchange that can be extracted from application logic. A default Etag is one of those things. For a really expansive demonstration check out one of the implementations of webmachine (I've worked with the Erlang and Ruby ones).

In other words, there's a lot of HTTP processing that the application logic shouldn't be responsible for, but there's also a lot of semantics that the application has to have some control over. The conclusion I draw is that centralizing the HTTP handling is going to be problematic no matter where you put it.

util/logging/messages.go
```diff
@@ -58,15 +58,15 @@ type (
 
 	*/
 
-	temporaryOldStyleLogging interface {
+	TemporaryOldStyleLogging interface {
```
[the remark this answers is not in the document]
nyarly: In this case, TOSL is on its way out, and I'd rather hear Hound complain about it until we can remove it entirely.

server/handle_state_deployments_test.go
samsalisbury: Ah, the line below if the typo then :)

samsalisbury: I'll fix that on merge @nyarly ^

test/http_state_manager_test.go
```diff
@@ -79,21 +78,32 @@ func TestWriteState(t *testing.T) {
...
 	)
 
 	serverScoop := struct{ Handler graph.ServerHandler }{}
 	di.Add(&config.Verbosity{})
+	di.Add(sous.User{})
+	di.Add(&config.PolicyFlags{})
+	di.Add(&config.OTPLFlags{})
+	if err := di.Test(); err != nil {
+		t.Fatalf("invalid graph: %s", err)
+	}
```
samsalisbury: This test call is probably not useful here, as you rightly noticed. Since we do not care about completeness in tests, only that the graph is complete enough to perform the injection asked of it. In case it's not then `MustInject` will panic so we'll know. The only thing we really care about is testing for cycles (which always cause deadlocks). Since Add now panics when creating a cycle, the Test call is most certainly redundant. This is why I recommended **resetting** this branch back to the alternate one I opened to lop off these unwanted changes. Since work has been done over the top of them now, I'll go ahead and tidy this up...

test/server_test.go
```diff
@@ -64,10 +64,15 @@ func (suite serverTests) prepare(extras ...interface{}) http.Handler {
...
 	g.Add(&config.DeployFilterFlags{})
 	g.Add(graph.DryrunBoth)
+	g.Add(&config.PolicyFlags{})
+	g.Add(&config.OTPLFlags{})
 
 	serverScoop := struct {
 		Handler graph.ServerHandler
 	}{}
+
+	suite.Require().NoError(g.Test())
```
samsalisbury: Again, I think this `Test` call should be lopped off.

util/logging/messages.go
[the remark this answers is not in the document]
samsalisbury: I believe Hound will no longer complain once this is merged.

util/logging/messages/clienthttpresponse.go
```diff
@@ -85,7 +100,14 @@ func buildHTTPLogMessage(server bool, rq *http.Request, statusCode int, response
 	return m
 }
 
-func newHTTPLogEntry(server bool, resName, method, urlstring string, status int, rqSize, rzSize int64, dur time.Duration) *HTTPLogEntry {
+func newHTTPLogEntry(
```
samsalisbury: Hmm, a 10 param func with 2 bools and lots of ints looks like a foot gun waiting to happen... I know this was mostly already here, not really part of this PR, but for readability and to avoid programmer error I would favour at least putting the bools and ints into a struct to be passed in together so they benefit from field labels.

util/logging/messages/messages_test.go
```diff
@@ -9,60 +9,165 @@ import (
...
+			"url-pathname":    "/api",
+			"url-querystring": "a=a",
+			"duration":        int64(30000),
+			"body-size":       int64(0),
+			"response-size":   int64(123),
+			"status":          200,
+		})
+}
+
+func TestReportClientHRequestFields(t *testing.T) {
```
samsalisbury: H=HTTP? For the sake of 3 chars I think it's worth expanding to avoid cognitive load on the reader.

samsalisbury: Same goes for usages below. I'll make these changes on this branch.

samsalisbury: Another reason to expand these to match the name of the function under test is that it makes `go test -run XXX` more intuitive, as I can now pass the name of the function I want to run tests for rather than a derivative name.

util/logging/testsupport.go
```diff
@@ -151,6 +154,32 @@ func (wds writeDonerSpy) Done() {
...
+			lm.EachField(func(name string, val interface{}) {
+				if name == "call-stack-trace" {
+					return
+				}
+				line = line + fmt.Sprintf("%s=%v ", name, val)
+			})
+		}
+		t.Log(line)
+	}
+}
```
samsalisbury: 👍

util/restful/httpclient_test.go
```diff
@@ -96,16 +97,28 @@ func TestClientRetrieve(t *testing.T) {
...
 	var variableFields []string
-	variableFields = append(logging.StandardVariableFields, logging.HTTPVariableFields...)
-	// XXX call-stack-function is actually a problem here.
-	variableFields = append(variableFields, "channel_name", "call-stack-function")
+	variableFields = append(
+		logging.StandardVariableFields,
+		"url",          // depends on the httptest server's random port
+		"url-hostname", // likewise
+		"duration",
+		"call-stack-function", // live log entries are squirrelly
```
samsalisbury: Haha, just spat out my coffee... I should get out more :P so we can't be sure what this value will end up as? That seems totally fine, better not to have it fixed anyway as would mean code-level changes breaking tests. We should probably not rely on `call-stack-function` to be a stable identifier anyway, but instead be explicit with fields that identify important messages.

server/data_test.go
samsalisbury: Making this change on the branch now.

sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 1487 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

samsalisbury:
Approved my own changes... Let's go through this and hopefully get it merged in standup @nyarly

PR #588, opened 2018-02-16
fix https://opentable.atlassian.net/browse/SOUS-874

JC1738:
add back the + check in front of file, not checking comment, will add that later if we really need it.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -53,7 +53,7 @@ def check_for_debug
     file.path =~ /_test.go$/
   end.each do |file|
     file.patch.each_line do |patch_line|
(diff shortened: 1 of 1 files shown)


sous-machine-account:
LGTM

</p>

Dangerfile
```diff
@@ -53,7 +53,7 @@ def check_for_debug
     file.path =~ /_test.go$/
   end.each do |file|
     file.patch.each_line do |patch_line|
-      if /[^(\/\/)](spew\.Pr|spew\.Du|spew.\Fp|spew\.Fd)/ =~ patch_line
+      if /^\+[^+].*(spew\.Pr|spew\.Du|spew.\Fp|spew\.Fd)/ =~ patch_line
```
samsalisbury: Maybe `&&` this with the removed commented out check to get the best of both worlds?

PR #597, opened 2018-02-26
General Logging

JC1738:
Update the general logger to have a way to auto extract IDs to variables passed to it.  They can be stored in separate fields and be easier to search by in logstash.


sous-machine-account:
LGTM

</p>

PR #607, opened 2018-02-28
fix SOUS-887

JC1738:
the objects passing in http are too large... which we should talk about seperately, it's 2 mb we are going over wire, this will eventually be too much for http protocol


sous-machine-account:
LGTM

</p>

PR #609, opened 2018-03-01
Postgres dedupe behaves correctly now.

nyarly:


sous-machine-account:
LGTM

</p>

util/restful/router.go
```diff
@@ -60,8 +61,15 @@ type (
...
-			// need to log this here, or we lose the real stack
-			xlog.LogSink.Warnf("%+#v\n%s", p, string(debug.Stack()))
+			if pe, is := p.(error); is {
+				url := "<unknown>"
+				if xlog.Request != nil {
+					url = xlog.Request.RequestURI
+				}
+				logging.ReportError(xlog.LogSink, errors.Wrapf(pe, "%q\n%s", url, string(debug.Stack())))
+			} else {
+				messages.ReportLogFieldsMessage("Panic while processing request", logging.WarningLevel, xlog.LogSink, p, xlog.Request)
```
JC1738: Any size consideration for logging xlog.Request?  Now new code will trace out the truncated message to stderr so singularity will pick it up, so should be fine if it is too large will find out.

nyarly: The URL and stacktrace should be within the limit. We log more than that routinuely.

The whole Request is probably pushing it though... Ultimately, I don't think that branch ever gets followed, though.

util/sqlgen/fieldset_test.go
```diff
@@ -27,3 +27,22 @@ func TestRowDedupe(t *testing.T) {
 		t.Errorf("Expected first value to be 1, got %v", vals[1])
 	}
 }
+
+func TestRowDedupe_without_candidates(t *testing.T) {
```
samsalisbury: A test combining candidate with non-candidate fields would be useful as well, testing additional behaviour.

PR #613, opened 2018-03-01
Fixed all-deploys

nyarly:
The /all-deploy-queues endpoint was trying to serialize a map[sous.DeploymentID]int.
This was a problem because DeploymentID is a struct - which Go allows as map keys,
but JSON does not allow as object field names. There was the further issue that,
if simply converted to strings, clients would need to reconstruct the DeploymentID
to retrieve the specific queue.

Therefore, the structure of the response JSON changed. It returns an object keyed
by deploymentID strings, but where the values are objects with the full DeploymentID
included.

The change in the data structure meant that the tests needed to be overhauled,
and I wound up unrolling the complex table test driver.

To facilitate debugging, I refined the logging in the restful package to ensure
that we get logging of HTTP exchanges in some corner cases.

### DIFF
--- a/server/handle_alldeployqueues.go
+++ b/server/handle_alldeployqueues.go
@@ -34,12 +34,14 @@ func (r *AllDeployQueuesResource) Get(_ http.ResponseWriter, _ *http.Request, _ 
 // Exchange returns deploymentsResponse representing all queues managed by this
 // server instance.
 func (h *GETAllDeployQueuesHandler) Exchange() (interface{}, int) {
+	data := DeploymentQueuesResponse{Queues: map[string]QueueDesc{}}
+
 	queues := h.QueueSet.Queues()
-	m := map[sous.DeploymentID]int{}
 	for did, q := range queues {
-		m[did] = q.Len()
+		data.Queues[did.String()] = QueueDesc{
+			DeploymentID: did,
+			Length:       q.Len(),
+		}
 	}
-	return DeploymentQueuesResponse{
-		Deployments: m,
-	}, 200
+	return data, 200
 }
--- a/server/handle_alldeployqueues.go
+++ b/server/handle_alldeployqueues.go
@@ -39,11 +39,7 @@ func (h *GETAllDeployQueuesHandler) Exchange() (interface{}, int) {
 	for did, q := range queues {
 		m[did] = q.Len()
 	}
-	return deploymentsResponse{
+	return DeploymentQueuesResponse{
 		Deployments: m,
 	}, 200
-}
-
-type deploymentsResponse struct {
-	Deployments map[sous.DeploymentID]int
 }
--- a/server/handle_alldeployqueues_test.go
+++ b/server/handle_alldeployqueues_test.go
@@ -26,94 +26,96 @@ func TestNewAllDeployQueuesResource(t *testing.T) {
 }
 
 func TestGETAllDeployQueuesHandler_Exchange(t *testing.T) {
+	t.Run("empty queue", func(t *testing.T) {
+		data, status := setupExchange(t)
+		assertSuccess(t, status)
+		dqr := assertIsDeploymentQueuesResponse(t, data)
+		assertQueueLength(t, dqr, sous.DeploymentID{}, 0)
+		assertLength(t, dqr, 0)
+	})
 
-	testCases := []struct {
-		desc       string
-		dids       []sous.DeploymentID
-		wantResult DeploymentQueuesResponse
-	}{
-		{
-			desc:       "empty queue",
-			dids:       []sous.DeploymentID{},
-			wantResult: DeploymentQueuesResponse{Deployments: map[sous.DeploymentID]int{}},
-		},
-		{
-			desc: "one DeploymentID",
-			dids: []sous.DeploymentID{newDid("one")},
-			wantResult: DeploymentQueuesResponse{Deployments: map[sous.DeploymentID]int{
-				newDid("one"): 1,
-			}},
-		},
-		{
-			desc: "two unique DeploymentIDs",
-			dids: []sous.DeploymentID{newDid("one"), newDid("two")},
-			wantResult: DeploymentQueuesResponse{Deployments: map[sous.DeploymentID]int{
-				newDid("one"): 1,
-				newDid("two"): 1,
-			}},
-		},
-		{
-			desc: "same DeploymentID twice",
-			dids: []sous.DeploymentID{newDid("one"), newDid("one")},
-			wantResult: DeploymentQueuesResponse{Deployments: map[sous.DeploymentID]int{
-				newDid("one"): 2,
-			}},
-		},
+	t.Run("one DeploymentID", func(t *testing.T) {
+		data, status := setupExchange(t, newDid("one"))
+		assertSuccess(t, status)
+		dqr := assertIsDeploymentQueuesResponse(t, data)
+		assertQueueLength(t, dqr, sous.DeploymentID{}, 0)
+		assertLength(t, dqr, 1)
+		assertQueueLength(t, dqr, newDid("one"), 1)
+	})
+
+	t.Run("two unique DeploymentIDs", func(t *testing.T) {
+		data, status := setupExchange(t, newDid("one"), newDid("two"))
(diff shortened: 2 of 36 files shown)


sous-machine-account:
LGTM

</p>

shell.nix
```diff
@@ -1,5 +1,5 @@
-#{ pkgs ? import ~/dev/nixpkgs {} }:
-{ pkgs ? import <nixpkgs> {} }:
+{ pkgs ? import ~/dev/nixpkgs {} }:
+#{ pkgs ? import <nixpkgs> {} }:
```
JC1738: @nyarly Not sure I understand this change or what this file does?

util/restful/statusmiddleware.go
```diff
@@ -45,22 +45,11 @@ func (ph *StatusMiddleware) errorBody(status int, rq *http.Request, w io.Writer,
...
-	if status >= 400 {
-		messages.ReportLogFieldsMessage("Status >= 400", logging.WarningLevel, ph.LogSink)
-		ph.errorBody(status, r, w, data, nil, nil)
-	}
-	if status >= 200 && status < 300 {
-		messages.ReportLogFieldsMessage("Status >= 200 && Status < 300", logging.DebugLevel, ph.LogSink)
-	}
-	// XXX in a dev mode, print the panic in the response body
-	// (normal ops it might leak secure data)
+	ph.errorBody(status, r, w, data, nil, nil)
```
JC1738: Does this always suppose to call errorBody no matter what the status?

server/handle_alldeployqueues_test.go
```diff
@@ -26,96 +26,98 @@ func TestNewAllDeployQueuesResource(t *testing.T) {
...
+func assertIsDeploymentQueuesResponse(t *testing.T, data interface{}) DeploymentQueuesResponse {
+	dr, ok := data.(DeploymentQueuesResponse)
+	if !ok {
+		t.Fatalf("got a %T; want a %T", data, dr)
+		return DeploymentQueuesResponse{}
+	}
+	return dr
+}
+
+func assertLength(t *testing.T, dr DeploymentQueuesResponse, wantLen int) {
```
samsalisbury: "assertNumQueues" would be more specific.

server/handle_alldeployqueues_test.go
```diff
@@ -26,96 +26,98 @@ func TestNewAllDeployQueuesResource(t *testing.T) {
...
+	gotLen := len(dr.Queues)
+	if gotLen != wantLen {
+		t.Fatalf("got %d queued deployments; want %d", gotLen, wantLen)
+	}
+}
+
+func assertQueueLength(t *testing.T, dr DeploymentQueuesResponse, did sous.DeploymentID, wantCount int) {
+	gotCount := dr.Queues[did.String()].Length
+	if gotCount != wantCount {
+		t.Errorf("got %d queued rectifications for %q; want %d", gotCount, did.String(), wantCount)
```
samsalisbury: `%q` verb already calls `.String()` so we can drop that explicit call here.

server/handle_alldeployqueues_test.go
```diff
@@ -26,96 +26,98 @@ func TestNewAllDeployQueuesResource(t *testing.T) {
...
+
+func assertSuccess(t *testing.T, status int) {
+	const wantStatusCode = 200
+	if status != wantStatusCode {
+		t.Errorf("got %d; want %d", status, wantStatusCode)
 	}
 
 }
 
+func assertIsDeploymentQueuesResponse(t *testing.T, data interface{}) DeploymentQueuesResponse {
```
samsalisbury: Add `t.Helper()`

server/handle_alldeployqueues_test.go
```diff
@@ -26,96 +26,98 @@ func TestNewAllDeployQueuesResource(t *testing.T) {
...
+}
+
+func assertLength(t *testing.T, dr DeploymentQueuesResponse, wantLen int) {
+	gotLen := len(dr.Queues)
+	if gotLen != wantLen {
+		t.Fatalf("got %d queued deployments; want %d", gotLen, wantLen)
+	}
+}
+
+func assertQueueLength(t *testing.T, dr DeploymentQueuesResponse, did sous.DeploymentID, wantCount int) {
```
samsalisbury: Add `t.Helper()`

server/handle_alldeployqueues_test.go
```diff
@@ -26,96 +26,98 @@ func TestNewAllDeployQueuesResource(t *testing.T) {
...
-
-			testCount := dr.Deployments[sous.DeploymentID{}]
-			if testCount != 0 {
-				t.Errorf("got %d for empty DeploymentID expected 0", testCount)
-			}
-		})
+	return handler.Exchange()
+}
+
+func assertSuccess(t *testing.T, status int) {
```
samsalisbury: "assertStatusCode" would be more specific. Also add `t.Helper()` to get a more useful line number.

server/handle_alldeployqueues.go
```diff
@@ -34,16 +34,14 @@ func (r *AllDeployQueuesResource) Get(_ http.ResponseWriter, _ *http.Request, _
...
+	data := DeploymentQueuesResponse{Queues: map[string]QueueDesc{}}
+
 	queues := h.QueueSet.Queues()
-	m := map[sous.DeploymentID]int{}
 	for did, q := range queues {
-		m[did] = q.Len()
+		data.Queues[did.String()] = QueueDesc{
+			DeploymentID: did,
+			Length:       q.Len(),
+		}
```
samsalisbury: Nice, we should probably add links out to the individual queue endpoints as well as an enhancement; not part of this PR though.

test/server_integration_test.go
```diff
@@ -194,6 +194,14 @@ func (suite integrationServerTests) TestUpdateStateDeployments_Update() {
 	suite.Len(data.Deployments, 3)
 }
 
+func (suite integrationServerTests) TestGetAllDeployQueues() {
+	data := server.DeploymentQueuesResponse{}
+	updater, err := suite.client.Retrieve("./all-deploy-queues", nil, &data, nil)
+	suite.NoError(err)
+	suite.Len(data.Queues, 0)
+	suite.NotNil(updater)
+}
```
samsalisbury: Thanks for adding this, I'll see if I can expand this a bit to check for the non-zero case as well.

server/handle_alldeployqueues_test.go
```diff
@@ -26,96 +26,98 @@ func TestNewAllDeployQueuesResource(t *testing.T) {
...
-		{
-			desc: "same DeploymentID twice",
-			dids: []sous.DeploymentID{newDid("one"), newDid("one")},
-			wantResult: deploymentsResponse{Deployments: map[sous.DeploymentID]int{
-				newDid("one"): 2,
-			}},
-		},
+}
+
+func setupExchange(t *testing.T, dids ...sous.DeploymentID) (interface{}, int) {
```
samsalisbury: I'll rename this `setupAllDeployQueuesExchange` as there are a lot of methods named `Exchange`.  Also add `t.Helper()` for better failure line numbers.

util/restful/router.go
```diff
@@ -129,45 +130,45 @@ func (mh *MetaHandler) genericHandling(resName string, factory ExchangeFactory,
 // GetHandling handles Get requests.
 func (mh *MetaHandler) GetHandling(resName string, factory ExchangeFactory) httprouter.Handle {
 	return func(rw http.ResponseWriter, r *http.Request, p httprouter.Params) {
-		rw, data, status := mh.genericHandling(resName, factory, rw, r, p)
-		rw.Header().Add("Access-Control-Allow-Origin", "*") //XXX configurable by app
-		mh.renderData(status, rw, r, data)
+		lrw, data, status := mh.genericHandling(resName, factory, rw, r, p)
```
samsalisbury: Did this fix an issue, or just avoiding shadowing?

util/restful/statusmiddleware.go
samsalisbury: Agree this looks confusing, `errorBody` seems to write an error body no matter what the status as well.

PR #618, opened 2018-03-02
Synchronizing GCI issue

nyarly:


sous-machine-account:
LGTM

</p>

util/logging/callerinfo.go
```diff
@@ -35,16 +36,18 @@ type (
...
 // It's very common to call this like:
 //   GetCallerInfo(logging.NotHere())
 // which excludes the function that actually calls GetCallerInfo
 // (usually a message constructor) from the logged call stack.
 func GetCallerInfo(excluding ...Excluder) CallerInfo {
-	if !notGCI.sensible {
-		notGCI = NotHere().(excludedFrame)
-	}
+	gci := NotHere().(excludedFrame)
+	gciOnce.Do(func() {
```
JC1738: isn't once preventing the same goroutine from resetting where the call stack is on subsequent logging calls?

nyarly: The purpose of the notGCI thing is to exclude GetCallerInfo from the logged call stack. Once we've captured that exclusion, there's no need to exclude it subsequent times.

JC1738: But isn’t there multiple calls with different locations?  What’s the scope of once?  By goroutine?

nyarly: There definitely are. This snippet is about excluding GetCallerInfo itself from the stack so that we can accurately report where it is called from.

PR #621, opened 2018-03-06
remove postgres for now, since don't have that on TC, change how get …

JC1738:
…packages to also use docker


sous-machine-account:
LGTM

</p>

JC1738:
@samsalisbury you are correct, they will fail, but after this PR I will work in dockerizing the the postgres stuff

PR #627, opened 2018-03-08
#trivial Migrated postgres and liquibase to Docker Files

JC1738:


JC1738:
whops accidentally merged

PR #628, opened 2018-03-08
lib: bandage close of closed channel #trivial

samsalisbury:
- The close(done) was panicking occasionally even though I cannot
  find any other reference to that channel.
- Could be a bug in sync.Once or something weirder (is the once being
  copied in a way that breaks the pattern somewhere?).


sous-machine-account:
LGTM

</p>

PR #629, opened 2018-03-08
Hit specific server + return all query params in Location from PUT /single-deployment #trivial

samsalisbury:
'sous newdeploy' now selects the appropriate server to issue PUT /single-deployment to. Previously it always tried to put the the server in user's config which is not always appropriate.

Also fixed: PUT /single-deployment returns a Location header pointing to a rectification which can be watched for completion. However, the URL in Location header was missing DeploymentID query params, so the R11n could never be found. This PR adds the remaining query params so this relative URL is complete.

### DIFF
--- a/cli/sous_newdeploy.go
+++ b/cli/sous_newdeploy.go
@@ -18,7 +18,7 @@ import (
 type SousNewDeploy struct {
 	DeployFilterFlags config.DeployFilterFlags `inject:"optional"`
 	StateReader       graph.StateReader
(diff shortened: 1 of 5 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

PR #636, opened 2018-03-09
#trivial Update queue to inspect BuildArtifact, added it when not present

JC1738:
Debugging code, noticed buildArtifact error message, this is an attempt to put it in.

### DIFF
--- a/lib/r11n_queue_set.go
+++ b/lib/r11n_queue_set.go
@@ -19,6 +19,7 @@ type (
 	R11nQueueSet struct {
 		set  map[DeploymentID]*R11nQueue
 		opts []R11nQueueOpt
+		reg  Registry
 		sync.RWMutex
 	}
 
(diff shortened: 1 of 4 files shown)


sous-machine-account:
LGTM

</p>

lib/r11n_queue_set.go
```diff
@@ -55,6 +57,11 @@ func (rqs *R11nQueueSet) PushIfEmpty(r *Rectification) (*QueuedR11n, bool) {
 func (rqs *R11nQueueSet) Push(r *Rectification) (*QueuedR11n, bool) {
 	rqs.Lock()
 	defer rqs.Unlock()
+
+	if r.Pair.Post.BuildArtifact == nil {
```
nyarly: The problem with putting this here is that we'll block on inserting the new deployment until we can get its Docker name, which can be time consuming.

We could merge this, but we should follow up by moving it into the queue handler

JC1738: @nyarly Think I understand yes, we are just trying to get all the bits in somewhere so can test whatelse is missing.  Added below write to GDM since that is also missing, can you take a look at that.  SOmething is not quite right, we get a Write error

samsalisbury:
Closing in favour of #637

PR #639, opened 2018-03-12
Fixes sous-without-workspace issue.

nyarly:
It's been a longstanding bug that when e.g. `sous query gdm` or
`sous manifest get -repo ...` was run outside of a git repo, it would crash
with a git error. This PR fixes that, since those use cases are pretty useful.

Also cleans up some logging & console interactions.

### DIFF
--- a/util/logging/messages/generalmsg.go
+++ b/util/logging/messages/generalmsg.go
@@ -329,7 +329,7 @@ func (l logFieldsMessage) EachField(fn logging.FieldReportFn) {
 	}
 
 	if l.jsonObj != nil {
-		if n, err := l.jsonObj.ArrayCount("message", "array"); n > 0 && err != nil {
+		if n, err := l.jsonObj.ArrayCount("message", "array"); n > 0 && err == nil {
 			fn("json-value", l.jsonObj.String())
 		}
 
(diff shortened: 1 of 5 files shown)


sous-machine-account:
LGTM

</p>

util/logging/messages/generalmsg.go
```diff
@@ -192,12 +192,6 @@ type logFieldsMessage struct {
 func (l logFieldsMessage) WriteToConsole(console io.Writer) {
 	if l.console {
 		fmt.Fprintf(console, "%s\n", l.composeMsg())
-		if l.jsonObj != nil {
-			fmt.Fprintf(console, "%s\n", l.jsonObj.StringIndent("", " "))
-		}
-
-		fmt.Fprintf(console, "Fields: %s\n", strings.Join(l.Fields, ","))
-		fmt.Fprintf(console, "Types: %s\n", strings.Join(l.Types, ","))
 	}
```
JC1738: the interesting thing about this is while it's good not to have this in the Command output, server side write to console looking at singularity logs will no longer show the fidelity of the object.  It's almost like we need a ServerWriteToConsole that will trigger a json object to be outputed to console.  I think I will add that in a subsequent pull request.

PR #640, opened 2018-03-13
`sous server -autoresolver=false`

nyarly:
Adds a flag to sous server to disable the autoresolver.
This is only intended to be used for local debugging efforts,
so it has not been made a configuration option.

There's also a couple of logging cleanups and linter satisfaction changes.

### DIFF
--- a/lib/resolver.go
+++ b/lib/resolver.go
@@ -43,25 +43,24 @@ func (r *Resolver) queueDiffs(dcs *DeployableChans, results chan DiffResolution)
 	var wg sync.WaitGroup
 	for p := range dcs.Pairs {
 		sr := NewRectification(*p)
-		messages.ReportLogFieldsMessageWithIDs("Adding to queset", logging.ExtraDebug1Level, r.ls, p, sr)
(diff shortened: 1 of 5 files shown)


lib/resolver.go
```diff
@@ -43,25 +43,24 @@ func (r *Resolver) queueDiffs(dcs *DeployableChans, results chan DiffResolution)
...
-			result, ok := r.QueueSet.Wait(did, queued.ID)
+			result, ok := r.QueueSet.Wait(p.ID(), queued.ID)
 			if !ok {
-				messages.ReportLogFieldsMessageWithIDs("failed to QueueSet.Wait", logging.ExtraDebug1Level, r.ls, queued.ID, p, sr)
+				messages.ReportLogFieldsMessageWithIDs("Failed to QueueSet.Wait", logging.ExtraDebug1Level, r.ls, queued.ID, p, sr)
 				reportR11nAnomaly(r.ls, sr, r11nWentMissing)
 			}
 			results <- result
-		}()
+		}(p)
```
nyarly: Linter complained about the use of the captured p in the goroutine. Made it a parameter so that it would be present as a scope variable.

sous-machine-account:
LGTM

</p>

PR #645, opened 2018-03-14
server: /deploy-queue-item fixes

samsalisbury:
- Now returns DiffResolution and correct queue position.

### DIFF
--- a/server/handle_r11n_test.go
+++ b/server/handle_r11n_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	sous "github.com/opentable/sous/lib"
+	"github.com/stretchr/testify/assert"
 )
 
 // TestNewR11nResource checks that the same queue set passed to the
@@ -115,13 +116,6 @@ func TestR11nResource_Get_no_errors(t *testing.T) {
 // positions.
 func TestGETR11nHandler_Exchange_static(t *testing.T) {
 
-	wantNilResolution := func(t *testing.T, r r11nResponse) {
-		t.Helper()
-		if r.Resolution != nil {
-			t.Errorf("want nil Resolution; got %v", r.Resolution)
-		}
-	}
-
 	queues := sous.NewR11nQueueSet()
 	queuedOne1, ok := queues.Push(newR11n("one"))
 	if !ok {
@@ -136,6 +130,13 @@ func TestGETR11nHandler_Exchange_static(t *testing.T) {
 		t.Fatal("setup failed to push r11n")
 	}
 
+	wantNilResolution := func(t *testing.T, r r11nResponse) {
+		t.Helper()
+		assert.Nil(t, r.Resolution)
+		if r.Resolution != (*sous.DiffResolution)(nil) {
+			t.Errorf("want nil Resolution; got %v (%[1]T) %t", r.Resolution, (r.Resolution != (*sous.DiffResolution)(nil)))
+		}
+	}
 	wantStringReponse := func(t *testing.T, body interface{}, wantResponse string) {
 		t.Helper()
 		gotResponse, ok := body.(string)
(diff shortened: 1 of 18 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

server/handle_r11n_test.go
```diff
@@ -127,93 +129,282 @@ func TestGETR11nHandler_Exchange(t *testing.T) {
...
-			t.Errorf("got status %d; want %d", gotStatus, wantStatus)
-		}
-		gotPos := gotResponse.QueuePosition
-		const wantPos = 0
-		if gotPos != wantPos {
-			t.Errorf("got queue position %d; want %d", gotPos, wantPos)
-		}
+		a.wantStatus200(t, gotStatus)
+		gotResponse := a.wantR11nResponse(t, body)
+		a.wantQueuePos(t, gotResponse, -2)
```
nyarly: n.b. about 6% of the time, this gets -1 instead.
Try `go test ./server -run afterprocessing -count 100`

server/handle_single_deployment.go
```diff
@@ -136,7 +136,7 @@ func (psd *PUTSingleDeploymentHandler) Exchange() (interface{}, int) {
 		return psd.err(400, "Error parsing body: %s.", err)
 	}
 
-	messages.ReportLogFieldsMessageToServerConsole("Exchange PutSingleDeplymentHandler", logging.ExtraDebug1Level, log, did, psd.Body)
```
JC1738: @nyarly I liked the server console, since don't have to rotate out to kafka, and not much more overhead...

server/handle_r11n_test.go
```diff
@@ -127,93 +129,282 @@ func TestGETR11nHandler_Exchange(t *testing.T) {
...
+		gdh := &GETR11nHandler{
+			QueueSet:          queues,
+			DeploymentID:      newDid("two"),
+			R11nID:            queuedTwo2.ID,
+			WaitForResolution: true,
 		}
+		body, gotStatus := gdh.Exchange()
+		a.wantStatus200(t, gotStatus)
+		gotResponse := a.wantR11nResponse(t, body)
+		a.wantQueuePos(t, gotResponse, -1)
```
nyarly: n.b. about 6% of the time, this gets -2 instead.
Try `go test ./server -run afterprocessing -count 100`

It seems like it should be -2? But that it doesn't matter much?

nyarly: The issue I'm concerned with is that it'll lead to flaky tests. How should we resolve this?

util/logging/config.go
```diff
@@ -12,6 +12,7 @@ type Config struct {
 	Basic struct {
 		Level          string `env:"SOUS_LOGGING_LEVEL"`
 		DisableConsole bool
+		ExtraConsole   bool `env:"SOUS_EXTRA_CONSOLE"`
```
JC1738: so idea is just add true for dev server and anywhere else?

server/handle_single_deployment.go
JC1738: I see, it just needs extraconsole....

util/logging/kafkasink.go
```diff
@@ -67,7 +67,7 @@ func (sink *kafkaSink) ID() string {
 	return sink.id
 }
 
-func (sink kafkaSink) shouldSend(lvl Level) bool {
+func (sink *kafkaSink) shouldSend(lvl Level) bool {
```
JC1738: why make this a ptr ref?

util/logging/messages/generalmsg.go
```diff
@@ -231,7 +234,7 @@ func (l logFieldsMessage) returnIDs() (ids string, values string) {
 func (l logFieldsMessage) composeMsg() string {
 	return l.msg
 }
-func buildLogFieldsMessage(msg string, console bool, serverConsole bool, withIDs bool, loglvl logging.Level) logFieldsMessage {
+func buildLogFieldsMessage(msg string, console bool, withIDs bool, loglvl logging.Level) logFieldsMessage {
```
JC1738: surprised tests didn't need changed, thought there was a call to this from the unit tests.

server/handle_r11n_test.go
samsalisbury: It should be minus one as this was the second item in the queue, thus it finished last, and the first item is -2 since it finished first. I don't think the update of queue position is synchronised with the item being processed. I'll look into improving this test now.

server/handle_single_deployment.go
nyarly: And now the server console stuff is available for all ToConsole, but controlled by config, so users don't see it.

util/logging/config.go
nyarly: Yes

util/logging/kafkasink.go
nyarly: The kafkaSync has a WaitGroup, which the immediate receiver copies. It's obviously not an issue, but vet complains, and it's not a big deal to use a * instead

util/logging/messages/generalmsg.go
nyarly: There were :)

PR #649, opened 2018-03-20
WIP:  Sous 897 poll deploy queue

JC1738:
Not ready for merge, error after merging master running unit tests.  @samsalisbury It's 

`goroutine 2053 [chan send, 2 minutes]:
github.com/opentable/sous/vendor/github.com/samsalisbury/psyringe.(*ctor).finishWithError.func1.1(0xc42053a7e0, 0x0, 0x0)
        /home/jc/go/src/github.com/opentable/sous/vendor/github.com/samsalisbury/psyringe/ctor.go:125 +0x6d
created by github.com/opentable/sous/vendor/github.com/samsalisbury/psyringe.(*ctor).finishWithError.func1
        /home/jc/go/src/github.com/opentable/sous/vendor/github.com/samsalisbury/psyringe/ctor.go:123 +0x68`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,13 +87,21 @@ help:
 	@echo
 	@echo "Add VERBOSE=1 for tons of extra output."
 
-build-debug:
+build-debug: build-debug-linux build-debug-darwin
+
+build-debug-linux:
 	@if [[ $(SOUS_VERSION) != *"debug" ]]; then echo 'missing debug at the end of semv, please add'; exit -1; fi
 	echo "building debug version" $(SOUS_VERSION) "to" $(BIN_DIR) "with" $(CONCAT_XGO_ARGS)
-
 	mkdir -p $(BIN_DIR)
-	xgo $(CONCAT_XGO_ARGS) --targets=linux/amd64  ./
-	mv ./artifacts/bin/sous-linux-amd64 ./artifacts/bin/sous-$(SOUS_VERSION)
+	xgo $(CONCAT_XGO_ARGS) --targets=linux/amd64 ./
+	mv ./artifacts/bin/sous-linux-amd64 ./artifacts/bin/sous-linux-$(SOUS_VERSION)
+
+build-debug-darwin:
+	@if [[ $(SOUS_VERSION) != *"debug" ]]; then echo 'missing debug at the end of semv, please add'; exit -1; fi
+	echo "building debug version" $(SOUS_VERSION) "to" $(BIN_DIR) "with" $(CONCAT_XGO_ARGS)
+	mkdir -p $(BIN_DIR)
+	xgo $(CONCAT_XGO_ARGS) --targets=darwin/amd64 -out darwin ./
+	mv ./artifacts/bin/darwin* ./artifacts/bin/sous-darwin-$(SOUS_VERSION)
 
 clean:
 	rm -rf $(COVER_DIR)
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,8 @@ DESC := $(shell git describe)
 DATE := $(shell date +%Y-%m-%dT%H-%M-%S)
 DEV_VERSION := "$(DESC)-devbuild-$(DATE)"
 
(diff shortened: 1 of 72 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

JC1738:
@samsalisbury Getting closer for ready for merge.  It's working round trip, it's got a slight deficiency compared to OTPL deploy due to how they check for deploys.  If you look at the OTPL scripts, they continue to poll on /pending and grep if the deployment id is present, they wait till not present then they switch to /history/request.  This ensures that deployment is in final state, we just check history, which seems to say SUCCEEDED well before the deployment finishes.  We might (probably) will need to mimic this behaviour, might be something to look at tomorrow if looking for things to move this forward?

lib/resolver.go
```diff
@@ -49,6 +49,16 @@ func (r *Resolver) queueDiffs(dcs *DeployableChans, results chan DiffResolution)
 			logging.ReportError(r.ls, err)
 			continue
 		}
+		// SameKind == "no diffs"
+		if p.Kind() == SameKind {
+			messages.ReportLogFieldsMessageWithIDs("Not adding equal diff", logging.ExtraDebug1Level, r.ls, p)
+			continue
+		}
+		// until we have states, not going to bother with no ops
+		if p.Post.NumInstances == 0 || p.Post.DeploySpec().Version.String() == "0.0.0" {
```
samsalisbury: Setting "NumInstances: 0" is currently the only way to stop a deployment through Sous, so we can't consider that a no-op. It would be kind of a no-op if the action was "Create" but not for "Modify". I'll see if I can sensibly make that change here. Likewise, setting version to "0.0.0" could conceivably not be a no-op, although I think that case can be ignored for now.

Makefile
```diff
@@ -103,6 +103,16 @@ build-debug-darwin:
 	xgo $(CONCAT_XGO_ARGS) --targets=darwin/amd64 -out darwin ./
 	mv ./artifacts/bin/darwin* ./artifacts/bin/sous-darwin-$(SOUS_VERSION)
 
+build-debug-deploy-linux: build-debug-linux
+	rm ~/go/bin/sous
```
samsalisbury: I would call this "install-debug-linux" as it modifies things in your presumed path. Also, I use `GOPATH=$HOME` not `$HOME/go`, so this wouldn't remove any file for me usually... Line 109 below runs 'sous version' so it's looking for sous in the `$PATH`, however the sous in `$GOPATH/bin` might not be the first one found in the path, or even at all necessarily. Man these things get fiddly, but I would at a push rather `SOUS_BIN=$(which sous) && rm $SOUS_BIN && cp ./artifacts/bin/sous-darwin-$(SOUS_VERSION) $SOUS_BIN` as that will switch out the sous in your path with the new one...

samsalisbury: Note the above suggested script should also handle the case that sous is installed with some other package manager (e.g. brew.sh on mac) and probably gracefully uninstall rather than just `rm`ing the binary. See the `install-dev` recipe from Makefile (below) which uses a similar approach (instead of poking into your current path's sous, this script tries to remove 2 likely sous binaries, and then asks you to run `hash -r` at the end as this seemed cleaner at the time esp when dealing with multiple shells at the same time. It's still far from perfect, but def a little bit safer):

```Makefile
install-dev:
	brew uninstall opentable/public/sous || true
	rm "$$(which sous)" || true
	go install -ldflags "-X main.VersionString=$(DEV_VERSION)"
	echo "Now run 'hash -r && sous version' to make sure you are using the dev version of sous."
```

PR #658, opened 2018-03-26
Refactors out some duplication of code and coupling

nyarly:
The "schedulerDID" field is replaced by a UUID on the DeployablePair, for
tracking across API requests.


sous-machine-account:
LGTM

</p>

ext/singularity/requeststate.go
```diff
@@ -26,7 +27,11 @@ func (r *deployer) Status(reg sous.Registry, clusters sous.Clusters, pair *sous.
 		return nil, errors.Errorf("No cluster found for %q. Known are: %q.", clusterName, clusters.Names())
 	}
 
-	depID := pair.Post.SchedulerDID
+	if pair.UUID == uuid.Nil {
+		pair.UUID = uuid.NewV4()
+	}
+	depID := computeDeployIDFromUUID(pair.Post, pair.UUID)
```
samsalisbury: This seems more complex/more likely to go wrong than just storing a plain SingularityRequestID in the pair. I can see the benefit of having a UUID in addition to this (in fact rectifications already have a UUID, so could possibly reuse this rather than generating an additional one?).

server/handle_single_deployment.go
```diff
@@ -223,58 +218,3 @@ func (psd *PUTSingleDeploymentHandler) Exchange() (interface{}, int) {
...
-	depBase := strings.Join([]string{
-		versionTrunc,
-		uuidEntire,
-	}, "_")
-
-	if len(depBase) > maxDeployIDLen {
-		return depBase[:(maxDeployIDLen)]
-	}
-	return depBase
-}
```
samsalisbury: Nice to see this gone.

lib/deployable_pair.go
```diff
@@ -10,6 +14,8 @@ type (
 		Prior, Post  *Deployable
 		name         DeploymentID
 		ExecutorData interface{}
+		// Allows us to track a deployable pair over time and across API requests.
+		UUID uuid.UUID
```
samsalisbury: What would you think of making this an opaque string, that just happens to be a UUID in practice? That would avoid the complexity of relying on uuid.UUID's (Unm|M)arshalText methods.

PR #670, opened 2018-04-02
Postgres docker

JC1738:
This is a way to run postgres and liquibase without interfering with Make file or others.  People can specialize there taskfile.yml:  https://github.com/go-task/task


sous-machine-account:
LGTM

</p>

Taskfile.yml
```diff
@@ -0,0 +1,117 @@
+version: '2'
```
nyarly: I'm not clear on what advantages we get out of adding a different build tool to the project.

Makefile
```diff
@@ -73,9 +73,12 @@ COVER_DIR := /tmp/sous-cover
 TEST_VERBOSE := $(if $(VERBOSE),-v,)
 SOUS_PACKAGES:= $(shell go list -f '{{.ImportPath}}' ./... | grep -v 'vendor')
 SOUS_PACKAGES_WITH_TESTS:= $(shell go list -f '{{if len .TestGoFiles}}{{.ImportPath}}{{end}}' ./...)
+SOUS_TC_PACKAGES=$(shell docker run --rm -v $(PWD):/go/src/github.com/opentable/sous -w /go/src/github.com/opentable/sous golang:1.10 go list -f '{{if len .TestGoFiles}}{{.ImportPath}}{{end}}' ./... | sed 's/_\/app/github.com\/opentable\/sous/')
```
nyarly: It seems like this is adding a Go docker container to run commands in. Is that right? What advantage does that give us over running Go natively?

Taskfile.yml
nyarly: It seems like there's a `task` binary added to the repo. Is that built from this?

Makefile
JC1738: Teamcity agents don't have GO natively.  So with this, don't have to install anything on the agent.  Second advantage though not as much is easier to rev versions.  @nyarly

Taskfile.yml
JC1738: Task binary is just for the teamcity agents.  It's easy to install task locally, but the TC agents are super clean, so no go, no easy way to get it.  It's a linux bin.

PR #688, opened 2018-04-13
#trivial Remove global logger from 3 types

samsalisbury:
Beginning work on removing global logger...

Note I've removed some logs in `.Equal` methods, they need to be reinstated at call sites. It does not seem reasonable to inject a logger into these DTOs, but callers of `.Equal` should have more context so once reinstated the logs should be a lot more useful.\

**UPDATE** it has been proposed that we might not need the removed logging lines at all, I agree that these are trace-level logs that probably aren't generally useful, so I'm happy to just leave them removed.

### DIFF
--- a/lib/deploy_config.go
+++ b/lib/deploy_config.go
@@ -113,7 +113,6 @@ func (dc *DeployConfig) String() string {
 
 // Equal is used to compare DeployConfigs
 func (dc *DeployConfig) Equal(o DeployConfig) bool {
-	messages.ReportLogFieldsMessage("checking equal", logging.ExtraDebug1Level, logging.Log, dc, o)
 	diff, _ := dc.Diff(o)
 	return !diff
 }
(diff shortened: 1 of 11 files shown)


sous-machine-account:
LGTM

</p>

lib/deploy_config.go
```diff
@@ -113,7 +113,6 @@ func (dc *DeployConfig) String() string {
 
 // Equal is used to compare DeployConfigs
 func (dc *DeployConfig) Equal(o DeployConfig) bool {
-	messages.ReportLogFieldsMessage("checking equal", logging.ExtraDebug1Level, logging.Log, dc, o)
 	diff, _ := dc.Diff(o)
```
samsalisbury: I propose that it's the responsibility of one of the transitive the callers of `Equal` to log this message if it's important. Preferably somewhere up the stack where there is more context.

lib/deploy_config.go
```diff
@@ -181,19 +180,15 @@ func (dc DeployConfig) Clone() (c DeployConfig) {
...
 		return false
 	}
 
 	for name, value := range e {
 		if ov, ok := o[name]; !ok || ov != value {
-			messages.ReportLogFieldsMessage("Envs: !=", logging.ExtraDebug1Level, logging.Log, e, o, name, value, ov)
 			return false
 		}
 	}
-	messages.ReportLogFieldsMessage("Envs: == !", logging.ExtraDebug1Level, logging.Log, e, o)
```
samsalisbury: Likewise, the caller of Equal (or or its caller recursively) could log these messages with more context. Since Equal does not return this detailed info, we might want to instead call `Diff` or make this Equal method return more details.

PR #689, opened 2018-04-17
Logging contexts

nyarly:


util/logging/logset_test.go
```diff
@@ -17,10 +19,16 @@ func TestLogSetContext(t *testing.T) {
...
-		msg := sent[0].PassedArgs().Get(1)
-		assert.Equal(t, msg, "a carrot!")
-	}
+	require.Len(t, sent, 1)
+	msg := sent[0].PassedArgs().Get(1)
+	entry := msg.(*logrus.Entry)
+	data := entry.Data
+	assert.Equal(t, 100, data["logged"])
+	assert.Equal(t, 10, data["gc-value"])
+	assert.Equal(t, 20, data["override"])
```
samsalisbury: This def makes sense to (that later-added values win).

util/logging/logset_test.go
```diff
@@ -17,10 +19,16 @@ func TestLogSetContext(t *testing.T) {
...
+	assert.Equal(t, 1, data["child-value"])
+	assert.Equal(t, "test.child.grandchild", data["logger-name"])
+	assert.Equal(t, `{"message":{"array":[],"redundant":{"logged":["extra"]}}}`, data["json-value"])
```
samsalisbury: It seems slightly confusing that, reading left to right, earlier values win in the `Deliver` call. But, I can also see the point of having the most important message first in that call.. Hmm, guess there's no easy way to satisfy both things without making it more complex to call (e.g. adding a preceding `WithContext(...).Deliver(...)` or something. So, happy to keep this as-is, just need to make sure we're all aware of this.

samsalisbury: Left that comment on the wrong commit.. re-adding in the right place.

util/logging/logset_test.go
```diff
@@ -0,0 +1,34 @@
...
+	require.Len(t, sent, 1)
+	msg := sent[0].PassedArgs().Get(1)
+	entry := msg.(*logrus.Entry)
+	data := entry.Data
+	assert.Equal(t, 100, data["logged"])
+	assert.Equal(t, 10, data["gc-value"])
+	assert.Equal(t, 20, data["override"])
+	assert.Equal(t, 1, data["child-value"])
+	assert.Equal(t, "test.child.grandchild", data["logger-name"])
+	assert.Equal(t, `{"message":{"redundant":{"logged":["extra"]}}}`, data["json-value"])
```
samsalisbury: Re-adding comment in the right place:

> It seems slightly confusing that, reading left to right, earlier values win in the Deliver call. But, I can also see the point of having the most important message first in that call.. Hmm, guess there's no easy way to satisfy both things without making it more complex to call (e.g. adding a preceding WithContext(...).Deliver(...) or something. So, happy to keep this as-is, just need to make sure we're all aware of this.

PR #698, opened 2018-04-20
add kind flag

JC1738:


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

cli/sous_init.go
```diff
@@ -57,12 +57,29 @@ func (si *SousInit) AddFlags(fs *flag.FlagSet) {
...
+	kindOk := false
+
+	switch kind {
+	case sous.ManifestKindService:
+		kindOk = true
+	case sous.ManifestKindScheduled:
+		kindOk = true
+	default:
+		kindOk = false
+	}
```
nyarly: It seems like we should at least include on-demand as well. Maybe worker? I think once doesn't make sense for Sous (iow, use on-demand instead)

JC1738: @nyarly I wasn't sure which ones are actually supported by Sous, I thought it was just http-service and scheduled, the other constants exist, but I figured they were invalid?

cli/sous_init.go
```diff
@@ -67,11 +67,18 @@ func (si *SousInit) Execute(args []string) cmdr.Result {
...
 
+	m := si.Target.Manifest
+
 	switch kind {
 	case sous.ManifestKindService:
 		kindOk = true
 	case sous.ManifestKindScheduled:
 		kindOk = true
+		for _, v := range m.Deployments {
+			v.DeployConfig.Startup.SkipCheck = true
```
JC1738: This doesn't seem to work, no errors, but looking at whats in GDM afterwards, doesn't reflect SkipCheck = true

JC1738: map_state_to_deployments.go seems to doing stuff with Startup and Unmerged Defaults, could be overrighting here?

cli/sous_init.go
```diff
@@ -74,9 +74,18 @@ func (si *SousInit) Execute(args []string) cmdr.Result {
 		kindOk = true
 	case sous.ManifestKindScheduled:
 		kindOk = true
-		for _, v := range m.Deployments {
-			v.DeployConfig.Startup.SkipCheck = true
+
+		var dsm map[string]sous.DeploySpec
+		dsm = make(map[string]sous.DeploySpec)
+
+		for k := range m.Deployments {
```
JC1738: this is done this way since working with values of map isn't necessarily allowed since they represent non-addressable locations and compiler will fail. 

See:
https://github.com/golang/go/issues/11865
https://stackoverflow.com/questions/32751537/why-do-i-get-a-cannot-assign-error-when-setting-value-t...

config/deploy_filter.go
```diff
@@ -19,6 +19,7 @@ type DeployFilterFlags struct {
 	Tag      string
 	Revision string
 	Cluster  string
+	Kind     string
```
nyarly: I don't think Kind makes sense as a field here. The `DeployFilterFlags` struct exists to generate a `ResolveFilter` (c.f. `BuildFilter`) which gets used to filter the GDM and otherwise refer to deployments and manifests by their IDs. `Kind` is a configuration field though. It probably makes more sense as a field on the `SousBuild` command struct itself.

PR #701, opened 2018-04-24
Trace IDs

nyarly:
OT-RequestId is added by client-side http clients, and logged on the server side.

N.b. there's no explicit tests for this behavior - it would require an e2e test to do properly.

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/samsalisbury/psyringe"
 	"github.com/samsalisbury/semv"
+	uuid "github.com/satori/go.uuid"
 )
 
 type (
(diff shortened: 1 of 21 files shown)


sous-machine-account:
# LGTM!

</p>

graph/graph.go
```diff
@@ -165,6 +166,7 @@ func BuildBaseGraph(version semv.Version, in io.Reader, out, err io.Writer) *Sou
 	graph := newSousGraph()
 	graph.Add(
 		version,
+		sous.TraceID(uuid.NewV4().String()),
```
samsalisbury: This made me think... is it right to assign a single TraceID for the lifetime of the process? In client (CLI) mode it makes sense for sure, and I can see it being overridden in child loggers created in the HTTP handlers. Having it here also means 'sous server' invocations have a TraceID which also seems like correct behaviour. So, I concur that this is the right place to put this. See note below about whether we should generate new TID in the case that the client doesn't send one...

util/restful/router.go
```diff
@@ -218,19 +226,22 @@ func (mh *MetaHandler) InstallPanicHandler() {
...
 		Exchanger: h,
 		Request:   r,
 		Params:    p,
 	}
 }
 
-func (mh *MetaHandler) injectedHandler(factory ExchangeFactory, w http.ResponseWriter, r *http.Request, p httprouter.Params) Exchanger {
-	h := factory(mh.routeMap, w, r, p)
+func (mh *MetaHandler) injectedHandler(factory ExchangeFactory, resName string, w http.ResponseWriter, r *http.Request, p httprouter.Params) Exchanger {
+	tid := r.Header.Get("OT-RequestId")
```
samsalisbury: In the case that it's an old client, or a new client erroneously not setting the header, should we not generate a new TraceID here to at least group together subsequent log entries? (I.e. tid could be an empty string here which seems like a wasted opportunity). From memory, I believe that is the behaviour of other users of Ot-RequestID.

PR #706, opened 2018-04-25
doc: tc-pipelines.md #trivial

samsalisbury:
Beginning documenting aspirations for auto TC pipelines.


sous-machine-account:
# LGTM!

</p>

doc/tc-pipelines.md
```diff
@@ -0,0 +1,97 @@
...
+  - ci-sf.AutoDeploy.Git.Branch: master
+  - pp-sf.AutoDeploy.IfSmokeTestsPassedIn: ci-sf
+  - prod-sc.AutoDeploy.Git.Tag.Semver.NoPrerelease: true
+  - prod-ln.AutoDeploy.Disabled: true
+
+By default, unit and integration tests must pass before any deployment is allowed.
+
+## What might this look like to use?
+
+New command: 'sous cd'
```
nyarly: I'm :-1: on this. Much prefer a separate `sous-teamcity` tool to manage building pipelines. It's a very different process and outside of Sous's own scope.

doc/tc-pipelines.md
```diff
@@ -0,0 +1,97 @@
...
+'sous build' && 'sous deploy -cluster X'.times(clusters)
+However users always need more than this:
+
+- Run self-contained unit and integration tests
+- Run smoke tests against deployed applications
+- Control which envs are deployed to when (e.g. deploy all to CI, passing smoke tests to PP, and manually to prod).
+
+These 3 features would require more information than we have right now.
+
+### Maybe we need to add some more properties to the manifest...
```
nyarly: Really hesitant to add more to the manifest. It takes what's a descriptive artifact and starts making it into a program. Maybe one day we can completely replace TC, but for now I'd vote for adding "stub" build steps for unit tests, smoke tests, etc.

JC1738: other option is that multiple containers, some of the containers that are generated are the the tests and the execute of the conatiner runs tests that pass or fail.  I guess though that still doesn't solve what the execute step is, unless the users provide the Dockerfile for the containers?

doc/tc-pipelines.md
samsalisbury: There are benefits of having this part of the 'sous' command, namely discoverability. However, that doesn't mean we need to keep the tool proper in the same repo. 'sous cd' could just be a configurable external tool, and maybe sous server could even tell the client which tool and how to configure it.

doc/tc-pipelines.md
samsalisbury: Yes, we can't really leave any steps blank, as that would be too much work and would stop people upgrading their pipelines when improvements are made. Really adding these commands is closer to replacing the Dockerfile than to replacing TeamCity. One compromise could be to add this stuff as metadata for now, and have the configured external 'sous cd' command check it for validity etc.

doc/tc-pipelines.md
nyarly: I still disagree. This is akin to building `sous auto-sensu` or whatever. I think it's a big scope creep of Sous, but it's an *excellent* example of tools and services peripheral to Sous.

PR #712, opened 2018-04-30
Scoop the traceID to display RequestID after action run

JC1738:
Works now, problem was exported TraceID vs traceID


JC1738:
image

Not sure I trust the code coverage, this is what is saying is different in coverage from Master...

PR #716, opened 2018-05-04
Remove old postgres and add new docker posgres/liquibase #trivial

JC1738:


JC1738:
This works on my machine (famous words have been uttered throughout the years) @nyarly when you get a chance, can you?

1.  Stop your Postgres instance
2.  Pull this branch
3.  Run make test-unit

And see what you think.

sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.travis.yml
```diff
@@ -1,35 +1,16 @@
 sudo: required
 services:
 - docker
-- postgresql
```
JC1738: Going to let the makefile handle the postgres and liquibase, nothing needed to install in base travis builder.

JC1738:
@samsalisbury Cool, add a change to makefile

image

Makefile
```diff
@@ -371,33 +378,20 @@ artifacts/sous_$(SOUS_VERSION)_amd64.deb: artifacts/$(LINUX_RELEASE_DIR)/sous
...
-postgres-test-prepare: $(DEV_POSTGRES_DATA_DIR)/postgresql.conf postgres-create-testdb
-
-postgres-create-testdb: postgres-start
-	createdb -h localhost -p $(PGPORT) $(TEST_DB_NAME) > /dev/null 2>&1 || true
-	liquibase $(LIQUIBASE_TEST_FLAGS) update
+postgres-start:
+	docker stop postgres > /dev/null 2>&1 || true
+	rm -rf $(DEV_POSTGRES_DATA_DIR)
+	install -d -m 0700 $(DEV_POSTGRES_DATA_DIR)
+	docker run -d --name postgres -p $(PGPORT):5432 --rm --user $(USER_ID):$(GROUP_ID) -v /etc/passwd:/etc/passwd:ro -v $(DEV_POSTGRES_DATA_DIR):/var/lib/postgresql/data postgres:10.3
```
samsalisbury: Mounting /etc/passwd from my machine scares me for some reason, even read-only. Also it doesn't seem to have the desired effect (my user ID is not registered on the machine), I believe that's because it's mounting from the Docker Machine rather than from my Mac... There is a way to have docker-machine mount a host dir into itself for mounting into containers, but I would rather avoid docker-machine specific code.

I'm curious if it's necessary to use specific real user IDs from the host machine at all? I get that this is necessary for files on the volume to be accessible by the developer on their local machine, however is that itself necessary? Could we just use a named volume for the Postgres database instead (i.e. one managed automatically by Docker).

I've just taken a stab at using Docker-managed volume, and at least `make postgres-start` is working now...

samsalisbury: See 7abe2eaa31e9c929b9c2c53b00689af1d134224d

PR #729, opened 2018-05-25
Slack v2

JC1738:
Additional configuration to make slack handling more configurable.

### DIFF
--- a/cli/sous_deploy.go
+++ b/cli/sous_deploy.go
@@ -71,14 +71,25 @@ func (sd *SousDeploy) Execute(args []string) cmdr.Result {
 func (sd *SousDeploy) slackMessage(action actions.Action, err error) {
 
 	var slackURL, slackChannel string
+	var additionalChannels map[string]string
+
 	d, ok := action.(*actions.Deploy)
 
 	if ok {
 		slackURL = d.Config.SlackHookURL
 		slackChannel = d.Config.SlackChannel
+		additionalChannels = d.Config.AdditionalSlackChannels
 	}
 
-	if len(slackURL) < 1 || len(slackChannel) < 1 {
+	if additionalChannels == nil {
+		additionalChannels = make(map[string]string)
+	}
+
+	if len(slackURL) > 0 && len(slackChannel) > 0 {
+		additionalChannels[slackChannel] = slackURL
+	}
+
+	if len(additionalChannels) < 1 {
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


sous-machine-account:
# LGTM!

</p>

cli/sous_deploy.go
```diff
@@ -71,14 +71,25 @@ func (sd *SousDeploy) Execute(args []string) cmdr.Result {
 func (sd *SousDeploy) slackMessage(action actions.Action, err error) {
```
samsalisbury: I know it's not part of this PR, missed this originally.. I wonder if this should just take a `config.Config`, and a `sous.Deployment` (or even `sous.DeploySpec`?) and the error. Would just seem slightly simpler than casting the action, and pulling out its fields. Not blocking over this, just struck me as simpler.

config/config.go
```diff
@@ -51,6 +51,8 @@ type (
 		SlackHookURL string `env:"SOUS_SLACK_HOOK_URL"`
 		// SlackChannel that should receive messages
 		SlackChannel string `env:"SOUS_SLACK_CHANNEL"`
+		// AdditionalSlackChannels that should receive messages
+		AdditionalSlackChannels map[string]string `env:"SOUS_ADDITIONAL_SLACK_CHANNELS"`
```
samsalisbury: Curious why this is a separate value rather than just changing `SlackChannel` to a map or slice?

cli/sous_deploy.go
samsalisbury: Could at least accept an `*actions.Deploy` and force the cast (actually type assertion) to happen outside this func?

cli/sous_deploy.go
```diff
@@ -71,14 +71,25 @@ func (sd *SousDeploy) Execute(args []string) cmdr.Result {
...
 	var slackURL, slackChannel string
+	var additionalChannels map[string]string
+
 	d, ok := action.(*actions.Deploy)
 
 	if ok {
 		slackURL = d.Config.SlackHookURL
 		slackChannel = d.Config.SlackChannel
+		additionalChannels = d.Config.AdditionalSlackChannels
+	}
```
samsalisbury: Could we return early in case this cast fails? Or is there still useful behaviour once we know it's not an `*actions.Deploy`?

cli/sous_deploy_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestSlackMessage_just_hookandchannel(t *testing.T) {
+	sd := &SousDeploy{}
+	config := &config.Config{
+		SlackHookURL: "http://foo.com",
+		SlackChannel: "bar",
+	}
+
+	dAction := returnAction(t, config)
+	assert.NotPanics(t, func() { sd.slackMessage(dAction, nil) }, "shouldn't panic")
```
samsalisbury: I wonder if there are any deeper assertions than not panicking for these tests? Could we test that the message has the right fields set? Maybe if we separate constructing the slack payload from sending it, the payload itself could be tested in isolation?

config/config.go
JC1738: Yes, was a feature add, it would require a bit of work to get to slice since I don't think we do environment variable to slice (where we do for struct) wouldn't be impossible.

cli/sous_deploy_test.go
JC1738: Yes, though the bare minimum needed since this feature is in no way block deploy is that the function does panic.  But to actually test it delivers, would require testing payload, but then again we don't own the slack part, so all really confirming is slack url is set, so really not testing anything important.

PR #731, opened 2018-05-31
SOUS-919: Primary store toggle

nyarly:
Adds a configuration option `DatabasePrimary` (overridden with
`SOUS_DATABASE_IS_PRIMARY`) which selects which of the datastores to "prefer."
The other is merely updated, but not considered for the current state. This is
the current behavior, with Git as primary and Postgres as secondary.

When the DatabasePrimary config is set, Postgres is considered the sole source
of truth.  This should **not** yet be used in deployment - running smoke tests
with this enabled makes it clear that the "empty" database isn't complete
enough (the tests fail).

It's possible that the has-been-run-as-secondary database is ready to be used
as primary, but I'd like to have a from-scratch solution ready before we try to
roll that out.  We could try it, I suppose in dev and see what happens...

Regardless, this work is ready to merge, and _can_ be deployed. It just
shouldn't be toggled on yet.

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -624,10 +624,15 @@ func newInMemoryClient(srvr ServerHandler, log LogSink) (HTTPClient, error) {
 }
 
 func newServerStateManager(c LocalSousConfig, log LogSink, gm gitStateManager, dm distStateManager) (*ServerStateManager, error) {
-	primary := gm
-	secondary := dm
+	var primary, secondary sous.StateManager
+	primary = gm
+	secondary = dm
 
-	duplex := storage.NewDuplexStateManager(primary.StateManager, secondary.StateManager, log.Child("duplex-state"))
+	if c.DatabasePrimary {
+		primary, secondary = secondary, primary
+	}
+
+	duplex := storage.NewDuplexStateManager(primary, secondary, log.Child("duplex-state"))
 	return &ServerStateManager{StateManager: duplex}, nil
 }
 
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -630,6 +630,19 @@ func newServerStateManager(c LocalSousConfig, log LogSink, gm gitStateManager, d
 
 	if c.DatabasePrimary {
 		primary, secondary = secondary, primary
+		logging.Deliver(log,
+			logging.InformationLevel,
+			logging.SousGenericV1,
+			logging.GetCallerInfo(),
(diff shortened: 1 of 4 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

graph/graph.go
```diff
@@ -621,36 +623,53 @@ func newInMemoryClient(srvr ServerHandler, log LogSink) (HTTPClient, error) {
...
 	err := mdb.Err
 	if err == nil {
-		secondary, err = newDistributedStorage(mdb.Db, c, tid, rf, log)
+		dist, err = newDistributedStorage(mdb.Db, c, tid, rf, log)
 	}
 
 	if err != nil { // because DB Err wasn't nil, or distributed didn't set up well
 		logging.ReportError(log, errors.Wrapf(err, "connecting to database with %#v", c.Database))
-		secondary = storage.NewLogOnlyStateManager(log.Child("database"))
+		dist = storage.NewLogOnlyStateManager(log.Child("database"))
```
JC1738: If error setting up DB, would the NewLogOnlyStateManager still have a log.Child("database") or wouldn't that be git?

nyarly: Oh, that's an important point. Let me think about how to address it.

graph/graph.go
```diff
@@ -621,36 +623,53 @@ func newInMemoryClient(srvr ServerHandler, log LogSink) (HTTPClient, error) {
...
 
-func newServerStateManager(c LocalSousConfig, mdb MaybeDatabase, tid sous.TraceID, rf *sous.ResolveFilter, log LogSink, gmf gitStateManagerFactory) (*ServerStateManager, error) {
-	var secondary sous.StateManager
+func newServerStateManager(c LocalSousConfig, log LogSink, gm gitStateManager, dm distStateManager) (*ServerStateManager, error) {
+	var primary, secondary sous.StateManager
+	primary = gm
+	secondary = dm
+
+	if c.DatabasePrimary {
+		primary, secondary = secondary, primary
```
lachlanh: neat formulation but seems overly terse
primary = dm
secondary = gm 
would be my choice for readability.

PR #733, opened 2018-06-04
Lowercase

JC1738:
Was able to build a mixed case that was giving me error before, but not sure if need to change anywhere else?


sous-machine-account:
# LGTM!

</p>

lachlanh:
test ?

JC1738:
@samsalisbury Not sure what's going on here, they all pass continually no matter what I do on TeamCity and Fail on Travis, doesn't make sense?

ext/docker/builder.go
```diff
@@ -95,7 +96,7 @@ func (b *Builder) applyMetadata(bp *sous.BuildProduct) error {
 	bp.VersionName = b.VersionTag(bp.Source, bp.Kind)
 	bp.RevisionName = b.RevisionTag(bp.Source, bp.RevisionName, bp.Kind, time.Now())
 
-	c := b.SourceShell.Cmd("docker", "build", "-t", bp.VersionName, "-t", bp.RevisionName, "-")
+	c := b.SourceShell.Cmd("docker", "build", "-t", strings.ToLower(bp.VersionName), "-t", strings.ToLower(bp.RevisionName), "-")
```
samsalisbury: I wonder if this is too late to force lowercase.. Seems that as soon as we know it's a docker image, we should be producing compatible all-lowercase identifiers. This means we can just pass buildproducts around and know they are consistent everywhere, in logs etc, rather than patching up here.

Right now we manually construct a `sous.BuildProduct` in numerous places, where we manually set its `ID` field to that of a Docker image. I think a more solid approach could be to make a constructor for `sous.BuildProduct` e.g. `NewDockerBuildProduct(...)` that takes raw input and produces valid Docker image identifiers internally e.g. looking at current usage, it could take parameters `(sous.SourceID, "runnable|builder", baseAdvisories []Advisory, extraAdvisories ...Advisory)`. This itself could be easily tested.

JC1738:
Closing this for now, and putting issue on blocked, don't think this is the right solution/fails in travis.

PR #734, opened 2018-06-05
Deployuser

JC1738:


samsalisbury:
I think it would be worth adding a smoke test assertion about the username being set correctly. See e.g. `assertNilHealthCheckOnLatestDeploy` in `test/smoke/testsingularityassertions_test.go` which could be a good template to use for this (it has access to the latest deployment/deploymarker with the user in it).

ext/singularity/requeststate.go
```diff
@@ -43,7 +43,7 @@ func (r *deployer) Status(reg sous.Registry, clusters sous.Clusters, pair *sous.
 
 	tgt, err := BuildDeployment(reg, clusters, singReq, r.log)
 
-	tgt.SchedulerURL = fmt.Sprintf("http://%s/request/%s", url, reqID)
```
JC1738: @nyarly I have no idea why mine has this with http:// already, I not sure which one is right?

test/smoke/testsingularity_utils_test.go
```diff
@@ -92,7 +92,7 @@ func (s *Singularity) GetRequestForDeployment(t *testing.T, did sous.DeploymentI
 func (s *Singularity) GetLatestDeployForDeployment(t *testing.T, did sous.DeploymentID) *dtos.SingularityDeployHistory {
 	t.Helper()
 	reqID := s.mustGetReqID(t, did)
-	deps, err := s.client.GetDeploys(reqID, 100, 0)
+	deps, err := s.client.GetDeploys(reqID, 100, 1)
```
JC1738: Note, now we are actually passing arguements, 0 was an invalid page value @samsalisbury @nyarly

JC1738: Also means where we were passing webcache == false in a couple of places was never actually being honored, so now it will, which I guess is a good thing??

ext/singularity/deployer.go
```diff
@@ -110,7 +113,7 @@ func (r *deployer) Rectify(pair *sous.DeployablePair) sous.DiffResolution {
 		messages.ReportLogFieldsMessage("SameKind", logging.InformationLevel, r.log, postID, version, resolution)
 		return resolution
 	case sous.AddedKind:
-		messages.ReportLogFieldsMessageToConsole(fmt.Sprintf("Starting an AddedKind %s:%s", postID, version), logging.ExtraDebug1Level, r.log, pair)
+		messages.ReportLogFieldsMessageToConsole(fmt.Sprintf("Starting an AddedKind %s:%s by user %s", postID, version, user), logging.ExtraDebug1Level, r.log, pair)
```
nyarly: Ought to just make User an EachFielder and drop it in the args.

ext/singularity/requeststate.go
nyarly: I think it *should* have the scheme already. The old way is really strange.

test/smoke/newdeploy_test.go
```diff
@@ -89,7 +89,8 @@ func TestSousDeploy(t *testing.T) {
 
 					assertActiveStatus(t, f, did)
 					assertSingularityRequestTypeService(t, f, did)
-					assertNonNilHealthCheckOnLatestDeploy(t, f, did)
+          assertNonNilHealthCheckOnLatestDeploy(t, f, did)
```
nyarly: gofmt?

test/smoke/testsingularity_utils_test.go
nyarly: I guess we'll see :)

PR #736, opened 2018-06-06
Allow editing of Singularity Request ID

samsalisbury:
**EDITED**

Adds a new field to deployments in the manifest: `SingularityRequestID` if not empty, this causes all subsequent deployments from any source (rectifier or direct) to be deployed to a Singularity request with that ID. If blank, uses the globally unique auto generated ID already in use. The uniqueness of `SingularityRequestID` is per _singularity_ cluster (which may be host to many _sous_ clusters). Rather than requiring up-front definitions of singularity clusters, that table is populated during manifest edits.

TODO:

- [x] Decide if we want to block edits to already-deployed tasks, or
- [x] provide a proper migration path that cleans up the old request from Singularity.

This PR results in old Singularity requests being deleted when a rename of SingularityRequestID is performed.

Caveats:

Whilst it would seem beneficial to check if there is another singularity request with the same ID before deploying to it, this would cause problems with our proposed take-over of existing Singularity requests, since they will all be preexistent. We could make that a rule e.g. on the first ever deployment of a thing which would mitigate that issue.


sous-machine-account:
# LGTM!

</p>

lib/deploy_config.go
```diff
@@ -147,33 +148,42 @@ func (dc *DeployConfig) Diff(o DeployConfig) (bool, []string) {
...
 	}
-	c.Volumes = dc.Volumes.Clone()
-	c.Startup = dc.Startup
-	c.Schedule = dc.Schedule
+	return metadata
+}
 
-	return
+// Clone returns a deep copy of this DeployConfig.
+func (dc DeployConfig) Clone() DeployConfig {
```
nyarly: Nice

lib/deployment_diff.go
```diff
@@ -48,10 +48,7 @@ func (d DeployStates) Diff(other Deployments) *DeployableChans {
 // Diff computes the differences between two sets of Deployments
 func (d Deployments) Diff(other Deployments) *DeployableChans {
 	difr := newDiffer(d)
-	go func(d *differ, o Deployments) {
-		d.diff(o)
-	}(difr, other)
-
+	go difr.diff(other)
```
nyarly: :+1:

nyarly:
Regarding changing the request ID, we talked about making "manifest edit" be the official way to do that, and push more common changes to the deploy command. I think that might be an okay way to make this work and not have to invest in "hiding" the output.

cli/sous_deploy.go
```diff
@@ -16,10 +15,7 @@ import (
 type SousDeploy struct {
 	SousGraph *graph.SousGraph
 
-	DeployFilterFlags config.DeployFilterFlags `inject:"optional"`
-	waitStable        bool
-	force             bool
-	dryrunOption      string
+	opts graph.DeployActionOpts
```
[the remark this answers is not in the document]
samsalisbury: Should be straightforwards to move it to cli.

database/singularity-request-id.xml
```diff
@@ -0,0 +1,26 @@
...
+	</createTable>
+	<addForeignKeyConstraint baseColumnNames="component_id"
+		baseTableName="singularity_deployment_bindings"
+		constraintName="singularity_deployment_bindings_component_id_fkey"
+	    referencedColumnNames="component_id" referencedTableName="components" />
+	<addForeignKeyConstraint baseColumnNames="cluster_id"
+		baseTableName="singularity_deployment_bindings"
+		constraintName="singularity_deployment_bindings_cluster_id_fkey"
+	    referencedColumnNames="cluster_id" referencedTableName="clusters" />
+	<addPrimaryKey columnNames="component_id, cluster_id"
```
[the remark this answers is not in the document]
samsalisbury: Are there any downsides to using a multi-column PK? My gut instinct is always to use naturally occurring PKs when they exist, as here, as that makes a strong assertion about the structure of the data and avoids adding additional needless IDs. I know many people use synthetic PKs always, but I've never heard a good reason why... Not against making this change if it simplifies our code or is an improvement in some way

ext/storage/postgresstatemanager_test.go
```diff
@@ -77,7 +77,7 @@ func TestPostgresStateManagerWriteState_success(t *testing.T) {
 	}
 	suite.Equal(int64(4), suite.pluckSQL("select count(*) from deployments"))
 
-	assert.Len(t, suite.logs.CallsTo("Fields"), 13)
+	assert.Len(t, suite.logs.CallsTo("Fields"), 14)
```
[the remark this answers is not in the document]
samsalisbury: Hmm, maybe yeah. I wonder if there is a different assertion altogether that gives us what we need? I'll see what I can do..

ext/storage/postgresstatemanager_test.go
```diff
@@ -117,13 +117,21 @@ func TestPostgresStateManagerWriteState_success(t *testing.T) {
 	suite.require.NoError(err)
 
 	for diff := range oldD.Diff(newD).Pairs {
+		different, differences := diff.Post.Deployment.Diff(diff.Prior.Deployment)
```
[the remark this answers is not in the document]
samsalisbury: Right, there was a problem with DeployConfig.Diff which ultimately led to this code being added. I fixed DeployConfig.Diff, and added a test for it specifically where there was none before, so we can probably remove this code...

ext/storage/postgresstatemanager_write.go
```diff
@@ -238,14 +249,21 @@ func depID(row sqlgen.RowDef, dep *sous.Deployment) {
 }
 
 func compID(row sqlgen.RowDef, dep *sous.Deployment) {
+	compIDFunc(row.FD, dep)
+}
+
+func compIDFunc(f sqlgen.FieldDefFunc, dep *sous.Deployment) {
```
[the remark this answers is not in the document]
samsalisbury: Yes, merging them into one makes sense, but it's no longer needed since we don't need upsert. I'll remove the new funcs for now as doing that refactor would touch a lot of unrelated code.

cli/sous_deploy.go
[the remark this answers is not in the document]
samsalisbury: Oh, Graph cannot import CLI, and deploy is already using the action pattern, I think this was already the right way around so leaving as-is.

samsalisbury:
Right, there's no code here to try to hide the output, it's just a regular manifest field.

cli/sous_deploy.go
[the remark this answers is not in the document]
nyarly: My point is that importing `graph` is, in general, a big hassle because it means that `graph` can't import some packages. Part of what the `graph` package let's us avoid is import cycles, but only if it's never imported elsewhere. The places where `graph` got imported are antipatterns.

The whole question can be avoided just by passing the fields explicitly to GetDeploy.

database/singularity-request-id.xml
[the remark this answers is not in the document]
nyarly: Artificial PKs (i.e. auto-increment) are useful because they do decouple the identity of the record from its structure. What you're getting by using natural PKs is a unique constraint - which we'd certainly want. But by making the unique constraint explicit and separate from the PK, you gain the flexibility to change e.g. what's in the uniqueness set (i.e. if we add fields in the future).

I think there are/were query optimization reasons around always using a single PK and always being integers, but those reasons get stale over time and there's a certain amount of superstition that develops. In the end, I land on the decoupling over everything else.

database/singularity-request-id.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+  <changeSet author="sam" id="8">
+	<createTable tableName="singularity_deployment_bindings">
+	  <column name="singularity_deployment_bindings_id" type="SERIAL" autoIncrement="true">
+	    <constraints primaryKey="true" />
+	  </column>
+	  <column name="component_id" type="INT">
+	    <constraints foreignKeyName="component_id" />
```
nyarly: As we discussed this morning: because there's a 1:many relationship to deployments, they should have a FK into this table.

The FKs to `clusters` and `components` should live in *either* `deployments` or `singularity_deployment_bindings` - there's a normalization argument that it should be here, but there's also a data migration problem.

I would suggest, then, that this table only have the `request_identifier` and its own PK for the time being, and at some future date we can concern ourselves with migrating to the "more correct" schema.

nyarly: Parenthetically, for review, my correctness argument runs that:
* there can be deployments for every cluster/component pair
* there can be requests for every cluster/component pair
* there must be a request for every deployment, and there can be 0+ deployment per request.
* any piece of information should exist exactly once in the DB.

With those constraints, it makes sense that either table have the cluster/component FKs, but not both. Because of the 0+ relationship of request/deployments, there has to be a FK from deployments to deployment_bindings. If there's no deployment for a deployment binding, what cluster is it in? Therefore the FKs belong on that table.

The migration plan looks like:

Create the columns on the deployment binding table, and augment the DB update code to update both columns.

On a subsequent deploy, include a boot-time script to copy all the values from deployments to deployment bindings, and just use the deployment binding keys.

On a final deploy, drop the deployments table FKs.

ext/singularity/deployer.go
```diff
@@ -236,12 +242,27 @@ func (r *deployer) RectifySingleModification(pair *sous.DeployablePair) (err err
...
 	if changesReq(pair) {
 		reportDeployerMessage("Updating request", pair, diffs, data, nil, logging.DebugLevel, r.log)
-		if err := r.Client.PostRequest(*pair.Post, reqID); err != nil {
+		if err := r.Client.PostRequest(*pair.Post, desiredReqID); err != nil {
+			return err
+		}
+	} else if desiredReqID != currentReqID {
+		// NOTE: This message is WarningLevel whilst the feature is new.
+		// TODO SS: Turn this down to debug level once we're happy with it.
+		m := fmt.Sprintf("Creating request %q to replace %q", currentReqID, desiredReqID)
```
lachlanh: is the wording flipped around here ? seems like it is creating desiredReqID ?

lib/deploy_config.go
```diff
@@ -147,33 +148,42 @@ func (dc *DeployConfig) Diff(o DeployConfig) (bool, []string) {
...
 	}
-	c.Resources = make(Resources)
-	for k, v := range dc.Resources {
-		c.Resources[k] = v
+	return env
+}
+
+// Clone returns an independent copy of m.
+func (e Metadata) Clone() Metadata {
+	if e == nil {
```
lachlanh: comment dosen't match signature

server/handle_manifest.go
```diff
@@ -116,6 +118,8 @@ func (pmh *PUTManifestHandler) Exchange() (interface{}, int) {
 	dec := json.NewDecoder(pmh.Request.Body)
 	dec.Decode(m)
 
+	fmt.Fprintf(os.Stdout, "PUT Manifest received =================>\n\n\n%# +v\n\n\n<==================", m)
```
lachlanh: cleanup?

lachlanh:
can't comment inline, are you sure you want to checkin logs ? test/smoke/broken-smoketest-mainserver.log

ext/singularity/deployer.go
```diff
@@ -193,10 +193,16 @@ func (r *deployer) RectifySingleCreate(d *sous.DeployablePair) (err error) {
...
+
+	reqID := d.Post.Deployment.DeployConfig.SingularityRequestID
+	if reqID == "" {
+		reqID, err = computeRequestID(d.Post)
+		if err != nil {
+			return err
+		}
 	}
+
+	// TODO SS: Check this reques ID does not already exist.
```
lachlanh: is this needed ?

ext/singularity/deployer.go
samsalisbury: Good catch

lib/deploy_config.go
samsalisbury: True, I’ll fix it

server/handle_manifest.go
samsalisbury: Ah missed one, yep this needs removing, tx

ext/singularity/deployer.go
samsalisbury: No, comment should be removed!

samsalisbury:
Re the log file, I thought this PR removed it, but doing this on my phone so hard to check.. since the bug it recorded was fixed in an earlier PR, it should be removed.

samsalisbury:
OK to merge this tomorrow once those issues mentioned are resolved?

lib/deploy_config.go
samsalisbury: Changed the receiver name to `m` everywhere rather than change comment.

PR #738, opened 2018-06-07
Init flags messages/fail for -kind on-demand

samsalisbury:
The change to the error message should be uncontroversial. However, this change explicitly breaks any attempt to `sous init -kind on-demand`. AFAIK (looking at current usage at OpenTable) we are not supporting on-demand jobs, so best to let people know as early as possible.


sous-machine-account:
# LGTM!

</p>

JC1738:
Merging ALL TC and the travis push passed

PR #742, opened 2018-06-08
Smoke test bits #trivial

samsalisbury:
- Run DB and GIT based tests in parallel rto save time.
- Output was "12 out of 10 tests passed", reversed those figures...
- Renamed file to remove trace of 'newdeploy' command.


sous-machine-account:
# LGTM!

</p>

samsalisbury:
The TeamCity failure was caused by it trying to run the old `make test-smoke-nofail` I've just altered the config for that to `make test-smoke`

samsalisbury:
The CodeCov report is bogus, so this is mergeable as-is @opentable/sous

samsalisbury:
@nyarly I see this line in the Makefile https://github.com/opentable/sous/blob/master/Makefile#L453 setting limit to 1000... I think if this is an issue we need some way to limit connections in the code itself, on a per-server basis, so we can make sure the tests don't fail because of this, and also for use in production. Is there any mechanism like this already in place, or would that be a totally new feature?

PR #743, opened 2018-06-11
Sous 955 runmount strategy

lachlanh:
This has a matching change in sous-build-containers, the runmount docker image is pushed so runmount will be available as soon as we deploy this version


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 670 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

ext/docker/runmount_buildpack_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func getContext() *sous.BuildContext {
+	sh, _ := shell.Default()
+	ctx := &sous.BuildContext{
+		Sh: sh,
+	}
+	return ctx
+}
+
+func TestRunmountBuildpack_DetectRunmount(t *testing.T) {
```
samsalisbury: Thanks for using this test naming style, super easy to find & run the tests you want like this. One improvement could be to add `_runmount` instead of `Runmount` to keep the "test case label" it separate from the method name under test.

ext/docker/runmount_buildpack_test.go
```diff
@@ -0,0 +1,54 @@
...
+	createDockerfile("FROM docker.otenv.com/sous-otj-autobuild-runmount:local")
+	defer removeDockerfile()
+
+	rmbp := NewRunmountBuildpack()
+
+	result, _ := rmbp.Detect(getContext())
```
samsalisbury: I would say it's essential to assert a nil error here as well, as most Go control flow is dictated by the nilness or not of errors...

ext/docker/runmount_buildpack_test.go
```diff
@@ -0,0 +1,54 @@
...
+	assert.True(t, result.Compatible)
+}
+
+func TestRunmountBuildpack_DetectNotRunmount(t *testing.T) {
+	createDockerfile("FROM docker.otenv.com/sous-otj-autobuild:local")
+	defer removeDockerfile()
+
+	rmbp := NewRunmountBuildpack()
+
+	result, _ := rmbp.Detect(getContext())
```
samsalisbury: Same here, that second (error) return value is a crucial part of the Detect method's contract.

ext/docker/selector.go
```diff
@@ -22,6 +22,13 @@ func NewBuildStrategySelector(ls logging.LogSink, rc docker_registry.Client) sou
 
 // SelectBuildpack tries to select a buildpack for this BuildContext.
 func (s *selector) SelectBuildpack(ctx *sous.BuildContext) (sous.Buildpack, error) {
+	rmbp := NewRunmountBuildpack()
+	dr, err := rmbp.Detect(ctx)
+	if err == nil && dr.Compatible {
+		reportStrategyChoice("runmount container", s.log)
+		return rmbp, nil
+	}
```
samsalisbury: 👍

ext/docker/runmount_buildpack.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, fmt.Errorf("%s does not exist", dfPath)
+	}
+
+	messages.ReportLogFieldsMessage("Runmount dockerfile detection", logging.DebugLevel, logging.Log, dfPath)
+
+	dockerContent, _ := readDockerfile()
+
+	// TODO LH simplest check so far, scan docker content for runmount
+	result := sous.DetectResult{
+		Compatible: strings.Contains(dockerContent, "runmount"),
```
samsalisbury: I wonder if we should constrain this to the `FROM` line at least? A collision is unlikely, but that seems straightforward enough to add. Ultimately, some kind of capabilities-based detection would seem a better solution, but I think this simple kind of check is preferable for the time being, as the role of detection will soon be diminished in favour of explicit selection of build strategies, so the extra investment is not worth it. Long way of saying 👍

ext/docker/runmount_buildpack.go
```diff
@@ -0,0 +1,125 @@
...
+// Detect implements Buildpack on RunmountBuildpack
+func (rmbp *RunmountBuildpack) Detect(ctx *sous.BuildContext) (*sous.DetectResult, error) {
+	dfPath := filepath.Join(ctx.Source.OffsetDir, "Dockerfile")
+	if !ctx.Sh.Exists(dfPath) {
+		return nil, fmt.Errorf("%s does not exist", dfPath)
+	}
+
+	messages.ReportLogFieldsMessage("Runmount dockerfile detection", logging.DebugLevel, logging.Log, dfPath)
+
+	dockerContent, _ := readDockerfile()
```
samsalisbury: We can't ignore this error, if the read fails the rest of this method is meaningless, so need to return early if err != nil.

ext/docker/runmount_builder_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+func TestRunmountBuild_extractFiles(t *testing.T) {
+	testRunSpec := getTestRunSpec()
+	sh, ctl := shell.NewTestShell()
+	_, cctl := ctl.CmdFor("docker", "cp")
+	cctl.ResultSuccess("", "")
+	ctx := sous.BuildContext{
+		Sh: sh,
+	}
+	builders, _ := constructImageBuilders(testRunSpec)
```
samsalisbury: I don't think we should ignore this error, if it's somehow not nil in this simple case we really need to know.

ext/docker/runmount_builder_test.go
```diff
@@ -0,0 +1,141 @@
...
+	err := extractFiles(ctx, testContainerBuildID, "/tmp", builders)
+	assert.NoError(t, err)
+}
+
+func TestRunmountBuild_templateDockerfile(t *testing.T) {
+	testRunSpec := getTestRunSpec()
+	buildDir := "/tmp"
+	builders, _ := constructImageBuilders(testRunSpec) //could abstract this for testing
```
samsalisbury: OK, I'm sounding like a broken record at this point (maybe those other comments are below though) I don't think there are good reasons to ignore errors, even in tests. If it's making tests noisy, a good idea is to add a test helper func that wraps this, takes a `*testing.T` and always asserts the error is nil, and then call that instead. E.g.:

```
func mustConstructImageBuilders(t *testing.T, runspec MultiImageRunSpec) []*runnableBuilder {
    t.Helper()
    r, err := constructImageBuilders(runspec)
    if err != nil {
        t.Fatal(err)
    }
    return r
}
```

ext/docker/runmount_builder.go
```diff
@@ -0,0 +1,317 @@
...
+		msg := "Deploy spec invalid:"
+		for _, f := range flaws {
+			msg += "\n\t" + f.Repair().Error()
+		}
+		return errors.New(msg)
+	}
+	return nil
+}
+
+func constructImageBuilders(runSpec MultiImageRunSpec) ([]*runnableBuilder, error) {
```
samsalisbury: Should remove the `error` result parameter as it's always nil.

ext/docker/runmount_builder_test.go
samsalisbury: I have just realised that `constructImageBuilders` always returns a nil error, therefore instead that second result parameter should be removed to avoid all these underscores.

ext/docker/runmount_builder_test.go
samsalisbury: Just saw that this method always returns nil error, so the right thing is to remove that return value. The above may still be interesting though?

ext/docker/runmount_builder.go
```diff
@@ -0,0 +1,314 @@
...
+	"strings"
+
+	sous "github.com/opentable/sous/lib"
+)
+
+func build(ctx sous.BuildContext) (string, error) {
+	fmt.Println("starting runmount build")
+
+	cmd := []interface{}{"build"}
+	// if localImage == false {
```
JC1738: do you still need commented code?

ext/docker/runmount_builder.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	// fmt.Println("output : ", output)
+
+	// TODO LH need to figure out what the end state of this should be.
+	// Think it needs to detect failure, should test this and return error
+	return nil
+
+}
+
+// need to create the container with the mount and then copy out of it
```
JC1738: Is this a TODO?

ext/docker/runmount_builder_test.go
```diff
@@ -0,0 +1,139 @@
...
+var testRunID = "A90110"
+
+func TestRunmountBuilder_build(t *testing.T) {
+	sh, ctl := shell.NewTestShell()
+	_, cctl := ctl.CmdFor("docker", "build")
+	cctl.ResultSuccess(fmt.Sprintf("Successfully built %s", testBuildID), "")
+	ctx := sous.BuildContext{
+		Sh: sh,
+	}
+	buildID, _ := build(ctx)
```
samsalisbury: If this test fails, but we didn't assert the error was nil, it might be hard to understand why it failed... Better to explicitly fail on the non-nil error as well to give future maintainers more to go on.

ext/docker/runmount_builder_test.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	containerID, err := createMountContainer(ctx, testBuildID)
+	assert.Empty(t, err)
+	assert.Equal(t, testContainerBuildID, containerID)
+}
+
+func getTestRunSpec() MultiImageRunSpec {
+	testRunSpec := MultiImageRunSpec{}
+	specF, _ := os.Open("./testdata/runmountbuilder/run_spec.json")
```
samsalisbury: `os.Open` can fail for loads of reasons, ignoring that here doesn't seem wise (failure could look like `run_spec.json` is an empty file in the case that there is a permissions error, or other). The more relevant info in a test failure the less time people will need to spend working out the cause :)

samsalisbury: This might be a good candidate for the test helper pattern I mentioned in another comment, where you can just fail the test in this func rather than making the calling test handle the error explicitly.

ext/docker/runmount_builder_test.go
```diff
@@ -0,0 +1,139 @@
...
+	tempDir := "./testdata/runmountbuilder"
+	testRunSpec := getTestRunSpec()
+	sh, ctl := shell.NewTestShell()
+	_, cctl := ctl.CmdFor("docker", "cp")
+	cctl.ResultSuccess("", "")
+	ctx := sous.BuildContext{
+		Sh: sh,
+	}
+
+	runSpec, _ := extractRunSpec(ctx, tempDir, testContainerBuildID)
```
samsalisbury: I'm just harping on the same point now... :P

ext/docker/runmount_buildpack_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	sous "github.com/opentable/sous/lib"
+	"github.com/opentable/sous/util/logging"
+	"github.com/opentable/sous/util/shell"
+	"github.com/stretchr/testify/assert"
+)
+
+// return the directory for cleanup
+func createDockerfile(content string) string {
+	dir, _ := ioutil.TempDir("", "sous-test")
```
samsalisbury: `ioutil.TmpDir` can definitely fail with a useful error... (See line 94 here https://golang.org/src/io/ioutil/tempfile.go?s=2138:2195#L94 0

ext/docker/runmount_buildpack_test.go
```diff
@@ -0,0 +1,64 @@
...
+	f.WriteString(content)
+	return dir
+}
+
+func removeDockerfile(dir string) {
+	os.RemoveAll(dir)
+}
+
+func getContext(dir string) *sous.BuildContext {
+	sh, _ := shell.Default()
```
samsalisbury: image

ext/docker/runmount_buildpack_test.go
```diff
@@ -0,0 +1,64 @@
...
+	return ctx
+}
+
+func TestRunmountBuildpack_DetectRunmount(t *testing.T) {
+	dir := createDockerfile("FROM docker.otenv.com/sous-otj-autobuild-runmount:local")
+	defer removeDockerfile(dir)
+
+	rmbp := NewRunmountBuildpack(logging.SilentLogSet())
+
+	result, _ := rmbp.Detect(getContext(dir))
```
samsalisbury: Same here..

ext/docker/runmount_buildpack_test.go
```diff
@@ -0,0 +1,64 @@
...
+	assert.True(t, result.Compatible)
+}
+
+func TestRunmountBuildpack_DetectNotRunmount(t *testing.T) {
+	dir := createDockerfile("FROM docker.otenv.com/sous-otj-autobuild:local")
+	defer removeDockerfile(dir)
+
+	rmbp := NewRunmountBuildpack(logging.SilentLogSet())
+
+	result, _ := rmbp.Detect(getContext(dir))
```
samsalisbury: ...

PR #749, opened 2018-06-20
HTTP writer now records updates to defs

nyarly:


sous-machine-account:
# LGTM!

</p>

lib/defs.go
```diff
@@ -0,0 +1,190 @@
...
+	vs = append(vs, prefixed("metadata ", ds.Metadata.Diff(o.Metadata))...)
+	vs = append(vs, prefixed("envdefs ", ds.EnvVars.Diff(o.EnvVars))...)
+
+	return vs
+}
+
+// Diff reports the differences.
+func (evs EnvDefs) Diff(os EnvDefs) []string {
+	if len(evs) != len(os) {
+		return []string{"lengths differ"}
```
samsalisbury: I can see that this avoids having to loop through both `evs` and `os`, but the output is a lot more informative if we always list specific items that are different rather than just that they have different lengths. Not blocking just an observation.

lib/defs.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	return vs
+}
+
+// Diff reports the differences.
+func (ed EnvDef) Diff(o EnvDef) []string {
+	vs := []string{}
+	if ed.Name != o.Name {
+		vs = append(vs, "names differ")
```
samsalisbury: Most `Diff` funcs in Sous report the specific values that are different e.g. `name: this: "a"; other: "b"` which I find really helpful especially in test failures. Not blocking obvs but might be worth adding eventually.

test/http_state_manager_test.go
```diff
@@ -171,4 +177,25 @@ func TestWriteState(t *testing.T) {
...
+
+	assert.ElementsMatch(t, onames, nnames)
+
+	t.Logf("Cluster names: %q", onames)
+
+	for _, n := range onames {
+		oc, nc := ocs[n], ncs[n]
+
+		assert.ElementsMatch(t, oc.AllowedAdvisories, nc.AllowedAdvisories)
+		t.Logf("Cluster advisories: %q: %q", n, oc.AllowedAdvisories)
```
samsalisbury: I removed a similar line to this in another PR, as the output was confusing (it was printed right after a FAIL) and led me to think this was the cause of the failure briefly. I think we should remove it here as well, as the assertion above will print relevant info if there is a problem.

PR #756, opened 2018-06-26
Add artifact command line

JC1738:
Work in progress, this is the command add and subcommand artifact to add docker image to sous.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

JC1738:
@samsalisbury Got an actual deploy working all the way through in dev-ci-sf.  Details commented on https://github.com/samsalisbury/otpl-deploy-scripts/commit/5f34ae819df30a150936771191c10bc43011a01e This branch needs a review and the advisory for the artifact isn't currently being set, the reason is needed to figure out a mechanism for whitelisting advisory once set, so just left a blank for testing for now.

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,61 @@
...
+	"strings"
+
+	"github.com/opentable/sous/config"
+	sous "github.com/opentable/sous/lib"
+	"github.com/opentable/sous/util/logging"
+	"github.com/opentable/sous/util/logging/messages"
+	"github.com/opentable/sous/util/shell"
+)
+
+//AddArtifact struct for normalize GDM action.
```
samsalisbury: Comment needs updating.

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,61 @@
...
+	Tag         string
+	DockerImage string
+	LocalShell  shell.Shell
+	LogSink     logging.LogSink
+	User        sous.User
+	Inserter    sous.ClientInserter
+	*config.Config
+}
+
+//Do executes the action for plumb normalize GDM.
```
samsalisbury: Same here.

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,61 @@
...
+	versionName := fmt.Sprintf("%s:%s", a.DockerImage, a.Tag)
+	messages.ReportLogFieldsMessageToConsole(fmt.Sprintf("Attempting to get Digest for: %s", versionName), logging.ExtraDebug1Level, a.LogSink)
+
+	//Now figure out the digest
+	output, err := a.LocalShell.Stdout("docker", "inspect", "--format={{index .RepoDigests 0}}", versionName)
+	if err != nil {
+		return err
+	}
+
+	digestName := strings.Trim(output, " \n\t\r")
```
samsalisbury: `strings.TrimSpace` is a built-in way to do this.

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,61 @@
...
+func (a *AddArtifact) UploadArtifact(versionName, digestName string) error {
+
+	art := sous.BuildArtifact{
+		VersionName:     versionName,
+		DigestReference: digestName,
+		Qualities:       []sous.Quality{{"otpl_tag", "advisory"}},
+	}
+
+	//TODO:  Does directory need to be present?
+	sid := sous.MakeSourceID(a.Repo, "", a.Tag)
```
samsalisbury: Since each directory can output a separate deployable artifact, we do need to include it in the identifier here.

cli/sous_add_artifact.go
```diff
@@ -0,0 +1,47 @@
...
+	opts      graph.ArtifactOpts
+}
+
+func init() { AddSubcommands["artifact"] = &SousAddArtifact{} }
+
+// Help prints the help
+func (*SousAddArtifact) Help() string {
+	return `Artifact for docker image to deploy.
+
+This is to add the artifact to sous for deploy.
```
samsalisbury: Help text should possibly not mention deploying, or at least should say something like: "Tell sous that this docker image represents a particular SourceID."

vendor/vendor.json
```diff
@@ -373,6 +385,306 @@
 			"revision": "1338e7e3e2c5144a40409b80e687e0532b9c656a",
 			"revisionTime": "2016-08-20T15:25:38Z"
 		},
+		{
+			"checksumSHA1": "75qVidE3jtMufGt8+r3bOhI13lo=",
+			"path": "github.com/opentable/sous",
```
samsalisbury: Seems odd that we've vendored sous into itself?

JC1738: @samsalisbury  Yep, weird didn't mean too, I just pushed a deletion of this from the vendor.json

cli/actions/add_artifact.go
JC1738: thanks, pushed new comments

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+//UploadArtifact add artifact to sous servers
+func (a *AddArtifact) UploadArtifact(versionName, digestName string) error {
+
+	art := sous.BuildArtifact{
+		VersionName:     versionName,
+		DigestReference: digestName,
+		Qualities:       []sous.Quality{{}},
+		//		Qualities:       []sous.Quality{{"otpl_tag", "advisory"}},
```
JC1738: added a new advisory

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	//TODO:  Does directory need to be present?
```
JC1738: fixed, padded offset through now

cli/sous_add_artifact.go
JC1738: Made change of wording of help.

JC1738:
@samsalisbury Thanks for the review, I think I addressed all your items, it should be ready when you get a chance to take a look again.  Thanks!

JC1738:
As far as code coverage, not sure why it's complaining, I added tests as well as if you click on details: https://codecov.io/gh/opentable/sous/compare/ea1517db235a3ad9e6223100c6f6c13626302612...7d94921ebe8041e1a1b16c3a385461c14ed38a83 it actually says missing base report can't give diffs.

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,62 @@
...
+	Inserter    sous.ClientInserter
+	*config.Config
+}
+
+//Do executes the action for add artifact.
+func (a *AddArtifact) Do() error {
+
+	messages.ReportLogFieldsMessageToConsole(fmt.Sprintf("Executing add artifact Repo: %s, Cluster: %s, Offset: %s, Tag: %s, DockerImage: %s", a.Repo, a.Cluster, a.Offset, a.Tag, a.DockerImage), logging.ExtraDebug1Level, a.LogSink)
+
+	versionName := fmt.Sprintf("%s:%s", a.DockerImage, a.Tag)
```
samsalisbury: I didn't notice this until I added the smoke test, but I think it may be simpler to just accept a full Docker Image name for the `-docker-image` flag rather than it accepting just the Docker Repository name, and forcing the same `-tag` to be used for both the docker image and the sous version tag. This would mean that _any_ docker image could be registered for a given SourceID rather than just one with matching tag... It should be a small change but want to discuss it before making that change here.

cli/actions/add_artifact.go
```diff
@@ -0,0 +1,61 @@
...
+	Inserter    sous.ClientInserter
+	*config.Config
+}
+
+//Do executes the action for add artifact.
+func (a *AddArtifact) Do() error {
+
+	messages.ReportLogFieldsMessageToConsole(fmt.Sprintf("Executing add artifact Repo: %s, Offset: %s, Tag: %s, DockerImage: %s", a.Repo, a.Offset, a.Tag, a.DockerImage), logging.ExtraDebug1Level, a.LogSink)
+
+	versionName := fmt.Sprintf("%s:%s", a.DockerImage, a.Tag)
```
samsalisbury: Re-pasting here (last time hit the wrong revision :/) I didn't notice this until I added the smoke test, but I think it may be simpler to just accept a full Docker Image name for the `-docker-image` flag rather than it accepting just the Docker Repository name, and forcing the same `-tag` to be used for both the docker image and the sous version tag. This would mean that _any_ docker image could be registered for a given SourceID rather than just one with matching tag... It should be a small change but want to discuss it before making that change here.

JC1738: I think I see what you mean, we would remove -tag altogether and just use docker-image to specify docker registry?  so it would be sous add artifact -repo github.com/company/project -offset myoffset?   So only two flags... one required?

samsalisbury: Would still need -tag as that's the Sous tag which we need to identify which source code revision the docker image represents. Would just mean the docker image can have any tag, which may or may not be necessary, but allowing that would ensure it works with any old docker tagging schema...

samsalisbury: In short, need enough to identify a SourceID (sous term for semver tag + SourceLocation(Repo, Offset)) that the docker image belongs to. Since we expect to have to run this without a git context, we need the -repo and -offset flags (repo required) to identify SourceLocation, and the -tag flag as mentioned.

JC1738: Yes, might be slightly confusing for end user (expecting both to be same), but I get it.  Maybe if you don't specify tag, it gets parsed from docker-image?  Or should we just be explicit since this command isn't going to be called manually most likely but wrapped in a script?

samsalisbury: Right, at the moment there is no constraint on Docker tags used in otpl-deploy, so they may or may not be semver (most are not from experience). So if we keep this as-is, it will probably force a lot of teams to adopt a new tagging scheme that's semver-compatible. I kind of like making people use semver just for the sake of uniformity, but I think it will slow adoption. I'm honestly not sure which is the right way to go... Having tags be the same definitely seems like a win for UX.

samsalisbury: This is a working implementation, let's see if there's a real issue before making any changes.

PR #759, opened 2018-07-05
GET artifact endpoint

JC1738:
created a PR so I can add comments, WIP, not ready for merge, added trivial so wouldn't bulk on changelog


server/handle_artifact.go
```diff
@@ -31,6 +40,31 @@ func newArtifactResource(ctx ComponentLocator) *ArtifactResource {
 	return &ArtifactResource{context: ctx}
 }
 
+// Get implements Getable on GDMResource
+func (ar *ArtifactResource) Get(_ *restful.RouteMap, ls logging.LogSink, writer http.ResponseWriter, req *http.Request, _ httprouter.Params) restful.Exchanger {
+	return &GETArtifactHandler{
+		Request:     req,
+		LogSink:     ls,
+		QueryValues: ar.ParseQuery(req),
+		Registry:    ar.context.Registry,
```
JC1738: I am seeing Registry come back as NULL, which I understand kindof, but I am doing the same thing as ar.context.Inserter and it is coming back populated, so how did Inserter get it's value set but Registry doesn't?  I imagine it's the graph, but both are explicit values it looks like in graph.go?  @samsalisbury

sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR: 882 lines of code changed</td>
    </tr>
  </tbody>
</table>

</p>

server/handle_artifact.go
```diff
@@ -31,6 +40,31 @@ func newArtifactResource(ctx ComponentLocator) *ArtifactResource {
...
+
+	ba, err := gh.GetArtifact(sid)
+	if err != nil {
+		return err, http.StatusNotAcceptable
+	}
+
+	return ba, http.StatusOK
+}
+
 // Put implements Putable on ArtifactResource, which marks it as accepting PUT requests
```
JC1738: The other problem I am seeing, is despite being NULL, just wanted to do a PUT which was working prior to this branch.  But after adding the Get Exchange Handler, it seems to get called despite the PUT call being executed against Artifact:

`curl -X PUT \
  'http://jc-linux.otcorp.opentable.com:8888/artifact?offset=&repo=github.com%2Fopentable%2Fhello&version=0.0.3' \
  -H 'Cache-Control: no-cache' \
  -H 'Content-Type: application/json' \
  -H 'If-None-Match: *' \
  -H 'Postman-Token: 1801308f-f796-40ba-8edb-0bd0f8885ebd' \
  -d '{"Type":"","VersionName":"docker.otenv.com/hello:0.0.3","DigestReference":"docker-ci-sf.otenv.com/hello@sha256:c6fd3d6af255de2309e541ed0affb3b6d3ef6024241b79a51ded94dd11505657","Qualities":[{"Name":"otpl_tag","Kind":"advisory"}]}'`

JC1738: My tracing looks like it might have something to do with the If-None-Match *, but I don't understand enough why it would skip the PUT exchange and decide to use the GET exchange instead?

JC1738: Figured out the u/r/router.go issue, line 189 mh.SynthResponse.  The question I would have, if I have If-None-Match *, couldn't we skip the get all together?

server/handle_artifact.go
JC1738: Ok, got Registry Populated, but seems like a sort of hack in graph.go when creating server force the setting Registry.

server/handle_artifact.go
JC1738: I got further, I now don't error in Get on the synthResponse by checking if NoImageFound error, and just passing 200 and nil back, but it still is returning and not executing Put with the message:  "412 Preconditioned Failed and body resource present for If-None-Match=*!"

JC1738: And yes, fixed this, get needs status not found to bypass e-tag trap.

JC1738: Ok, was not easy (for me) and still don't know why I had to do hack to get Registry populated, but Get Artifact now works :)

server/handle_artifact.go
samsalisbury: Yes that looks odd, I'll see if I can get it to just work via the component locator in the graph to avoid the hack, not obvious why it should be needed :/

PR #760, opened 2018-07-09
tests: downgrade Singularity to 0.16.2 #trivial

samsalisbury:
- Stability issues with 0.20.1 mean we're parking upgrades for now.


samsalisbury:
Closing as probably not needed yet, leaving branch in case we do...

PR #766, opened 2018-07-13
Sous 1011 singularity link

lachlanh:


sous-machine-account:
# LGTM!

</p>

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ with respect to its command line interface and HTTP interface
 ### Changed
 * Client: when receiving bad content type print the HTTP status code and
   text along with that error message.
-
+* Both: Send the singularity request URL back to the client and print to
```
JC1738: might want to just say request URL to be agnostic

server/handle_single_deployment.go
```diff
@@ -112,9 +113,32 @@ func (h *GETSingleDeploymentHandler) Exchange() (interface{}, int) {
...
 
-	return h.ok(200, nil)
+	return h.ok(200, links)
+}
+
+func makeSingularityURL(baseURL string, singularityRequestID string, deploymentID sous.DeploymentID) string {
+	var err error
+	if len(singularityRequestID) == 0 {
+		singularityRequestID, err = singularity.MakeRequestID(deploymentID)
+		if err != nil {
```
JC1738: could log these errors since you aren't passing them up the chain...

server/handle_single_deployment.go
```diff
@@ -112,9 +113,32 @@ func (h *GETSingleDeploymentHandler) Exchange() (interface{}, int) {
...
+			return ""
```
samsalisbury: I wonder if we should return this error to the user. Main reason being I can see people wanting to use this value in automated scripts, so just silently dropping it would be more confusing than, even, setting the URL to something like `fmt.Sprintf("Unable to determine SingularityRequest URL: %s", err)`.

server/handle_single_deployment.go
```diff
@@ -112,9 +113,32 @@ func (h *GETSingleDeploymentHandler) Exchange() (interface{}, int) {
 		return h.err(400, "Cannot deploy, current num instances set to zero, please update manifest.")
 	}
 
+	links := make(map[string]string)
+	baseURL := h.GDM.Defs.Clusters[did.Cluster].BaseURL
+	singularityRequestID := dep.DeployConfig.SingularityRequestID
+	singularityURL := makeSingularityURL(baseURL, singularityRequestID, did)
+	if len(singularityURL) > 0 {
```
samsalisbury: Just curious if there's a reason to do this rather than `if singularityURL != ""` ? I personally find that easier to read, not blocking over this though.

server/handle_single_deployment.go
```diff
@@ -112,9 +113,32 @@ func (h *GETSingleDeploymentHandler) Exchange() (interface{}, int) {
...
+
 	h.Body.Deployment = &dep
 
-	return h.ok(200, nil)
+	return h.ok(200, links)
+}
+
+func makeSingularityURL(baseURL string, singularityRequestID string, deploymentID sous.DeploymentID) string {
+	var err error
+	if len(singularityRequestID) == 0 {
```
samsalisbury: Likewise, `if singularityRequestID == "" {` seems simpler, unless there was another reason to do this?

server/handle_single_deployment.go
```diff
@@ -112,9 +113,32 @@ func (h *GETSingleDeploymentHandler) Exchange() (interface{}, int) {
...
+		singularityRequestID, err = singularity.MakeRequestID(deploymentID)
+		if err != nil {
+			return ""
+		}
+	}
+	singularityURL, err := singularity.MakeRequestURL(baseURL, singularityRequestID)
+	if err != nil {
+		singularityURL = ""
```
samsalisbury: Likewise, might be better to even just replace the URL with an error string here, would rather break someone's parsing and/or have someone tell us it isn't working than just look like we've dropped the feature...

PR #777, opened 2018-07-26
Disable --pull from docker build with env var + disable autoresolver in tests

samsalisbury:
If new env var `SOUS_BUILD_NOPULL` == `YES` then docker builds omit the `--pull` flag. This is used to speed up smoke tests and enable running them without an internet connection. (Note you still need an internet connection to run liquibase tasks, which need to be run before tests.) Simply using the `-dev` flag could have other implications in future as that flag has a broader meaning, so did not want to use that flags in tests.

In smoke tests, more commands are now printed out as the tests run which makes logs more useful, specifically `sous server` and many `git` commands now make it to the output, along with the directory they're being run in. This helped hunt down why the autoresolver was running unexpectedly in tests, was a simple missing `-` character :/


sous-machine-account:
# LGTM!

</p>

test/smoke/sousserver.go
```diff
@@ -95,12 +99,10 @@ func (i *sousServer) Start(t *testing.T) {
 	if !quiet() {
 		fmt.Fprintf(os.Stderr, "==> Instance %q config:\n", i.ClusterName)
 	}
-	// Run 'sous config' to validate it.
-	i.MustRun(t, "config", nil)
 
 	serverDebug := os.Getenv("SOUS_SERVER_DEBUG") == "true"
 
-	i.Service.Start(t, "server", nil, "-listen", i.Addr, "-cluster", i.ClusterName, "autoresolver=false", fmt.Sprintf("-d=%t", serverDebug))
+	i.Service.Start(t, "server", nil, "-listen", i.Addr, "-cluster", i.ClusterName, "-autoresolver=false", fmt.Sprintf("-d=%t", serverDebug))
```
JC1738: yep, hard to see missing -, good catch

PR #790, opened 2018-08-07
Jenkinscli

JC1738:
This isn't ready to be merged, but @samsalisbury might be worth a look to see approach, there is an assumption on cluster that should be validated.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

cli/actions/jenkins.go
```diff
@@ -0,0 +1,65 @@
...
+	//Write out Jenkins
+	//Push back metadata
+
+	//for now going to assume metadata for Jenkins file is CI-SF located, can change this in future
+	currentConfigMap := make(map[string]string)
+	mani := sous.Manifest{}
+	_, err := sj.HTTPClient.Retrieve("/manifest", sj.TargetManifestID.QueryMap(), &mani, nil)
+
+	// Eventually will make this configuration data in config.yaml
+	clusterWithJenkinsConfig := "CI-SF"
```
samsalisbury: Hmm, I think for now we should write this config to all clusters every time. It's duplicated info, but we don't necessarily have a ci-sf cluster so just doing this isn't safe. (Alternatively could sort the clusters alphabetically and always write to the first in that list, however what if that cluster gets removed? We don't support removal yet but still...). The real fix for this is to introduce a new project-level configuration blob (scoped to SourceLocation) that would provide CI config. That will take time though, for now I think using lowest alphabetically sorted cluster or all clusters is the best thing to do.

samsalisbury: Also, our cluster names are all lowercase by default so `ci-sf` not `CI-SF`

JC1738: @samsalisbury Yep, wasn't going to leave the hard coding, it was there to show what happening.  I actually was forced to add -cluster flag, so I am going to use that to pick where, so they can generate the file based off whatever -cluster they so choose.

JC1738:
Not sure what's with code coverage, that was the result before adding test.  After adding test, it didn't recalculate?

cli/actions/jenkins.go
```diff
@@ -0,0 +1,937 @@
...
+		"SOUS_VERSIONING_SCHEME":        "semver_timestamp",
+		"SOUS_JENKINSPIPELINE_VERSION":  "0.0.1",
+		"SOUS_RELEASE_BRANCH":           "master",
+		"SOUS_DEPLOY_PROD_QUERY_USER":   "YES",
+	}
+}
+
+func (sj *Jenkins) returnTemplate() string {
+
+	var template = `#!/usr/bin/env groovy
```
lachlanh: Really would like to see the template split out of the source for this.

lachlanh: Also we used @nyarly 's template builder for the above : "github.com/nyarly/inlinefiles/templatestore" do we need a new one ?

lachlanh:
Added comments around how the template is used. Other than that looks good.

cli/actions/jenkins.go
JC1738: After looking at the templatestore and .tmpl file option, given how much this is going to change, not sure going the route of *.tmpl and innlinefiles/templatestore (they are used in conjunction) is the best method @lachlanh.  Might be a good thing to discuss?  Would prefer to merge as is and open a new PR for a refactor of that type.

JC1738:
@samsalisbury @nyarly  There is an actual problem with a unit test, though I am not sure if this PR introduced it or not.  Do either recognize what this is doing, I am not too familiar with this test or what is happening here.  cc:/ @lachlanh 

Logging HTTP bodies to "/tmp/http-log244065062"
--- FAIL: TestResolveRecorder (0.01s)
    --- FAIL: TestResolveRecorder/Six_phases,_fourth_fails (0.00s)
        resolvestatus_test.go:138: expected log of resolutions to have 1 entries; has 0

PR #791, opened 2018-08-15
Sous 1007 remove db from unit tests #trivial

lachlanh:


sous-machine-account:
# LGTM!

</p>

Makefile
```diff
@@ -342,15 +342,13 @@ test-gofmt:
...
-	PGHOST=$(PGHOST) \
-	PGPORT=$(PGPORT) \
+test-unit-base: $(COVER_DIR) $(GO_FILES)
 	go test $(EXTRA_GO_TEST_FLAGS) $(EXTRA_GO_FLAGS) $(TEST_VERBOSE) \
 	  -tags=netcgo -covermode=atomic -coverprofile=$(COVER_DIR)/count_merged.txt \
 		-timeout 12m -race $(GO_TEST_PATHS) $(TEST_TEAMCITY)
 
 .PHONY: test-unit
-test-unit: postgres-clean test-unit-base
+test-unit: test-unit-base
```
samsalisbury: Nice, so we can rename test-unit-base test-unit now I believe..

ext/docker/image_mapping_test.go
```diff
@@ -1,3 +1,5 @@
+// +build integration
```
samsalisbury: I feel like there's a lot of logic being tested in this file that it would be nice to still test under unit tests. Having said that it would be a lot of work, so this is probably the best thing to do pragmatically.

lachlanh:
@samsalisbury this pr feels like a bit of a compromise, but it is passing on jenkins so should be ready to merge.

PR #800, opened 2018-10-22
Manifest validation now requires non-empty Owners field

yk12OT:


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ with respect to its command line interface and HTTP interface
 
 ## [Unreleased](//github.com.opentable/sous/compare/0.5.117...master)
 ### Fixed
+* Manifest validation now requires non-empty Owners field.
```
nyarly:
```suggestion
* Manifest validation now requires non-empty Owners field. NOTE: this will be a user experience hit, and needs to be highlighted to our users before deploy.
```

lib/manifest_test.go
```diff
@@ -153,6 +166,18 @@ var manifestTests = []struct {
 	},
 }
 
+func TestManifest_ValidateFatalFlaw(t *testing.T) {
```
lachlanh: for consistency it would be good to get this into the table test above, otherwise would like to have a different name for the test function, include something about the nature of the test. TestManifest_ValidateNoOwnersExpectFatalFlaw something like that.

sous-machine-account:
# LGTM!

</p>

nyarly:
@yk12OT It looks like there are several integration/smoke tests that are also broken by this change. They'll need to be updated as well.

cli/add_flags_test.go
```diff
@@ -19,6 +19,8 @@ func TestAddFlagsForRectify(t *testing.T) {
         flavor is a short string used to differentiate alternative deployments
   -offset string
         source code relative repository offset
+  -owners string
+        list of owners (comma separated)
```
samsalisbury: I really don't think comma-separated is a good idea here, parsing email addresses is non-trivial, and they can contain commas inside quoted strings. I hope my comments in Slack against doing this weren't misunderstood. See https://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-an-email-address for some discussion...

cli/rectify_test.go
```diff
@@ -16,10 +16,13 @@ func TestPredicateBuilder(t *testing.T) {
 	rs := []string{"github.com/ot/one", "github.com/ot/two"}
 	os := []string{"up", "down"}
 
+	var emptyStruct struct{}
```
samsalisbury: I would just use `struct{}{}` wherever you need an empty struct... Saves a character of typing as well!

samsalisbury: But the main reason is it's more idiomatic go, there is no question what is meant when reading `struct{}{}`, whereas `emptyStruct` provides indirection (we need to look up the definition) for no gain.

samsalisbury: If we want adding to the set to be less verbose, the better thing would be to add a method on `sous.OwnerSet` e.g. an `Add` method just taking the value to be added and encapsulating adding the `keyName: struct{}{}` mapping. E.g. `owners.Add("some@owner")`

cli/deploy_filter_flags.go
```diff
@@ -33,6 +33,10 @@ const (
 	-cluster CLUSTER
 		the deployment environment to consider`
 
+	ownersFlagHelp = `
+	-owners OWNERS
+		the list of owners emails (semicolon separated)`
```
samsalisbury: Semicolon is allowed in email addresses (inside quoted strings), it may be uncommon, but we need to allow the possibility. The complexity of parsing out email addresses is too great to be tagged on to this PR, we need to accept only a single email address per flag.

cli/deploy_filter_flags.go
```diff
@@ -35,7 +35,7 @@ const (
 
 	ownersFlagHelp = `
 	-owners OWNERS
-		the list of owners emails (semicolon separated)`
+		the list of owners emails (colon separated)`
```
samsalisbury: Colon is allowed in email addresses @yk12OT as mentioned on the previous comment/review, we need to avoid adding the complexity of parsing out email addresses to this PR.

cli/sous_init.go
```diff
@@ -111,6 +112,7 @@ func (si *SousInit) Execute(args []string) cmdr.Result {
 	}
 
 	m.Kind = kind
+	m.Owners = strings.Split(si.flags.Owners, ":")
```
samsalisbury: This might work in most cases, but there are cases where it won't work since colon is allowed in email addresses.

samsalisbury:
@yk12OT this looks good, thanks. I don't want to merge this just yet, as I would rather notify all the teams who will be affected by this before we release (and we have a backlog to release already), so should be able to merge later today or tomorrow.

PR #801, opened 2018-10-23
Release 0.5.120

samsalisbury:
We somehow skipped changelog update for 0.5.118 and 0.5.119; this PR also adds a late check that the changelog at least contains the tagged version somewhere prior to release. It's a weak check and happens late in the process. If it fails, we'll need to add a new semver tag and update the changelog to use that. Unfortunately there is nowhere earlier in the process we can detect this reliably since adding a semver tag is enough to trigger a binary release (first I considered using Danger, but that only applies to PRs not tags). Once an annotated tag is pushed it should not be changed, although I am not hardline on that in the case of failed builds, if it is changed quickly (since this is not a library we would be unlikely to break dependencies). Once any binary exists for a version in github releases, the tag should definitely not be modified under any circumstances.

In future it might be helpful to open PRs like this that update the changelog prior to tagging releases.

NOTE: This PR can be safely merged, but the merge commit should be tagged 0.5.120, which will trigger binary builds. ~That tagging is not safe to do yet, as changes here will have an impact on users so we need to communicate that first.~ EDIT: That's not true, this should be safe to tag, I was thinking #800 had been merged but not yet...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ env:
     - TEST_COMMAND="GO_TEST_RUN=TestSmoke/git/simple make test-smoke"
     - TEST_COMMAND="GO_TEST_RUN=TestSmoke/git/split make test-smoke"
 before_deploy: git clean -fd && git reset --hard HEAD && go get github.com/karalabe/xgo
-  && make semvertagchk && bundle exec make -j2 release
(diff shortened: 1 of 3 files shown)


sous-machine-account:
# LGTM!

</p>

PR #804, opened 2018-10-26
smoketest tidy up #trivial

samsalisbury:
Tidying up loose ends whilst adding tests for SOUS-1020. This may speed up running smoke tests.


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

samsalisbury:
Coverage failure here is caused by a race condition (not a data race that race detector might pick up) causing a line of code not to always be executed. We should fix except that code is likely to disappear in the medium term so not sure it's worth sinking any more time into it (it's come up a few times before and each time we think it's fixed it crops up again sooner or later).

samsalisbury:
It's marked #trivial as it only affects tests not the function of sous itself; just updated with master will merge once tests pass..

PR #805, opened 2018-10-26
Corrects confusing error message when manifest not found

samsalisbury:
Resolves SOUS-1020; includes #804; See https://github.com/opentable/sous/pull/805/files#diff-91432d762aa99853969261d3158edbb4R15 for test case covering the bug report, and https://github.com/opentable/sous/pull/805/commits/e13914ca4845dc165bbe1da6887be4c8e7f07aac for the resolution.

PR #806, opened 2018-10-29
Bugfix docker ref casing

samsalisbury:
Includes #805 to avoid merge conflicts. See changelog diff for details.

### DIFF
--- a/ext/docker/names_test.go
+++ b/ext/docker/names_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	sous "github.com/opentable/sous/lib"
+	"github.com/opentable/sous/util/logging"
 )
 
 func TestImageRepoName(t *testing.T) {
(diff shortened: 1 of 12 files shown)


sous-machine-account:
# LGTM!

</p>

ext/docker/names_test.go
```diff
@@ -0,0 +1,94 @@
...
+		// Kind: docker
+		{
+			"docker",
+			sous.SourceLocation{Repo: "github.com/user1/repo1"},
+			"example.org/user1/repo1-docker",
+		},
+		{
+			"docker",
+			sous.SourceLocation{Repo: "github.com/User1/Repo1"},
+			"example.org/user1/repo1-docker",
```
nyarly: This makes me really nervous, since two projects with different SourceLocations build the same container.

_Github_ considers differently named repos the same, but that's not inherent to git (in fact I think quite the contrary.)

I'm not sure precisely how to address this. Maybe we could start rejecting sids with uppercase letters in the repo?

samsalisbury: Hm, I don't think we can just reject them since there are many git repos that have uppercase letters. Maybe for repos that include uppercase letters we append a string with the indexes of the upper-case letters e.g. 'github.com/opentable/MyRepo` becomes `github.com/opentable/myrepo--22-24`. Haha something inside me wants to do that, but a simple hash of the true repo name is probably best... Exactly how to do that we should discuss. We could also special case `github.com` repos to avoid tagging extra stuff on the name, and just do something much more ugly for non-github repos (e.g. just use a hash as the docker repo name and be done with it). Pragmatism makes me prefer that last option as we don't need to deal with tricky suffixes to repo names at all. If people want ergonomic names for other non-github repos, we could add that as a feature later.

samsalisbury: Code added...

1. Refactor to remove the `stripRE` parameter which was passed all the way through down the stack but in the end never referenced where it was intended to be. All values for `stripRE` were also identical at the top of the stack, so best to keep it clean and not pass this effective constant around.

2. As proposed above, GitHub URLs continue to be treated as a special case as they already were with the special regex. This preserves the right level of uniqueness for a GitHub repo -> docker image mapping (but see caveat below on offset dirs).

3. Non-github repos ("generic" in the code) are hashed to produce a 1-1 mapping between case-sensitive source repo name and all-lowercase docker repo name.

Caveat: single github repos that have 2 or more offset dirs (both producing deployables) that differ only by case will produce the same docker repo name. In practice this is very unlikely to happen, as it would be confusing in general to work with. However I think we'll have to cover that case as well, perhaps by treating any SourceLocation with a Dir that contains an uppercase character in the same way as we treat generic repos, i.e. just use a hash.

So I think there's a bit more work to do, but leaving in the current state for further discussion.

lachlanh:
This is a pretty big PR. @nyarly do you want to take another look through before it gets merged ? Looks good to me.

PR #808, opened 2018-11-07
Refactor of build strategy detection

nyarly:


nyarly:
This is in prep for refactoring runmount detection to look for BUILD_OUT instead of grepping for runmount in the Dockerfile

PR #811, opened 2018-11-15
sous scale command and 'query gdm' options

samsalisbury:
See changelog in this diff for the interesting parts of this PR.

It also includes a fair few extra unit tests for code not really modified by the change to increase coverage.

### DIFF
--- a/cli/actions/manifest_get_test.go
+++ b/cli/actions/manifest_get_test.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"bytes"
+	"fmt"
 	"os"
 	"testing"
 
@@ -11,43 +12,58 @@ import (
 	"github.com/opentable/sous/util/restful/restfultest"
 	"github.com/samsalisbury/semv"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestManifestGet(t *testing.T) {
-	out := &bytes.Buffer{}
-	project1 := sous.SourceLocation{Repo: "github.com/user/project"}
+	assertError := func(manifestReturned interface{}, gotHTTPErr, wantErr bool) {
+		var httpErr error
+		if gotHTTPErr {
+			httpErr = fmt.Errorf("an error")
+		}
+		out := &bytes.Buffer{}
+		project1 := sous.SourceLocation{Repo: "github.com/user/project"}
 
-	var up restful.Updater
-	cl, control := restfultest.NewHTTPClientSpy()
-	smg := &ManifestGet{
-		TargetManifestID: sous.ManifestID{
-			Source: sous.SourceLocation{
-				Repo: project1.Repo,
+		var up restful.Updater
+		cl, control := restfultest.NewHTTPClientSpy()
+		smg := &ManifestGet{
+			TargetManifestID: sous.ManifestID{
+				Source: sous.SourceLocation{
+					Repo: project1.Repo,
+				},
+				Flavor: "chocolate",
 			},
-			Flavor: "chocolate",
-		},
-		HTTPClient: cl,
+			HTTPClient: cl,
 
-		OutWriter:      out,
-		LogSink:        logging.NewLogSet(semv.MustParse("0.0.0"), "", "", os.Stderr),
-		UpdaterCapture: &up,
+			OutWriter:      out,
+			LogSink:        logging.NewLogSet(semv.MustParse("0.0.0"), "", "", os.Stderr),
+			UpdaterCapture: &up,
+		}
+
+		control.Any(
+			"Retrieve",
+			manifestReturned, restfultest.DummyUpdater(), httpErr,
+		)
+
+		err := smg.Do()
+		if wantErr {
+			assert.Error(t, err)
+			return
+		}
+		require.NoError(t, err)
+
+		if assert.Len(t, control.Calls(), 1) {
+			assert.Regexp(t, "/manifest", control.Calls()[0].PassedArgs().String(0))
+			params := control.Calls()[0].PassedArgs().Get(1).(map[string]string)
+			assert.Contains(t, params, "repo")
+			assert.Contains(t, params, "flavor")
+			assert.Equal(t, params["flavor"], "chocolate")
+		}
+
+		assert.Regexp(t, "github", out.String())
 	}
 
-	control.Any(
-		"Retrieve",
-		sous.ManifestFixture("simple"), restfultest.DummyUpdater(), nil,
-	)
+	assertError(sous.ManifestFixture("simple"), false, false)
+	assertError(sous.ManifestFixture("simple"), true, true)
 
-	err := smg.Do()
-	assert.NoError(t, err)
-
-	if assert.Len(t, control.Calls(), 1) {
-		assert.Regexp(t, "/manifest", control.Calls()[0].PassedArgs().String(0))
-		params := control.Calls()[0].PassedArgs().Get(1).(map[string]string)
-		assert.Contains(t, params, "repo")
-		assert.Contains(t, params, "flavor")
-		assert.Equal(t, params["flavor"], "chocolate")
-	}
-
-	assert.Regexp(t, "github", out.String())
 }
--- a/cli/queries/artifact.go
+++ b/cli/queries/artifact.go
@@ -1,0 +1,30 @@
+package queries
+
+import (
+	"strings"
+
+	sous "github.com/opentable/sous/lib"
+	"github.com/opentable/sous/util/restful"
+)
+
+// ArtifactQuery supports querying artifacts by their SourceID only.
+type ArtifactQuery struct {
+	Client restful.HTTPClient
+	User   sous.User
+}
+
+// ByID returns the single artifact matched by sid. It returns nil, nil if there
+// is no match and no error determining that.
+func (q *ArtifactQuery) ByID(sid sous.SourceID) (*sous.BuildArtifact, error) {
+	ba := &sous.BuildArtifact{}
+	header := q.User.HTTPHeaders()
+
+	_, err := q.Client.Retrieve("./artifact", sid.HTTPQueryMap(), ba, header)
+	if err == nil {
+		return ba, nil
+	}
+	if strings.HasPrefix(err.Error(), "404 Not Found") {
+		return nil, nil
+	}
+	return nil, err
+}
--- a/cli/queries/artifact.go
+++ b/cli/queries/artifact.go
@@ -23,7 +23,7 @@ func (q *ArtifactQuery) ByID(sid sous.SourceID) (*sous.BuildArtifact, error) {
 	if err == nil {
 		return ba, nil
 	}
-	if strings.HasPrefix(err.Error(), "404 Not Found") {
+	if strings.Contains(err.Error(), "404 Not Found") {
 		return nil, nil
 	}
 	return nil, err
--- a/cli/queries/artifact.go
+++ b/cli/queries/artifact.go
@@ -28,3 +28,11 @@ func (q *ArtifactQuery) ByID(sid sous.SourceID) (*sous.BuildArtifact, error) {
 	}
 	return nil, err
 }
+
+// Exists returns true if the artifact exists. If it returns a non-nil error,
+// the other return value is undefined, and it was not possible to determine
+// if the artifact exists or not.
+func (q *ArtifactQuery) Exists(sid sous.SourceID) (bool, error) {
+	a, err := q.ByID(sid)
+	return a != nil, err
+}
--- a/cli/queries/deployment.go
+++ b/cli/queries/deployment.go
@@ -87,7 +87,7 @@ func (q *Deployment) parseFilters(s string) ([]boundDeployFilter, error) {
 	for _, p := range parts {
 		kv := strings.Split(p, "=")
 		if len(kv) != 2 {
-			return nil, cmdr.UsageErrorf("filter %q not valid; format is <name>=(true|false)")
+           return nil, cmdr.UsageErrorf("filter %q not valid; format is <name>=(true|false)", p)
 		}
 		k, v := kv[0], kv[1]
 		f, err := q.getFilter(k)
--- a/cli/queries/deployment.go
+++ b/cli/queries/deployment.go
@@ -1,0 +1,165 @@
+package queries
+
+import (
+	"log"
+	"strconv"
+	"strings"
+	"sync"
+
+	sous "github.com/opentable/sous/lib"
+	"github.com/opentable/sous/util/cmdr"
+)
+
+// DeploymentQuery supports querying deployments.
+type DeploymentQuery struct {
+	StateManager  sous.StateManager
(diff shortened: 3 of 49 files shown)


sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

lachlanh:
Feel like this should be split up, too distinct features added in a large PR, very hard to review.

samsalisbury:
@lachlanh I agree in hindsight, one of the reasons this got so large was that it touched untested code, so needed to add a lot of tests for things not essential to this change. As I did that, I ended up having to touch more untested code, add tests for that, etc., there is a domino effect... I got this feedback from @yk12OT too, so we went through the code together and ended up making a couple of changes. The upshot is that coverage is up overall (although not for this diff - I had to stop at some point, as was chasing my tail with it).

I'll see what I can do to make this easier to review... it's definitely an issue, but I fear splitting it up might be quite hard to do at this stage...

graph/graph.go
```diff
@@ -171,7 +171,8 @@ func newUser(c LocalSousConfig) sous.User {
 	return c.User
 }
 
-func newSousGraph() *SousGraph {
+// NewSousGraph returns a new sous graph.
+func NewSousGraph() *SousGraph {
```
nyarly: Why was this exported?

test/smoke/smoke_test.go
```diff
@@ -46,6 +26,28 @@ func TestSmoke(t *testing.T) {
...
 	})
 
+	m.Run("scale", func(t *testing.T, f *fixture) {
+		p := setupProject(t, f, f.Projects.HTTPServer())
+		flags := &sousFlags{kind: "http-service", tag: "1.2.3", cluster: "cluster1", repo: p.repo}
+		reqID := f.DefaultSingReqID(t, flags)
+
+		assumeSuccessfullyDeployed(t, f, p, flags, reqID)
+
+		// TODO SS: Make scale provide same feedback as deploy (i.e. don't
```
nyarly: We should really prioritize this

cli/queries/deployment_test.go
```diff
@@ -0,0 +1,68 @@
+package queries
+
+import (
+	"testing"
+
+	sous "github.com/opentable/sous/lib"
+)
+
+func TestDeployment_ParseAttributeFilters_ok(t *testing.T) {
```
nyarly: It seems like `strings.Split(..., " ")` would satisfy this test.

cli/queries/deployment.go
```diff
@@ -0,0 +1,112 @@
...
+// DeploymentQueryResult is the result of the query.
+type DeploymentQueryResult struct {
+	// Deployments is the final query result.
+	Deployments sous.Deployments
+}
+
+// ParseAttributeFilters parses deployment filters in the format:
+//   <name>=<true|false> <name2>=<true|false>
+// It returns a valid set of deployment filters.
+func (q *Deployment) ParseAttributeFilters(s string) (*DeploymentAttributeFilters, error) {
```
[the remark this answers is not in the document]
samsalisbury: To my mind DeployFilterFlags are core to all sous operations and merit top-level flags, and their complex interaction with source context. Conversely these filters are only applicable to querying en mass, are completely self-contained, and are much more lightweight in their definitions, so will be trivial to add more as we go. Re parsing; I kept it as simple as possible (true/false only for values) but I agree the tests need improving. This felt like a natural fit for these kinds of arbitrary filters, I think this would be a valuable feature to get out soon as we have use cases for it right now.

cli/queries/deployment_test.go
samsalisbury: Not quite, but I agree these tests should be expanded to cover more cases and make stronger assertions.

graph/graph.go
samsalisbury: I think for tests in CLI package, on phone right now will check if it’s necessary tomorrow.

test/smoke/smoke_test.go
samsalisbury: I concur, would rather add that as a separate PR as doing it well will require a generic solution for sous manifest get as well as other single-field update commands we may add on future.

cli/queries/deployment_filters.go
```diff
@@ -0,0 +1,90 @@
...
+
+// MaxConcurrentArtifactQueries is the max number of concurrent artifact
+// queries. 100 is a conservative value to ensure we don't run out of file
+// descriptors locally.
+// NOTE: If users complain this is too slow we could make this configurable
+//       by env var, or perhaps lookup the max file descriptors via ulimit -n...
+const MaxConcurrentArtifactQueries = 100
+
+// DeploymentFilters is the argument that determines which deployments are
+// returned by a query.
```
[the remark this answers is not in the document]
samsalisbury: Resolved in error.. see answer above...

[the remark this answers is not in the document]
samsalisbury: Further, DFF is about identity of deployment/manifest, these other filters are about general true/false facts not related to identity, including cross-referencing with other entities (e.g. artifacts).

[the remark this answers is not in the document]
nyarly: I can see the "cross reference" case, but "owners are empty" and "numinstances=0" are both related to the deployment.

cli/queries/deployment.go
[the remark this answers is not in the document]
nyarly: My point wrt parsing is that these could easily be flags (`-no-owners` `-maxinstances=0`) and use existing shell/flag package parsing.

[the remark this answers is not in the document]
samsalisbury: See latest updates, I've added these as flags but without having to maintain them in the CLI package, which would have made adding new filters more painful. We still call `strconv.ParseBool` and we cannot use a `flag.BoolVar` because there are 3 possible states: both (i.e. ""), "true", "false". @nyarly

cli/actions/manifest_get_test.go
```diff
@@ -11,43 +12,58 @@ import (
...
+		}
+		require.NoError(t, err)
 
-	if assert.Len(t, control.Calls(), 1) {
-		assert.Regexp(t, "/manifest", control.Calls()[0].PassedArgs().String(0))
-		params := control.Calls()[0].PassedArgs().Get(1).(map[string]string)
-		assert.Contains(t, params, "repo")
-		assert.Contains(t, params, "flavor")
-		assert.Equal(t, params["flavor"], "chocolate")
+		if assert.Len(t, control.Calls(), 1) {
```
lachlanh: is there any need to conditionally assert in a test ? if the assert fails dosen't the test exit ?

cli/queries/artifact.go
```diff
@@ -0,0 +1,38 @@
...
+// is no match and no error determining that.
+func (q *ArtifactQuery) ByID(sid sous.SourceID) (*sous.BuildArtifact, error) {
+	ba := &sous.BuildArtifact{}
+	header := q.User.HTTPHeaders()
+
+	_, err := q.Client.Retrieve("./artifact", sid.HTTPQueryMap(), ba, header)
+	if err == nil {
+		return ba, nil
+	}
+	if strings.Contains(err.Error(), "404 Not Found") {
```
lachlanh: Not a huge fan of returning nil, nil here. Could you move the 404 check on the error to the exists function below ?

cli/queries/deployment_filters.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, name := range filterOrder {
+		if filter, ok := named[name]; ok {
+			f.filters = append(f.filters, filter)
+		}
+	}
+	return nil
+}
+
+func (f *DeploymentAttributeFilters) apply(ds sous.Deployments) sous.Deployments {
+	for _, f := range f.filters {
```
lachlanh: confusing variable name here. being a bit less terse will help readability (2 f's in scope at this point)

cli/queries/deployment_filters.go
```diff
@@ -0,0 +1,134 @@
...
+		}()
+	}
+	wg.Wait()
+	close(errs)
+
+	for err := range errs {
+		log.Println(err)
+	}
+	return filtered
+}
```
lachlanh: for my benefit, could you explain why we need the parallel code for hasImage ? I am not really sure what this function is doing ?

cli/queries/deployment_filters_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	trueResult := filter(ds, true)
+	falseResult := filter(ds, false)
+
+	gotTrue := trueResult.Len()
+	gotFalse := falseResult.Len()
+
+	wantTrue := 1
+	wantFalse := 2
+	if wantTrue+wantFalse != ds.Len() {
```
lachlanh: any reason not to use asserts on these test conditions ?

cli/queries/deployment_test.go
```diff
@@ -0,0 +1,26 @@
...
+func TestDeployment_Result(t *testing.T) {
+	sm := sous.NewDummyStateManager()
+	sm.State = sous.DefaultStateFixture()
+	aq := ArtifactQuery{}
+	q := Deployment{
+		StateManager:  sm,
+		ArtifactQuery: aq,
+	}
+	r, err := q.Result(DeploymentFilters{})
+	if err != nil {
```
lachlanh: as above asserts ?

lib/deployment_dumper_test.go
```diff
@@ -17,3 +18,64 @@ func TestDeploymentDumper(t *testing.T) {
...
+			t.Errorf("got deployment not in original set: %v", got)
+			continue
+		}
+		if !got.Equal(original) {
+			t.Errorf("deployment %q did not round-trip correctly: got %v; want %v",
+				got.ID(), got, original)
+		}
+	}
+
+}
```
lachlanh: quite alot of code to test j.Encode() from what I can see.

cli/queries/deployment_filters.go
nyarly: IIUC, this filter checks that the deployments in the GDM have images in the docker registry (or recorded in the search index) - rather than run dozens to hundreds of HTTP requests in series, it make some sense to parallelize them.

cli/actions/manifest_get_test.go
nyarly: Not so. All the failing asserts get reported. There's a related package testify/require that does what you're thinking of, and it can be helpful sometimes to use that. `assert` functions return a bool for exactly this reason.

For instance, in this case, if the `control.Calls()` array doesn't have at least 1 member, indexing into it will panic the tests.

cli/queries/artifact.go
nyarly: Yeah, https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

lachlanh:
@samsalisbury hopefully comments make sense. scale logic was pretty straightforward the query bits took much longer to get through but feel that I understand the bulk of it.

cli/queries/artifact.go
samsalisbury: I definitely want a distinction between “this artifact does not exist” (i.e. nil, nil) and “unable to determine existence of artifact” (nil, err). That could be done in the Exists func though, might make more sense there. I also agree pretty strongly with Dave Cheney’s article, esp about not checking content of error.Error, however it seemed that no status code was available on this error. I could either modify the restful package to surface 404, preferably with the 404 body as well, but that would be a big change. Alternatively I could circumvent that package & make a raw HTTP request to get the status code & body to check, but that would be a fairly big change just for this one check. When writing this I considered all these options and concluded that this imperfect check had the best cost/benefit ratio, and smoke tests prove that it works as expected in many cases (obv it could be fooled by a 404 from a different resource altogether), which is why to make it bomb-proof we also need to check the body not just the status code. I can live with it as is, but happy to spend the time modifying things up the stack if deemed crucial. I won’t make any change until we discuss as it will be quite an investment of time.

samsalisbury: I actually prefer returning (nil, nil) here rather than in "exists" to indicate "this artifact does not exist", because it's a simple interface to give 3 possible answers, so call sites can just check those 2 values. It would be possible to return (bool, artifact, error) so the bool could be checked for "existence", the error for whether or not existence was successfully determined, and the artifact will be nil or not depending on a combination of the bool and error (false or non-nil error means nil artifact), but this seems lik unnecessary complexity... Maybe an extra "MustByID" func that returns an error for both "does not exist" and "cannot determine existence" could be useful at some call sites, but would encourage just checking/reading the error at those call sites to determine whether we know for sure if it exists or not, so I'm not a fan of adding that func.

cli/queries/deployment_filters_test.go
samsalisbury: I find that the `testify/assert` package slows me down when writing tests (I have to keep looking up argument order, which is expected, which is actual?) It also makes some confusing assumptions about equality, which have bitten me before. I really want to trust the tests I write, but with the assert package there's always a lingering worry that my test is bad, not failing for exactly the right reason. So when starting a new test file, I tend to go with plain old Go code (which is less ambiguous) for tests, even though it's slightly more verbose. When adding or changing tests that already use that assert package, I continue to use it for continuity.

cli/queries/deployment_test.go
samsalisbury: This is an assertion, just on 3 lines instead of 1... But see above about why I find `testify/assert` unhelpful. Reading the above, we know for sure what the assertion is just from the visible test code, so it's easy to see what it's doing. With the assert package, we need to understand another mind (the author of that package) for what they mean by equality. (It's not the `=` operator, and it's not an obvious definition to me, it involves conversion to byte slices and reflection).

lib/deployment_dumper_test.go
samsalisbury: We are checking that each line of stdout is a valid JSON doc that accurately represents the deployment it's meant to. Some of this might be covered elsewhere, and probably is indirectly, but the output of this func is the bytes it writes out, so seems reasonable to assert those bytes are right... We could prob drop the round-trip assertions but I don't think they're brittle, as we'll always want deployments to round-trip to JSON and back.

samsalisbury:
Thanks for the reviews @lachlanh @nyarly, replies inline, I'll prob push a few commits later today to address a couple of issues.

cli/queries/deployment_filters.go
samsalisbury: Fixed in 915d5faa4e43a30b33adc22232d8979859cac3e2 @lachlanh

cli/queries/deployment_filters.go
samsalisbury: Done, the latest on this branch now treats handling of parallel filters separately from the contents of that filter. It also adds better error handling, as previously errors were reported but did not cause failure, so the careless user may have not realised their results were inaccurate due to failed filters. Now if any filter fails the whole operation is reported as an error as it should be. @lachlanh @nyarly

PR #812, opened 2018-11-19
POC: adding atomic http get counter

yk12OT:


yk12OT:
@samsalisbury - please take a look... not entirely sure if that is what we are looking for...

BTW, the ticket SOUS-1031 does not have any description what it is meant by "concurrent requests"...
As in real world (Einstein's general relativity) - everything is relative. "concurrent" form which observer's view point? Should we run a dedicated go-routine which would measure concurrency, i.e. will make quantized time intervals like 1ms or so, and tell how many GET requests were received during that quantized time interval?

yk12OT:
Also it is not clear how often should we report that metric to graphite... If we go with dedicated go-routing, at least we will have a code block where graphite push belongs to...

sous-machine-account:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

samsalisbury:
Hi @yk12OT thanks for looking into this, apologies the ticket wasn't totally clear. Let's take the simplest reading of the ticket (which I've now updated) to read the number of concurrent requests being handled by a single sous server process at a point in time. "Point in time" being defined by the value of an atomically incremented/decremented counter within that process, which should be incremented fairly early in the request handler, and decremented fairly late in the handler. This won't be a perfect measure but should provide a rough/good enough idea of the true value. For this measure we shouldn't quantise time but just report the value of that counter at the moment the sample is taken, say once per minute. (The "requests per second" measure can be derived from the time series of total requests handled by the process since it started, so we don't need to measure the directly, although in my opinion recording this kind of "denormalised" metric can sometimes be useful for query perf reasons.)

samsalisbury:
@yk12OT I should have mentioned that the logging package already has metrics pieces like "counter" and "timer" implemented, which we should use. See `util/logging/messages.go` for implementations, and search for usages to see examples (e.g. below).

I believe these metrics are all automatically sent to graphite, so simply calling these with an appropriately named counter should be enough.

```
$ ag -sQ '.IncCounter('
util/sqlgen/messages.go
125:            sink.IncCounter(strings.Join([]string{msg.mainTable, msg.dir.String(), "errs"}, "."), 1)                                                                                
129:    sink.IncCounter(strings.Join([]string{msg.mainTable, msg.dir.String(), "count"}, "."), int64(msg.rowcount))                                                                     

util/logging/silentmessageerror.go
32:     metrics.IncCounter("silent-messages", 1)

util/logging/messages/clienthttpresponse.go
171:            metrics.IncCounter(name, 1)

ext/docker/cachemessages.go
42:     ms.IncCounter("cache-hit", 1)
67:     ms.IncCounter("cache-miss", 1)
92:     ms.IncCounter("cache-error", 1)

lib/resolvecompletemessage.go
29:     m.IncCounter("resolution-count", 1)
```

PR #813, opened 2018-11-22
Test advisories #trivial

samsalisbury:
Adds a simple test that when no advisories are allowed, we cannot deploy anything with advisories.

This test feeds into to work on migrating to postgres, as a major issue there is lack of State.Defs.

The work in ./util/testmatrix allows us to write tests with alternative initial states, which will be useful for testing a number of features.


sous-machine-account:
# LGTM!

</p>

lachlanh:
Seems like it is doing more than just adding to advisories test ?This looks like it refactors a bunch of stuff into graph/storage, also a pretty big pr. Anyway to break this up or clarify what the intent of the pr is ?

.travis.yml
```diff
@@ -43,6 +43,8 @@ env:
     - TEST_COMMAND="GO_TEST_RUN=TestSmoke/git/simple make test-smoke"
     - TEST_COMMAND="GO_TEST_RUN=TestSmoke/git/split make test-smoke"
     - TEST_COMMAND="GO_TEST_RUN=TestBugs/git/simple make test-smoke"
+    - TEST_COMMAND="GO_TEST_RUN=TestDefs/git/simple make test-smoke"
+    - TEST_COMMAND="GO_TEST_RUN=TestDefs/db/simple make test-smoke"
```
samsalisbury: Runs the new TestDefs in test/smoke.

graph/graph.go
```diff
@@ -82,26 +80,6 @@ type (
...
-		sous.StateManager
-		Error error
-	}
-
-	gitStateManager struct {
-		sous.StateManager
-		Error error
-	}
-	diskStateManager struct{ sous.StateManager }
-
```
samsalisbury: All this code was moved to another file: graph/storage.go

graph/graph.go
```diff
@@ -234,11 +212,14 @@ func AddFilesystem(graph adder) {
...
 		newConfigLoader,
 		newMaybeDatabase, // we need to be able to progress in the absence of a DB.
+		newDuplexStateManager,
 		newServerStateManager,
 		newServerClusterManager,
 		newDistributedStateManager,
 		newGitStateManager,
 		newDiskStateManager,
+		newPrimaryStateManager,
+		newSecondaryStateManager,
```
samsalisbury: duplex, primary, secondary statemanagers exist primarily to make testing easier.

graph/graph.go
```diff
@@ -634,101 +615,6 @@ func newInMemoryClient(srvr ServerHandler, log LogSink) (HTTPClient, error) {
...
-		return nil, errors.New("no server configured for state management")
-	}
-	hsm := sous.NewHTTPStateManager(cl, tid, log.Child("http-state-manager"))
-	return &ClientStateManager{StateManager: hsm}, nil
-}
-
-func newHTTPStateManager(cl HTTPClient, tid sous.TraceID, log LogSink) *sous.HTTPStateManager {
-	return sous.NewHTTPStateManager(cl, tid, log.Child("http-state-manager"))
-}
-
```
samsalisbury: All this removed code was moved to another file: graph/storage.go.

graph/graph.go
```diff
@@ -738,36 +624,6 @@ func newStatusPoller(cl HTTPClient, rf *RefinedResolveFilter, user sous.User, lo
...
-// NewCurrentGDM returns the current GDM.
-func NewCurrentGDM(state *sous.State) (CurrentGDM, error) {
-	deployments, err := state.Deployments()
-	if err != nil {
-		return CurrentGDM{}, initErr(err, "expanding state")
-	}
-	return CurrentGDM{deployments}, initErr(err, "expanding state")
-}
-*/
-
```
samsalisbury: This commented out code was just deleted.

graph/graph_test.go
```diff
@@ -150,26 +150,6 @@ func TestComponentLocatorInjection(t *testing.T) {
...
-func TestStateManagerSelectsDuplex(t *testing.T) {
-	smgr := injectedStateManager(t, &config.Config{Server: "", StateLocation: "/tmp/sous"})
-
-	_, ok := smgr.StateManager.(*storage.DuplexStateManager)
-	if !ok {
-		t.Errorf("Injected %#v which isn't a DuplexStateManager", smgr.StateManager)
-	}
-}
-*/
-
```
samsalisbury: This commented code was just deleted.

graph/storage.go
```diff
@@ -0,0 +1,146 @@
+package graph
```
samsalisbury: This file contains code migrated from graph/graph.go. (graph/graph.go is a huge file, so trying to bring some order to make things easier to find. Also contains new code added to support testing.

graph/storage.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+func newHTTPStateManager(cl HTTPClient, tid sous.TraceID, log LogSink) *sous.HTTPStateManager {
+	return sous.NewHTTPStateManager(cl, tid, log.Child("http-state-manager"))
+}
+
+// newPrimaryStateManager returns the configured primary, and any error
+// encountered in its construction.
+func newPrimaryStateManager(c LocalSousConfig, git gitStateManager, dist DistStateManager) (primaryStateManager, error) {
```
samsalisbury: new func

graph/storage.go
```diff
@@ -0,0 +1,146 @@
...
+	if c.DatabasePrimary {
+		return dist.StateManager, dist.Error
+	}
+	return git.StateManager, git.Error
+}
+
+// newSecondaryStateManager returns the configured secondary state manager if
+// there were no errors constructing it. Otherwise it emits a log message with
+// the error and falls back to the log-only state manager.
+func newSecondaryStateManager(log LogSink, c LocalSousConfig, git gitStateManager, dist DistStateManager) secondaryStateManager {
```
samsalisbury: new func

graph/storage.go
```diff
@@ -0,0 +1,146 @@
...
+// If it returns a sous.GitStateManager, it emits a warning log.
+func newClientStateManager(cl HTTPClient, c LocalSousConfig, mdb MaybeDatabase, tid sous.TraceID, rf *sous.ResolveFilter, log LogSink) (*ClientStateManager, error) {
+	if c.Server == "" {
+		return nil, errors.New("no server configured for state management")
+	}
+	hsm := sous.NewHTTPStateManager(cl, tid, log.Child("http-state-manager"))
+	return &ClientStateManager{StateManager: hsm}, nil
+}
+
+func newDuplexStateManager(p primaryStateManager, s secondaryStateManager) duplexStateManager {
```
samsalisbury: new func

graph/storage_test.go
```diff
@@ -0,0 +1,276 @@
+package graph
```
samsalisbury: Previously there were no tests for the state manager constructors, this file covers (I think) all the scenarios with primary/secondary, and includes an integration test of the interaction of primary/secondary/duplex. We can now be really confident which state manger is being picked.

lib/deployment_manager_test.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func TestDeploymentManager_ReadDeployment(t *testing.T) {
 
-	innerState := DefaultStateFixture()
+	innerState := DefaultStateFixtureWithFlavorsOffsets()
```
samsalisbury: DefaultStateFixtureWithFlavorsOffsets is providing a legacy configuration of the state fixture used in tests. This was created to avoid having to change the tests themselves.

lib/state_fixture.go
```diff
@@ -28,51 +35,188 @@ func DefaultStateFixture() *State {
 	})
 }
 
+// DefaultStateFixtureWithFlavorsOffsets is similar to DefaultStateFixture but
```
samsalisbury: This file is only used by tests, to generate realistic test data. The changes here make data generation more modular so tests can say "give me the default test state fixture, except this one detail that is important to the test I'm writing".

lib/state_fixture.go
```diff
@@ -28,51 +35,188 @@ func DefaultStateFixture() *State {
...
+// GenerateManifestID generates a ManifestID using the supplied options.
+func GenerateManifestID(idx int, o ManifestIDOpts) ManifestID {
+	// NOTE SS: This seems overly complex, the justification is to allow
+	// existing tests to still work without changing the tests other than
+	// having them generate the manifest IDs they expect.
+	// It's not possible to use simple fmt.Sprintf calls because many
+	// of these tests expect the repo to contain multiple instances of
+	// the index.
+	// Rather than specialise this to those existing tests, just make it
+	// general purpose.
```
samsalisbury: This note justifies the annoying complexity of the changes in this file.

test/smoke/defs_test.go
```diff
@@ -0,0 +1,36 @@
...
+		flags := &sousFlags{kind: "http-service", tag: "1.2.3", cluster: "cluster1", repo: p.repo}
+
+		initBuild(t, p, flags, setMinimalMemAndCPUNumInst1)
+
+		stderr := p.MustFail(t, "deploy", flags.SousDeployFlags())
+		want := "Advisory unacceptable on image:"
+		if !strings.Contains(stderr, want) {
+			t.Errorf("got stderr %q; want it to contain %q", stderr, want)
+		}
+	})
```
samsalisbury: This is the test mentioned in the original description.

test/smoke/fixture.go
```diff
@@ -48,14 +48,13 @@ type fixture struct {
 
 var sousBin = mustGetSousBin()
 
-func newFixtureConfig(testName string, s testmatrix.Scenario) fixtureConfig {
+func newFixtureConfig(testName string, s scenario) fixtureConfig {
```
samsalisbury: Changes in this file enabled writing the advisories test.

test/smoke/runner.go
```diff
@@ -14,10 +14,25 @@ var sup *testmatrix.Supervisor
 // themselves.
 type runner struct{ *testmatrix.Runner }
 
+type test func(*testing.T, *fixture)
```
samsalisbury: Changes in this file enabled writing the advisories test.

test/smoke/singularity.go
```diff
@@ -152,7 +152,7 @@ func waitFor(t *testing.T, what string, timeout, interval time.Duration, f func(
 				}()
 				if err != nil {
 					// Log direct to stderr for live updates.
-					elapsed := startTime.Sub(time.Now())
+					elapsed := time.Now().Sub(startTime).Round(time.Second)
```
samsalisbury: Fixes test output.

graph/storage_test.go
```diff
@@ -0,0 +1,276 @@
...
+					StateManager: &storage.GitStateManager{},
+					Error:        tc.gitErr,
+				},
+				DistStateManager{
+					StateManager: &sous.DispatchStateManager{},
+					Error:        tc.dbErr,
+				},
+			)
+
+			scoop := struct {
```
lachlanh: @samsalisbury for future reference would be educational for me to understand what this is doing.

lachlanh:
Thanks for the commentary. The state manager refactor and tests make sense and look good. The refactoring of the smoke tests and test matrix I don't understand. I think it would be good to walk through how the smoke tests work at some point.

graph/storage_test.go
samsalisbury: Lines 175 to 177 define a variable `scoop` whose value `{}` is a new instance of the anonymous struct with one field `Got duplexStateManager`. This is used to extract (scoop out) a `duplexStateManager` from the graph `g` by injecting into it on line 179. This tests what duplexStateManager is produced by the minimal section of the graph defined above. Note we inject known config and errors into that graph via the tc param, to check how behaviour differs.

test/smoke/defs_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+)
+
+func noAllowedAdvisories(c *fixtureConfig) {
+	for _, name := range c.InitialState.Defs.Clusters.Names() {
+		cl := c.InitialState.Defs.Clusters[name]
+		cl.AllowedAdvisories = []string{}
+		c.InitialState.Defs.Clusters[name] = cl
+	}
+}
```
samsalisbury: @lachlanh the refactor in util/testmatrix (new RunScenario method) and in test/smoke allows this kind of func to be called to modify initial state before a test is run. Because setup for smoke tests is huge, and results in a huge amount of overall state (config files, databases, git repos etc), we don't want to repeat setup code in each test. For a while we could get away with a single initial state for all tests, and things were simple, however there are some tests that require specific initial states in order to be meaningful. Rather than have a totally separate way of creating those initial states, this kind of func allows us to just give the default initial state the specific attributes it needs for the test to be meaningful, and otherwise it remains the same.

Initial state needs to be known very early in test setup, as it dictates how many clusters there are, and therefore how many sous servers we are spinning up for the test. Each of those servers needs to know about all the others, but we cannot reconfigure them on the fly. For each server we first need to create a Postgres database (seeded with initial state), and if using Git, a clone of the test GDM (which itself contains the initial state). So initial state is not something we can change in the normal body of the test, which expects to begin with a client already configured to point to running, properly configured Sous servers.

test/smoke/defs_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+func TestDefs(t *testing.T) {
+
+	m := newRunner(t, matrix())
+
+	m.RunScenario("no-allowed-advisories", noAllowedAdvisories, func(t *testing.T, f *fixture) {
```
samsalisbury: RunScenario is similar to Run, except it allows a func to modify the intial state (actually anything in the fixture config) before the full, rich fixture object is created from it and passed to the test.

test/smoke/fixture.go
samsalisbury: @lachlanh The purpose of the testmatrix package is to abstract away the definition of matrices, the orchestration of setup and teardown of parallel tests based on these matrices, and the logging of summary data, so tests can just see a `Scenario` i.e. a single vector pulled from the matrix. The scenario contains a value for each of the dimensions defined in the matrix (we have the 2 dimensions at the moment: "store" (db or git) and "project" (simple or split) so each test is run once for every combination in the matrix, passed that combination as the Scenario.

testmatrix.Scenario requires doing lookups by string, and returns `interface{}` which isn't easy to use and requires casting. The func `unwrapScenario()` translates these generic scenarios into strongly typed `scenario` making them easier to use. This refactoring uses that strongly-typed `scenario`, and pushes the calling of `unwrapScenario` up into the `fixtureFactory` func.

test/smoke/fixture.go
```diff
@@ -125,6 +127,15 @@ func newFixture(t *testing.T, s testmatrix.Scenario) testmatrix.Fixture {
 	return tf
 }
 
+// newFixture transforms a testmatrix.Scenario into a sous-specific fixture.
+func newFixture(t *testing.T, s scenario) *fixture {
+	return newConfiguredFixture(t, s)
+}
+
+func fixtureFactory(t *testing.T, s testmatrix.Scenario) testmatrix.Fixture {
+	return newFixture(t, unwrapScenario(s))
```
samsalisbury: fixtureFactory deals purely in terms of `util/testmatrix` types and is passed to the testmatrix Runner so it can generate fixtures for the tests it runs. Now that we have the `RunScenario` func, we may want to eventually get rid of the `Run` func, and just require each test to specify its initial state. Then we would not need to pass a fixture factory to `testmatrix` at all. I did not go that far in this PR, as it would require modifying every smoke test.

samsalisbury: @lachlanh ^

test/smoke/runner.go
```diff
@@ -14,10 +14,25 @@ var sup *testmatrix.Supervisor
...
+
 // Run is analogous to Runner.Run, but accepts a func in terms of strongly typed
 // fixture rather than having to manually unwrap scenarios.
-func (r *runner) Run(name string, test func(*testing.T, *fixture)) {
-	r.Runner.Run(name, func(t *testing.T, c testmatrix.Context) { test(t, c.F.(*fixture)) })
+func (r *runner) Run(name string, test test) {
+	r.Runner.Run(name, func(t *testing.T, c testmatrix.Context) {
+		test(t, c.F.(*fixture))
+	})
```
samsalisbury: `runner.Run` just translates from strongly-typed `test`s and `fixture`s to generic `testmatrix.Test`s and `testmatrix.Context`s to avoid casting and allow writing strongly-typed tests. @lachlanh

test/smoke/runner.go
```diff
@@ -14,10 +14,25 @@ var sup *testmatrix.Supervisor
...
+}
+
+type fixtureConfigFunc func(*fixtureConfig)
+
+func (r *runner) RunScenario(name string, ff fixtureConfigFunc, test test) {
+	r.Runner.RunScenario(name, func(t *testing.T, s testmatrix.Scenario, lf *testmatrix.LateFixture) {
+		scenario := unwrapScenario(s)
+		fix := newConfiguredFixture(t, scenario, ff)
+		lf.Set(fix)
+		test(t, fix)
```
samsalisbury: `runner.RunScenario` is similar to `runner.Run` but translates to `testmatrix.RunScenario` and conveniently calls `LateFixture.Set` so we don't need to call it in every test. I'm not 100% happy with the `LateFixture` idea but it exists so that tests can register (`Set`) their fixtures with the `testmatrix.Runner` so that it can perform teardown at the right time. For now, since this rough edge is only in one place I think it's ok... @lachlanh

samsalisbury:
@lachlanh I've explained some of the testmatrix refactoring, motivations and compromises made above, please let me know if anything still isn't clear, thanks.

PR #817, opened 2018-11-29
Travis stages #trivial

samsalisbury:
Splits up Travis test stages stages into:

- Static static checks, also runs gometalinter for informational purposes only.
- Unit tests
- Integration and Smoke tests
- Release (only visible when building a tag)

These are run in the order shown above, in sequence, so we can get faster failures by guaranteeing fast checks run first. A failure in any stage cancels the rest of the build. I would like to think that if we use less compute on Travis then we might get a better QoS for the stuff we do run, but I've no idea if that's the case or not.


samsalisbury:
Note this also avoids installing unnecessary dependencies for build jobs that don't need them. We could probably go even further with this (maybe sudo and docker can be moved to the test stage only?) but having burnt half a day on this I'm leaving it as-is for now.

sous-machine-account:
# LGTM!

</p>

PR #820, opened 2019-01-08
[wip] +Migra -liquibase

samsalisbury:
Opening this for visibility, I intend to use this to do an actual migration in dev before considering it complete. There is some related work in our migrations repo (migra branch) that needs to be updated to work with this as well.


database/README.md
```diff
@@ -0,0 +1,71 @@
+# Sous Postgres Database
+
+Files in this directory represent the Sous database.
+
+- base.sql is the latest released version of the schema - only modified on release.
```
nyarly: If we're going this route, it seems like we should have versioned schemas (i.e. there should now be a `0.6.0.sql`) and an `unreleased.sql` as well as migrations between versions. Base might link to the last released version... But consider the case where some environments have been updated and some haven't...

nyarly: That said, I really start to wonder if that approach would be worth the hassle.

samsalisbury: Yes I considered keeping versioned schemas around as well, as it would mean we could automatically verify proposed migrations of live databases between any two versions. I did not go down that route yet to keep things a little simpler for now, but this approach keeps that option open if we think it's worth it in future. The existence of `pending.sql` is only to formalise the PR review process (forcing us to consider the proposed migration script as an individual piece). When moving between an older version (before previous release) and current, we would have to manually verify the migra-generated migration script as things stand here, but moving from previous to current could be done automatically by comparing the proposed live migration with the reviewed one. This shouldn't affect us immediately though as we are only migrating one generation.

It would be trivial to generate versioned schema files and migration scripts in retrospect, using git to traverse release tags. Maybe that is something we should add to the release script. I'll see how much work is involved there, but likely open a separate PR for it as this one is almost ready for use.

One issue is that this PR does not yet include a migration from 0.5.x to 0.6.0, (just the full 0.6.0 schema) so we are missing the human review of pending.sql, I'll get that fixed today.

Note on filenames: `pending.sql`is suggested by migra docs so went with that rather than 'unreleased'. `base.sql` mirrors the old `base.xml` and `head.sql` represents the schema for the head of the branch that changes it.

samsalisbury:
Closing until/if we decide to come back to this.

PR #825, opened 2019-01-16
bugfix: Build prefixed tags

samsalisbury:
See changelog diff in this PR for full details.


lib/source_context.go
```diff
@@ -126,9 +126,18 @@ func parseSemverTagWithOptionalPrefix(tagName string) (prefix string, v semv.Ver
 	return prefix, v, err
 }
 
+// ParseBuildTag should be used in all places where we parse build tags.
```
nyarly: This merits discussion. IMO either Sous Tags are semver (and we relax that for the nearestVersion and extract something) or they're opaque identifiers.

nyarly:
Only "requesting changes" because we should discuss first.

sous-machine-account:
# LGTM!

</p>

lib/source_context.go
samsalisbury: Just to summarise f2f talk about this: we already made the decision that we would accept prefixed tags as input, but discard the prefix for all internal operations and persistence. This PR addresses a new bug introduced in 0.6.0 where we broke that contract. We should also add a message when prefixed tags are used (from both detected context and `-tag` flag) that the prefix is ignored by sous, including in the manifest version field update that ensues from `sous deploy`). Since some teams are blocked deploying, we should get the fix out even if it means waiting for a follow-up PR to add the messaging, as this restores expected behaviour.

The new smoke test added is still broken due to the bug, so there is more work to do. I will follow this up on Friday if it's still outstanding, making this WIP for now. Internal users of sous are currently using 0.5.119 in TeamCity so are not blocked by this bug. See PR on teamcity integration repo effecting this change. @nyarly @lachlanh

PR #828, opened 2019-02-18
Fix server crash detection in smoke tests

samsalisbury:
This addresses the problem of flaky smoke tests, which was primarily caused by a sequencing error and calling `*os.Process.Wait` multiple times. Fixing that issue surfaced another one (see below), and resulted in some other changes for better logging, which needed to be fixed to get a green PR.

Test coverage is not good for this change, mostly because it modified some code that was already not coverd by unit tests (just by smoke tests). To start getting it back to a decent level, I split out 2 packages: `util/prefixpipe` and `util/testagents` in order to test those components separately. The former now has a simple test, but the latter still needs them. Since this PR is growing, I want to stop adding changes here and address test coverage and a few other things in a follow-up.

**New issue discovered (still needs investigation, but workaround in place):**
Sending HTTP requests to a sous server too early in its initialisation causes the server to first refuse connection, then return an empty response, and then die, with 3 HTTP calls, in that order. Only the first call needs to be during early initialisation, the remaining ones can be an arbitrarily long time later. (This can be reproduced reliably using `make test-smoke` if you remove that sleep, and a debugger can then be attached to see what's going on.)  This was happening rarely (for roughly 2% of tests). It is ameliorated for now using a sleep after the servers start, but we should find the bug in the server that allows an HTTP request to break it, so we can remove that sleep.

**Behavioural change of sous:**
Logging added to test process handling resulted in a change to behaviour, see CHANGELOG: Sous server now shuts down gracefully, draining HTTP requests, on a SIGINT or SIGTERM, and logs the signal and action taken. This was helpful in seeing which servers had received TERM and which KILL. Running locally, servers are occasionally receiving a KILL signal unexpectedly, which is logged, need to find out where that is coming from.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ var/
 .cleanprotect
 *.test
 .smoketest/
+.procman-pids
 
 
 ## Github/Gitignore: JetBrains
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ env:
   matrix:
     - TEST_COMMAND="bundle && bundle exec danger"
     - TEST_COMMAND="make test-integration"
-    - TEST_COMMAND="make test-unit"
+    - TEST_COMMAND="make compile-tests && make test-unit"
     - TEST_COMMAND="GO_TEST_RUN=TestOTPL/db/simple make test-smoke"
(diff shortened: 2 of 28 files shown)


sous-machine-account:
# LGTM!

</p>

